PR #4, opened 2019-04-15
oliver: 0003 "standev_phoronix_test_suite"

ode33:
Generally unreviewed, may be typos.

### DIFF
--- a/designs/0003-phoronix_test_suite.md
+++ b/designs/0003-phoronix_test_suite.md
@@ -1,0 +1,86 @@
+- Feature Name: stan-dev_phoronix_test_suite
+- Start Date: 2019-04-15
+- RFC PR: (leave this empty)
+- Stan Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Phoronix is an open-source tool for conducting software-based performance testing over a wide variety of hardware and OS-level configurations. Phoronix integrates natively with Openbenchmarking.org for composing results. Custom test-suites are added to the online repository and called locally using console commands like phoronix-test-suite install *[Test | Suite | OpenBenchmarking ID]*.
+
+# Motivation
+[motivation]: #motivation
+
+Benchmarking is crucial to the deployment of performant software. Properly testing algorithms and their implementations is non-trivial. Collecting results aids developer productivity and ultimately helps improve the optimization of the code base for all users. Enabling distributed testing ensures that the largest number of contributors can conveniently and accurately report how well their systems perform when executing reference code.
+
+How replicable are code-based performance improvements (or regressions) across a variety of hardware configurations? What code quality issues may be observed on certain computer configurations which may go undetected on other systems? Which unintuitive scenarios can we test for which information may not be avaiable through other sources?
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Users on Linux, Windows, Apple OS X, GNU Hurd, Solaris, or BSD Operating Systems download and install the Phoronix Test Suite and use several (two) console commands to install and benchmark their systems. These commands are generally `phoronix-test-suite install [*]` and `phoronix-test-suite benchmark [*]` with `[*]` denoting an arbitrary test-suite specification. Under certain workflows an internet connected test-suite user only needs to be able to install phoronix and obtain the Openbenchmarking ID of the test-suite being considered. 
+
+For test suite developers the process involves writing the test, specifying the system requirements for building it, and uploading to a team maintained Openbenchmarking repository. ALthough benchmarking in principally a hardware focused activity these test cases should be thought of as part of the debugging process for performance orientated end-user code. The results produced from benchmarking can be related to all manner of physical specifics about the state of the computer as it runs various tasks specified by the developer. These specifics include: compiler used, GPUs, Operating Systems, CPUs, disk drives, etc. Captured results from these subsytems are stored online and can be visually depicted using standardized, intuitive layouts. 
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+From the Phoronix Test Suite Guide:
+
+"Writing a test profile for the Phoronix Test Suite is a relatively quick and easy process for anyone familiar with common Linux commands and the basics of XML. To help you understand the design of the Phoronix Test Suite, this guide covers the steps needed to write a testing profile for a very simple application.
+
+The first step in the profile writing process is to, well, have a piece of software you’d like to use with the Phoronix Test Suite. This software can be closed-source or open-source and be virtually anything as long as it is compatible with an operating system that is supported by the Phoronix Test Suite.
+
+For this guide, the piece of software being used for demonstration is just a simple C++ program that calculates Pi to 8,765,4321 digits using the Leibniz formula. Below is this sample piece of software intended just for demonstration purposes."
+
+[C++ code omitted here]
+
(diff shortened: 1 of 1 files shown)


seantalts:
Hey, thanks for posting! Forgot about this, sorry. So is the proposal here whether to write a test suite that integrates with phoronix? And the motivation that it will then be hosted on their website and people can choose to run the benchmarks on their own personal hardware, which will automatically upload the results? 

I think anyone should feel free to do this if they like! This seems like the kind of thing where someone can just build it and if it's useful it will be used, you know? Or was this proposal to replace or change some existing system or design?

ode33:
No problem. What happens (on Linux anyways) is I can define a simple shell script with the commands and the test-suite runs a repeated trial over the calls for whatever metric we want to capture. I had a rudimentary test of this working on a couple CmdStan tests in about 30 minutes. It was trivial to link my account and send the formatted information to a aggregated online repository. The idea arose from a Discourse posting about awareness surrounding the differences in timing for different compiler and operating system setups. I see that kind of information posted all the time using this software: https://www.phoronix.com/scan.php?page=article&item=windows10-linux-browsers&num=2

I think that the architecture definitely supports quick build and test scenarios. If someone wanted reliable and consistent testing across different builds (something mentioned on Discourse) this would be the way to go. In most scenarios I don't think Phoronix is robust enough to handle all of the routine developer testing for a project like CmdStan, but it could be helpful in certain situations. Perhaps prior to a release or if a small batch of tests produced inconclusive results and more information would improve decision making.

seantalts:
How does it provide the "reliable and consistent testing across different builds?" That does sound good.

ode33:
The test is downloaded, built and run from a developer specified source. Test-suite developers can define which version of the build tools are to be used, or alternatively test to compare different build tools.

Once defined all test-suite users are running the same test regardless of other system variables. Major system specifications which differ are tracked and can be used to demarcate the results.

seantalts:
Who causes the test to be downloaded? Is phoronix basically a repo of benchmarks that users can choose to run on their hardware? And so getting results there would depend on a bunch of users reliably running these, right?

ode33:
Basically yes, there is a source repo for the test code. The user has to make sure they use the same test name as everyone else and select the option to upload their system information.

designs/0003-phoronix_test_suite.md
```diff
@@ -0,0 +1,86 @@
...
+Phoronix downloads, compiles, and executes the test-suite automatically. A standard phoronix test run is 3 repetitions with more added if the standard deviation of these exceed a threshold. Tests which attempt to profile large portions of the compiled code base will require extended compute times. 
+
+Reliance on benchmarking and performance hardware reports may distract from other important project-level goals.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The Git repository `seantalts/perf-math` and other scripts have been posted to various pull requests and issues on `stan-dev/math`. In most cases these are related to micro-benchmarks (relatively small code snippets). In other cases these tests may not be conveniently replicable or the reporting may lack comprehensive details which would be valuable for drawing qualitative comparisons. 
+
+In order to build on the capacity of the stan-dev team to draw conclusions regarding performance improvements and make better informed decisions for future code revisions a more convenient and extensible system could be introduced.
```
seantalts: Here we could also learn about other alternatives to phoronix

designs/0003-phoronix_test_suite.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Benchmarking is crucial to the deployment of performant software. Properly testing algorithms and their implementations is non-trivial. Collecting results aids developer productivity and ultimately helps improve the optimization of the code base for all users. Enabling distributed testing ensures that the largest number of contributors can conveniently and accurately report how well their systems perform when executing reference code.
+
+How replicable are code-based performance improvements (or regressions) across a variety of hardware configurations? What code quality issues may be observed on certain computer configurations which may go undetected on other systems? Which unintuitive scenarios can we test for which information may not be avaiable through other sources?
+
```
seantalts: I think we need more motivation for why phoronix in particular here.

designs/0003-phoronix_test_suite.md
```diff
@@ -0,0 +1,86 @@
...
+For test suite developers the process involves writing the test, specifying the system requirements for building it, and uploading to a team maintained Openbenchmarking repository. ALthough benchmarking in principally a hardware focused activity these test cases should be thought of as part of the debugging process for performance orientated end-user code. The results produced from benchmarking can be related to all manner of physical specifics about the state of the computer as it runs various tasks specified by the developer. These specifics include: compiler used, GPUs, Operating Systems, CPUs, disk drives, etc. Captured results from these subsytems are stored online and can be visually depicted using standardized, intuitive layouts.
```
seantalts: Here we need to know exactly how we'd use it in Stan. Is this just for Math? What is an example test you would write?

designs/0003-phoronix_test_suite.md
```diff
@@ -0,0 +1,86 @@
...
+
+For test suite developers the process involves writing the test, specifying the system requirements for building it, and uploading to a team maintained Openbenchmarking repository. ALthough benchmarking in principally a hardware focused activity these test cases should be thought of as part of the debugging process for performance orientated end-user code. The results produced from benchmarking can be related to all manner of physical specifics about the state of the computer as it runs various tasks specified by the developer. These specifics include: compiler used, GPUs, Operating Systems, CPUs, disk drives, etc. Captured results from these subsytems are stored online and can be visually depicted using standardized, intuitive layouts. 
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+From the Phoronix Test Suite Guide:
```
seantalts: Feel free to cite their guide for the how-to. Here we also want details on how this would integrate with Stan.

syclik:
I just read through it. I don't really see this as requiring a design document at this level unless I'm missing something. It looks like this was meant to augment our existing testing. I think it'd be good to discuss the specifics of what testing would be appropriate for phoronix and how reliable it is in providing these results. 

I guess I'm thinking should go the other way. We should define what sort of testing we want or require (maybe as a design doc) and then pick things that would help accomplish that? Does that make sense?

To be concrete, we could discuss how this would work for benchmarking Math. If it makes sense, we should be able to include since that would only affects Math; and in a sense, it doesn't actually affect design... this is additional instrumentation around the code base. (To rephrase: I'm all for it! Conditioned on it being low maintenance and reliable.)

bob-carpenter:
What we want is to avoid performance regressions in the math library.  How to achieve that within our computing and maintenance budget is up for debate.  Even how to test performance given our cross-platform support and probabilistic targets is going to be tricky.

ode33:
If we know when to expect a performance regression there are more efficient developer orientated tools than phoronix to isolate where it is being introduced. 

When the developer and review process overlooks some unknowable and a regression is introduced what is the current detection procedure and how might it be improved?

ode33:
A blunt hammer outline might be to have a baseline benchmark on the current versioned release and have it that the developer is responsible to replicate relevant test cases for their PR as part of the review process. So we know we aren't thoroughly testing everything within the scope of a change but we're making sure some bundle of representative models still performs equivalent or relatively well compared to the current release for the end user. 

As Bob mentions there is a lot of detail involved and the accuracy requirements and an expanding administrative scope for what may need to be tested cannot be easily defined in advance. Certainly not for me. On the other hand the phoronix online forum is a presently untapped resource as are the software developers or maintainers themselves. I'm not comfortable reaching out on behalf of the project though an anonymous post on the forum could help better outline some of this.

PR #5, opened 2019-05-05
adding parallel_autodiff RFC

wds15:
This RFC will enhance parallelization facilities in stan-math.

At the core it does implement

- parallel for with automatic sharding
- parallel reduce with automatic sharding and vectorization

To get there a few things need to be refactored and introduced

- independent AD
- refactored nested AD
- Intel TBB is proposed to implement the parallel algorithms

Comments are most welcome.


bbbales2:
Moved from Discourse:

```
The suggested building blocks per se only parallelize the *forward*
  sweep of reverse mode AD, but not the reverse sweep. Thus, the
  operations which involved `adj_jac_apply`, for example, are not
  speedup at all with this. However, this can easily be resolved by
  introducing a general AD tree compression facility which works just
  as `map_rect` does. Such a facility needs to know all the operands
  and can then do nested AD and perform on the spot gradient
  evaluation of the function of interest and store results as
  `operands_and_partials`.
```

What's the difficulty of adding mechanisms to parallelize the reverse sweep? Is there any way that this refactor might make it easier to do that reverse sweep? Does this refactor in any way hamper our ability to do the parallel reverse sweep? Or is that tangential?

designs/0003-parallel_autodiff.md
```diff
@@ -0,0 +1,701 @@
...
+  topological sort of the autodiff tree whenever that is permissible.
+- We are going from one global ad tape to many small ones whenever
+  parallelism is used. The memory pool so far was grown only once and
+  then stayed static in size. With many small ad tapes we will merge
+  the independent ones into the originating one and when doing so only
+  the ownership of the memory blocks will be transferred, but at each
+  re-evaluation we will re-allocate the memory again. This can be
+  addressed by a more clever memory pool. My suggestion is to monitor
+  the performance penalty incurred by this. However, note that
+  programs not using parallelism will not have any changed behavior.
```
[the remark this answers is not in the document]
wds15: In all honesty - I am not considering myself a wizard of multithreading at all (despite the fact that I am kicking this off). This is why I am so strongly suggesting to go with the Intel TBB. A ton of knowledge on that topic is sitting right there to be used. For multithreaded applications I learned from the TBB doc that allocators need to  be *scalable* and *cache aligned* in order to be fast. Please refer to the [Intel TBB doc](https://www.threadingbuildingblocks.org/tutorial-intel-tbb-scalable-memory-allocator) or the [Intel TBB ref](https://software.intel.com/en-us/node/506255?_ga=2.63533228.2071084492.1557255990-808285549.1557255990).

The material from Hoard should also be good to follow along.

wds15:
> What's the difficulty of adding mechanisms to parallelize the reverse sweep? 

I do not see any difficulty in running the reverse sweep in parallel. In fact, it will probably even be simpler since the AD tape is fixed at that point in time such that one does not need to worry about synchronization wrt to building the tree.

> Is there any way that this refactor might make it easier to do that reverse sweep?

I haven't thought too much about this, but the answer could be yes if we go with a more involved design which I wanted to avoid for the moment. So *if* we would keep the independent AD tapes separate (which would improve performance of the forward parallel sweep as well), then I could imagine that the reverse sweep can take advantage of that. So should we keep the independent AD tapes separate, then that structural information should be possible to take advantage of as *building* independent AD tape pieces should allow us to traverse these again in independence on the reverse sweep. This would be super cool! However, I am not sure if we can manage the dependency mess and my preference is to start safe and simple (but maybe we have an enlightening moment on this).

> Does this refactor in any way hamper our ability to do the parallel reverse sweep? Or is that tangential?

No, it does not do that, I think. My thinking so far was to do AD tape compression which results in forward and reverse sweep being done on the spot. This on the sport AD tree evaluation has CPU cache advantages, but as raised above it could be that we can do something even smarter which I did not have on my mind yet. I can give it some thinking, but I would be reluctant to commit to this - the door is certainly not shut for parallelizing the reverse sweep; not at all. This PR can make independence in the AD tree a structural feature of the AD tree and we can likely use this.

bob-carpenter:
The trick is going to be figuring out how to run things in parallel.  The expression graph can be topologically ordered and it can then be divided up into subgraphs that preserve this order.  We're hoping to do that with the new compiler through static analysis.  I don't know how we could do it
in parallel without heavy synchronization otherwise, as we need to make sure that a parent node's adjoint is fully computed before propagating down to children nodes (otherwise we lose dynamic programming and will probably get the wrong results).

bob-carpenter:
> It looks like it uses per thread local heaps and one global heap that the local heaps can give back memory to for reuse 

With autodiff, our memory gets allocated on a stack and can be collected all at once after the reverse sweep.  Our thread local autodiff data structures are structured local memory (stacks, not heaps), so this is similar to what we do with `map_rect`. Then we copy back into the global structure.  If we could somehow preallocate and only ever use the global structure rather than copying, I think it would be even more efficient, as the global structures can proceed without synchronization.  The only thing that needs to be synchronized is that we need to compute subexpressiosn before the expressions in which they're contained.

wds15:
@bob-carpenter I am not sure what you refer to with 

> I don't know how we could do it
in parallel without heavy synchronization otherwise, as we need to make sure that a parent node's adjoint is fully computed before propagating down to children nodes 

In the proposed PR the called function parallel for and parallel reduce *know* that things are independent from one another as this must be guaranteed by the user. So we don't have that problem of being unsure of this - at least in the context of this PR, I think.

bob-carpenter:
> I am not sure what you refer to

Running the reverse sweep in parallel.  It has to at least be synchronized so that parent adjoints are computed before child adjoints.

wds15:
@bob-carpenter Yes, that makes sense. The reverse sweep is hard, but maybe one has a chance given that we will be able to know what was independent in the forward loop; I am not sure at all though. Essentially, what did run in parallel in forward sweep can possible be run again in parallel on the reverse sweep. Maybe.

bob-carpenter:
> On May 7, 2019, at 4:13 PM, wds15 <notifications@github.com> wrote:
> 
> @bob-carpenter Yes, that makes sense. The reverse sweep is hard, but maybe one has a chance given that we will be able to know what was independent in the forward loop; I am not sure at all though. Essentially, what did run in parallel in forward sweep can possible be run again in parallel on the reverse sweep. Maybe.

If it can run parallel forward, then it can run parallel in reverse.  That's what I meant by it needing to be synchronized, though.  It's not possible in general, just in some special cases, like map_rect or maybe the parallel for loops.

designs/0003-parallel_autodiff.md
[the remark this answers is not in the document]
SteveBronder: Word! I think your intuitions right that this should be secondary atm.

yizhang-yiz:
> The trick is going to be figuring out how to run things in parallel. The expression graph can be topologically ordered and it can then be divided up into subgraphs that preserve this order. We're hoping to do that with the new compiler through static analysis. I don't know how we could do it
in parallel without heavy synchronization otherwise

This sounds like graph partitioning that PDE solvers have been utilizing, using libraries such as METIS or SCOTCH. This could be a much better uniform solution.

yizhang-yiz:
> The trick is going to be figuring out how to run things in parallel. The expression graph can be topologically ordered and it can then be divided up into subgraphs that preserve this order. We're hoping to do that with the new compiler through static analysis. I don't know how we could do it
> in parallel without heavy synchronization otherwise, as we need to make sure that a parent node's adjoint is fully computed before propagating down to children nodes (otherwise we lose dynamic programming and will probably get the wrong results).

This sounds like graph partitioning that PDE solvers have been utilizing, using libraries such as METIS or SCOTCH. This could be a much better uniform solution.

bbbales2:
@wds15 cool beans, sounds like this isn't limiting us as far as reverse mode, so as long as that's true, grooovy.

betanalpha:
Ultimately if there's no performance hit for reverse mode with this underlying change in architecture I don't have any strong opposition to moving in this direction to facilitate future parallelization opportunities.

That said, I think the prioritization of forward mode is a big misguided and wouldn't bear significant fruit for some time.  As Bob notes the reverse sweep inverts the graph dependencies which requires a careful synchronization for any parallelization to be exploited.  Without a roadmap for how to accomplish this synchronization the possible benefits of the new architecture will have limited scope for the foreseeable future.

This proposal also leapfrogs past support for parallelizing _within_ each node, either for constructing the Jacobian or computing Jacobian-adjoint products, which is more inline with the recent map_rect and GPU efforts.  It would be great to have stronger motivation for why these are insufficient for the time being.  In particular, tools that facilitated building "super nodes" with internal parallelization of the Jacobian or (Jacobian/Jacobian transpose)-vector products would allow for advances that immediately apply to forward and reverse mode (at least at first-order) without changing the entirety of the autodiff architecture.  It would also be easier to specify the available parallelization opportunities within the context of each node instead of sharding the entire expression graph.

It would be great to have more compelling motivation for this particular architecture design over, say, incremental improvements in within-node parallelization.  For what kinds of expression graphs would it be most effective, for which would it be least effective, how would it generalize to reverse mode, etc, etc.

Also in the Discourse there was some discussion about higher-order autodiff.  The current recursive implementation and the alternative implementations I discuss in the recent higher-order paper would benefit from the same expression graph decomposition/parallelization opportunities as first-order forward and reverse mode.

wds15:
Thanks for your feedback, @betanalpha 

Let me try to answer your points in order:

- **Why do I leapfrog within node parallelisation?** Well, we have something like this already attempted and I looked into it in some detail: This was the openMP PR which we discarded a while ago. The main issue I saw is that we very often do clever accumulation across the loops which do the main work within nodes (and openMP does not work on macOS clang). Parallelising these for loops then means that you have to maintain thread-local sums which means that we (1) loose exact reproducibility almost for sure and (2) one would have to significantly refactor the entire code base to handle this, I think. Furthermore, I do think that it is far more fruitful to go along the lines I am proposing, since much bigger chunks of work is parallelized. Also: This proposal does not disallow parallelising a given node. You can still do that (and in fact the Eigen operation can be paralleled already now if you turn on threading for Eigen; for example with the MKL).

In summary:

- The proposed solution will speedup almost any Stan program - not just those with specially crafted nodes
- This PR speeds up large chunks of nodes in a single go. Thus, we can also speedup models where a given node is just too small for speedup, but you need to parallelise the joint operation of many nodes => better scalability for a larger set of models
- This PR does allow you to speedup nodes easily. In fact, the way I am proposing to write the parallel for and reduce will give you immediatley the tools to very conveniently code up within node parallelisation. As long as a parallel for or reduce-sum will do the job we can make every node use these basic utilities (without having to deal with parallelisation themselves) => we can prepare our code base such that within node parallelisation is done for the user automatically in a way which is safe for developers since the parallelisation logic is taken care of

- **Parallelising the reverse sweep?** Well, this is not out of scope at all. With this design we will know what elements were run in forward mode in parallel and we can actually maintain that information without too much effort. Once we do that, then the reverse sweep can also run in parallel. I did not think that this is so critical when doing this such that I opted for a simpler design - but it is not too hard to develop this further into something which runs the reverse sweep in parallel as well.

I hope that answers your questions. Let me know if I missed some points or things are not clear.

yizhang-yiz:
> openMP does not work on macOS clang

If one googles "macos clang openmp" the first entry would be on how to turn on openmp switch for macos clang, which is what I've been using.

bob-carpenter:
The reverse sweep has different parallelization dependencies.  Suppose we have

u = ...
v = ...
y = f(u, v)
x = g(u, v)

In the forward sweep, f and g wait for both u and v to execute, then can execute in parallel without synchronizing access to u and v (modulo synching the expression graph).

In the reverse sweep, f only waits for y and g only waits for x, but they will both modify u and v's adjoints, so cannot be executed in parallel without synchronization.

betanalpha:
I am hesitant about how easy it will be to generalize this architecture 
to efficiently parallelize the reverse mode sweeps.  The context of
a parallel for or reduce defines the dependencies in the forward
sweep of the expression graph, not the backwards sweep.  As
has been previously noted this will at the very least require some 
graph analysis to figure out the synchronization for when things 
can be run in parallel.

In that vein I think that it’s a little too speculative too say that this
will speed up any Stan program!  Firstly that doesn’t specify what
calculation the Stan program is being used for (everything in the
current algorithm suite uses reverse mode).  Secondly any speed
up depends on the dependencies of the model and whether the
model has been written in a way that communicates the parallelization 
opportunities.

Following up with that I find that it’s much easier to identify and
utilize parallelization opportunities within specialized nodes.  As
you note you have to specify the context of a group of nodes
anyways, and this is already well suited to language constructs
like vectorization which define specialized super nodes.

A focus on meta programming tools that facilitate parallelizing
operands and partials implementations (i.e. Jacobian calculations)
or the newer Jacobian-adjoint methods would allow for parallelization 
of these super nodes (algebraic solver, ODE solver, vectorized
densities, matrix calculations, etc) without changing the architecture
of the entire autodiff system.

In my opinion the practical benefits for the architecture redesign 
are speculative at the moment, especially in the case of reverse 
mode, and that makes me hesitant about such drastic changes
relative to other approaches.  Again just my opinion!

yizhang-yiz:
> A focus on meta programming tools that facilitate parallelizing operands and partials implementations (i.e. Jacobian calculations) or the newer Jacobian-adjoint methods would allow for parallelization of these super nodes (algebraic solver, ODE solver, vectorized densities, matrix calculations, etc) without changing the architecture of the entire autodiff system.
I completely agree. These are easier to grasp in terms of performance guarantee.

yizhang-yiz:
IMO a real (semi) silver bullet for parallelization should be what @bob-carpenter pointed out: synchronization + (static) graph partitioning. This might be similar to what people do sparse linear solver: in so-called domain decomposition, one decomposes the spare matrix graph and mark the neighbors, so that each sub-graph(sub-domain) becomes independent when it's boundary nodes are updated by receiving/sending data from/to neighbors. The algorithm has same ingredients(synchronization & partitioning) as we face here. This could relieve us from manually utilizing the proposed reduce op and work for both forward and backward mode.

wds15:
> If one googles "macos clang openmp" the first entry would be on how to turn on openmp switch for macos clang, which is what I've been using.

I am aware of that you can get openMP to work on macOS...but it is not part of the stock Apple clang.

Apart from that I do find using the Intel TBB much better suited for stan-math as it is a C++ library. It has the advantage that the building blocks blend well with our codebase as opposed to compiler extensions.

@bob-carpenter + @betanalpha Well... parallelizing the reverse sweep was nothing I had on my radar initially. I purposefully left this out and in the draft I am suggesting to do AD tree compression just like map_rect does it. I would much prefer to move on at the moment, but keep this in our heads for later. As I say, the structural information can be contained of what did run parallel in the forward sweep... and modifying the adjoints in the reverse sweep could be done using atomic updates since the addition operation can be done using atomics as I am aware. How much you pay in terms of performance for the atomicity...don't know.

@yizhang-yiz I am not understanding what you are saying - should we do this PR; should we NOT do it; should we do something else? If something else, then it would be great to have something feasible in finite time. I also do not think that the graph things you point out would be ruled out by this PR here as proposed. That can probably be combined. So please give constructive comments on "Should this PR be implemented?".

@betanalpha I do think that a very large class of Stan models will benefit. How large remains to be seen, sure (and it won't be any program). The empirical evidence which made me conclude that is that a Poisson log-likelihood with just $10^4$ terms can be speed up by a factor of 2x (on 2 cores). This made me very optimistic about things.

I also do not think that what is proposed is such a huge change in fact (aside from nested AD, OK). So we generalize a bit of what is the AD tree (multiple tapes now) and what is allowed (independent things can be permuted). The nested AD is getting refactored to something which is a lot safer to use in the form of this PR. From the API perspective all what is new is the `ScopedChainableStack` - the rest stays as is.

yizhang-yiz:
> I am not understanding what you are saying - should we do this PR; should we NOT do it; should we do something else? If something else, then it would be great to have something feasible in finite time. I also do not think that the graph things you point out would be ruled out by this PR here as proposed. That can probably be combined. So please give constructive comments on "Should this PR be implemented?".

I think you are mistaking RFC as PR review. When providing comments in a "request for comments", I'm not at any position to answer "should this PR be implemented" while conducting discussions.

> The empirical evidence which made me conclude that is that a Poisson log-likelihood with just $10^4$ terms can be speed up by a factor of 2x (on 2 cores). This made me very optimistic about things.

I wouldn't say a 2-core run warrants "very optimistic" impression. Again since we are in an RFC allow me to point out that laying out a solid groundwork improves the chance that an RFC being accepted. I remember @seantalts have provided more detailed information on the process.

wds15:
Sorry... `s/PR/RFC/g`... and, of course, this is not a PR which leads to merging of the prototype which I put together.

And don't get me wrong - the graph things you propose do make sense, but to me that is very remote to the discussion of this RFC; unless you think this impacts it or you are concerned that this RFS is somehow interfering with that. I do not have the impression such that I think it is a good topic for a discourse discussion to grab the attention of stanc3 people who need to consider this if I get this right.

Maybe @seantalts can comment, but as I understand the intent of this PR to the design repo of this RFC is meant to align everyone on the design, it's concepts and get agreement on the forthcoming PRs to stan-math. In that regard any comment is welcome.

Also: If the benchmarks I am quoting are not good enough and we conclude that we need more of them. Then we need to settle on what to look into, etc.

designs/0003-parallel_autodiff.md
```diff
@@ -0,0 +1,701 @@
...
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Why is this design the best in the space of possible designs? I
+  think the approach is incurring minimal changes to Stan in order to
+  make it work with the TBB. The key is to refactor nesting and
+  introduce `ScopedChainableStack` in order to allow for local execution
+  contexts. The guiding principle is to have independent AD tapes used
+  in local contexts. This fits nicely into the TBB, but the introduced
+  abstraction is useful regardless of the TBB.
```
rok-cesnovar: Not sure there is any need to defend the choice of TBB but an added plus of Intel TBB is also that it works nicely with Intel MKL that we can already use to speedup Eigen. The same is true for OpenMP, but MKL+TBB is far superior - https://software.intel.com/en-us/articles/using-intel-mkl-and-intel-tbb-in-the-same-application.

This is a nice comparison on when one would chose TBB, OpenMP or native threading: https://software.intel.com/en-us/intel-threading-building-blocks-openmp-or-native-threads

One other option that is gaining momentum is SYCL ([example of a parallel for](https://www.codeplay.com/portal/sycl-tutorial-1-the-vector-addition)), but it suffers from the same issues as OpenMP in that it needs compiler extensions. And above all its still very very very young and doesnt have a lot of runtime support. 

My suggestion would be to extend this paragraph to cover all the alternatives and just write down why they were not chosen. There are going to be developers that come in and will ask "this is great but have you heard of XYZ" and we can just point them to this document. If you agree with my suggestion, I can help with this.

wds15: The empirical evidence you point sounds good to include and I certainly don't see any reason why not to... so if you would like to add that, we can; sure.

We should certainly note that assessments from Intel are possibly biased; so that should be pointed out openly (although the docs you link is not any marketing gaga, of course).

What would you suggest to add the comments? Should I grant you permissions to the repo?

rok-cesnovar: I am going to make a PR on your fork.

betanalpha:
> I am not understanding what you are saying - should we do this PR; should we NOT do it; should we do something else? If something else, then it would be great to have something feasible in finite time. I also do not think that the graph things you point out would be ruled out by this PR here as proposed. That can probably be combined. So please give constructive comments on "Should this PR be implemented?".
> 
> I think you are mistaking RFC as PR review. When providing comments in a "request for comments", I'm not at any position to answer "should this PR be implemented" while conducting discussions.
> 
> The empirical evidence which made me conclude that is that a Poisson log-likelihood with just $10^4$ terms can be speed up by a factor of 2x (on 2 cores). This made me very optimistic about things.
> 
> I wouldn't say a 2-core run warrants "very optimistic" impression. Again since we are in an RFC allow me to point out that laying out a solid groundwork improves the chance that an RFC being accepted. I remember @seantalts <https://github.com/seantalts> have provided more detailed information on the process.
> 
My understanding of the point of a RFC is in alignment with Yi’s,
hence my more questions about motivations, priorities, and 
alternatives.  We want to go through this discussion now before
too much time is spend on code that might not end up being
accepted.

In any case I think it might be helpful to step back at this point
and review some of the higher level possibilities.

Our current automatic differentiation system proceeds in two
stages.  In the first stage the expression graph is built up by
computing values at each node, connecting nodes, and, in 
most cases, eagerly evaluating the Jacobian of partial 
derivatives at each node.  In the second stage tangents
are propagated forwards in forward mode and adjoints
are propagated backwards in reverse mode.

There are opportunities for parallelization in both states.

In the first stage the value and any eager partial calculations
can be parallelized in the appropriate contexts.  This is
particularly opportune for “super nodes” that contain many
inputs and many outputs, such as the algebraic solver, as
well as constructs like parallel reduce.

In the second stage the propagations could be parallelized
in certain contexts, as could any lazily evaluated Jacobians
or Jacobian-tangent/Jacobian-adjoint products.  The
parallelization of the forward mode propagating would
follow the same pattern as the expression graph building
parallelization, but the reverse mode propagation is much
more subtle as we previously discussed. 

If the Jacobians have been eagerly evaluated then the
propagation is pretty cheap — just a series of multiplies
and adds.  Because the overall cost is typically dominated 
by the more expensive value and Jacobian calculations 
then I don’t think that there will be much speed up here.

In my opinion the biggest opportunity for useful parallelization
is in the parallelization of the value and Jacobian calculations.
If the Jacobians are computed eagerly then this can sometimes
be done with parallelization in the building of the expression
graph, although care would have to be taken to ensure 
reproducible results (for example creating node pointers
first before dispatching the parallel calculations with 
addresses for where the thread should save the outputs).
It can also be done within the nodes themselves.

If the Jacobians are computed lazily, which is ideal as we
move towards more direct Jacobian-adjoint products in 
reverse mode, then parallelization is trickier across the
entire expression graph (requires synchronization) but
still straightforward for within-node parallelization of the
Jacobian-adjoint products.

The other critical factor in all of this discussion is what
scope of models do we think will be popular going forwards.
For example if everyone was writing models with IID 
densities in for loops then Sebastian’s Poisson example
might be relevant.  Most users, however, rely on the efficient 
vectorized form of the densities which wraps all of the 
calculation within a single node that would suggest
prioritizing within-node parallelization.

In my opinion everything points to focusing on within-node
parallelization, in particular parallelizing adjoint-Jacobian
product calculations in functions with many inputs, such
as the ODE solvers, vectorized densities, and linear
algebra functions.  

My fear is that significant changes to the current architecture 
will cause too much indirection which will slow everything 
down save.  I don’t think that the relatively small number of 
applications that could overcome the overhead with the
parallelization opportunities afforded by this framework 
warrants that risk, at least not without additional studies.

Anyways, I hope that helps move the discussion forwards.
Get it?  It’s autodiff joke.

bob-carpenter:
Is the parallel for something to be used in user programs or is the
purpose what Michael's suggesting in parallelizing our function and
adjoint-vector calculations?  User programs bring up issues with
dealing with side effects like assignment.

Thanks, Michael.  That was super helpful in laying out the tradeoffs
between eager evaluation and parallelization.  I keep forgetting that
reverse mode isn't parallelized yet, so all those lazy derivative
calculations that are useful for cutting down on memory, also reduce
parallelization in map_rect.

The scalar functions are mostly lazy, but they typically have simple
calculations and reuse subexpressions.  The matrix functions are
largely eager, but some of the critical ones have been done more
lazily.  All the probability functions using operands_and_partials
are eager.  

An even worse case is functions written purely in templated form,
which create large subexpression graphs.

yizhang-yiz:
> Is the parallel for something to be used in user programs or is the purpose what Michael's suggesting in parallelizing our function and adjoint-vector calculations?

Following the two stages argument, IMO we should make parallelization for the 1st stage user-faced, I don't see how we can make the 2nd stage accessible to user.

designs/0003-parallel_autodiff.md
seantalts: I think this is a great idea even if the intel docs are a little biased.

wds15:
To start with autodiff humor... we are very much on a backward sweep here, I think... and I am not sure if we are working with an efficient implementation here...

Moreover, I have been discussing these ideas already on discourse a few times. So the opportunity to challenge them at this deep level is - in all honesty - from my perspective already over. There is a lot of work which went into all of this already; and this is being completely dismissed here... but ok, you don't care about that, I see... and your points are good ones, so to your points @betanalpha:

Right now a lot of stuff in our library evaluates eagerly on the forward sweep. All of these codes will benefit directly from the constructs which are proposed here as the work is done during the forward sweep.

Super-nodes are being created right now for various special cases - and these are great and useful; they also promise huge speedups on the GPU, etc. However, when we develop models we need flexibility and my guess is that a very large model class is being developed without the additional restrictions of super-nodes. The specific super-node one needs is likely just not yet there (and getting a new super-node into stan-math when you need it is not a feasible option to a modeller)

As conditional on parameters just about any likelihood can be chunked into IID parts, the proposed scheme makes sense for these models - and this is a very large class of models - just by the mere flexibility as opposed to super nodes, this is much more powerful. Also, as the poisson log-lik example shows, the tools in this RFC essentially give the user the ability to create super-nodes himself (since most of our stuff is eager right now).

If we stay for the moment with doing gradients, then the AD graph compression as done by `map_rect` provides also a means to parallelize the revere sweep as well. While this RFC does not attempt to do that, I think that we can plug such a graph compression thing into this framework. Doing so will trigger the reverse sweep on the spot and eagerly calculate the partials of the operands passed into such a construct. This construct would be a little bit more complicated as it needs to know the operands. Moreover, this pattern is likely inefficient if you do multiple passes over it - which is not needed for gradients.

However, in any case - what is proposed here is completely complementary to super-nodes! In the form proposed, the data structure for the AD tape doesn't even change at all.

In addition, I would very much think that the constructs proposed will make it a lot easier to create parallelized super-nodes.

I would suggest to look into the examples people have come up with on the forum where we saw the failure of `map_rect` to speed them up. These are examples which are relevant to our users obviously and speeding up these models should be in the scope of this work.

> Is the parallel for something to be used in user programs or is the
purpose what Michael's suggesting in parallelizing our function and
adjoint-vector calculations?  User programs bring up issues with
dealing with side effects like assignment.

The intention of this RFC is to introduce constructs which can be used inside stan-math to create super-nodes and to expose these constructs in a safe way to users. I am targeting as driving use case the user-facing part of it, but I do think that these are useful things in themselves - however, if the constructs themselves are not fit for every purpose inside stan-math then that is also ok - as in that case you can simply use some cool stuff from the Intel TBB which is a modern C++ library centered around parallelism.

designs/0003-parallel_autodiff.md
```diff
@@ -0,0 +1,701 @@
...
+    vector_d iarg(2);
+    iarg << 5, x_ref_2(i);
+    res(0) = f(iarg);
+    return res;
+  };
+
+  // the concatenate_row takes the returned std::vector<vector_d>
+  // and appends them horizontally together
+  vector_d parallel_result = stan::math::concatenate_row(
+      stan::math::parallel_map(count_iter(0), count_iter(num_jobs), loop_fun));
```
rok-cesnovar: So this will be something similar to [std::for_each](https://en.cppreference.com/w/cpp/algorithm/for_each) in C++17 right and could potentially be replaced once (if) we decide to move to C++17?

rok-cesnovar: Ah, you mention C++17 in the rest of the paragraphs, ignore me :)

rok-cesnovar:
> what is proposed here is completely complementary to super-nodes!

I think this is the largest point here. My understanding of all of this is that this does not break any other future plans or introduce any regressions. I have offered my help with this, so has @SteveBronder and we have also had a recent influx of developers that are interested in performance and they could potentially help out with parallelizing super-nodes or what this RFC is proposing. Both with the common goal of increasing performance. @wds15 is interested in speeding for these types of models and he is (hopefully still) willing to take charge.
What this RFC is proposing isnt going to get built overnight and speeding up super-nodes can happen in parallel. (a parallelism "joke" there) Are there any discussions/issues/wiki out there on paralelizing any of those super-nodes (algebraic solver, ODE solver, vectorized densities, matrix computation, etc..)?  

My personal view here is that something like this will need to be done eventually or we will be missing out on a large opportunity for easy to get speedups. We will also be ready to plug-in C++17 parallel STL once we get there.

In the super long run it also opens doors for experimenting with things like SYCL (C++ for heterogeneous architectures) that uses a similar style as TBB/parallel STL and is going be supported in Clang 9 and will eventually find its way in the C++ standard.

designs/0003-parallel_autodiff.md
wds15: C++17 parallel is nice... but clang and gcc anyway will use the TBB as the foundation to build up C++17 parallelism. To me the TBB is the better API to go with as it is more advanced. The C++17 feels like a v1.0 things with many things missing.

But to your comment: yeah the parallel_map is like a parallel for_each; with more control over the execution (for example, the blocked range concept of the TBB is nice, I think).

wds15:
I am still willing to work on this...given how this is evolving it will take a while until we get to agreement.

BTW, there is a fully working prototype, so we should probably do one more study to test the scaling a bit more broadly. I am just hesitant to pour my time into this given all of this headwind here. Also, I haven't seen a comment yet from @syclik ... is more time needed for studying this or is a meeting needed?

wds15:
Here is an independent evaluation of a few parallel techniques:

https://arxiv.org/pdf/1302.2837v2.pdf

TBB wins that one out.

seantalts:
Hey all! Just dropping in with a point-of-order note: It would be really useful for the reviewer (@syclik still right?) to summarize where we are and what needs to be researched before a decision can be made, if anything. I think the discussion has gone on long enough :) I'll take a stab at it for now just to try to add a little structure while I'm sitting at a café:

To me it looks like the questions were mostly around whether this disallows other kinds of parallelism, and it doesn't seem to, so we're maintaining optionality there. Of course there's a very valid language-design discussion to be had on the best way to bring parallelism to users of a domain-specific language, but I don't think we'll solve that problem in this PR :) I've been involved in that conversation personally for about 5 years now as compiler implementer in this and a couple other languages and it's certainly a nuanced discussion, but because we're maintaining optionality we don't have to decide with this design doc (and to a large extent we already started forging down the user-specified parallelism path with map_rect though obviously it needs a better API). So I think we can drop that issue w.r.t. whether this design is a good one for what it addresses.

There's a very valid concern that this could make single-threaded performance worse. To address this concern, I think we could benchmark that and figure out a way to code around it if that ended up being true. So to me this looks like it needs a little bit of research (or for the results of those benchmarks to be posted here).

So @wds15 if you have code for this already and want to post some benchmarks showing that it doesn't hamper single-threaded performance that might be a nice bit of evidence for the design. That said, we'll of course actually benchmark any code change PR that ends up getting put out to make sure of that whenever it's ready.

wds15:
The way I have planned this right now let's the single-core performance where it is if you leave out nested AD operations (I get to those below). With the current proposal the AD tape data structure is not changed. Only during the parallel execution phase we will have multiple AD tapes which are stichted to a single AD tape after the parallel execution phase.

In my mind I would start with having the usual `STAN_THREADS` on/off distinction. We may want to revise that at some point (and turn `STAN_THREADS` always on), but I do not see a need to make that call now. Turning it on does mean ~5% on average performance loss (on our ref set of models), but we do not have to bite that bullet now.

So maybe a few scenarios under `STAN_THREADS` being off:

- if you do not use parallel reduce or map then nothing changes
- if you use parallel reduce or map, then still nothing changes as we only use a single core
- non-refactored code which uses nested AD in an iterative way (ODE & algebra solver) will be seriously slowed down...but with a very easy refactoring using the new concepts that slow-down vanishes; so no speed is lost - I have done that with the SIR ODE example and can share benchmarks. This wasn't really what you asked for, but if it is helpful, I am happy to post those.

If you turn now `STAN_THREADS` on:

- if you don't use the parallel functions you still endup with the same AD strucutre
- with the current design the AD structure is only different during execution of the parallel commands during the forward sweep
- the simple design put forward involves merging together the independent AD tapes to a single one such that the AD structure is unchanged (only some permutations are happening) even after the parallel functions
- for the parallel reduce function the merging to a single AD tape should be unproblematic as the reduction should usually lead to very small AD tapes (if vectorized nodes are used)
- for parallel for and reduce operations without vectorization we may consider to have the independent AD tapes stay that way which requires linking those together. This is a possible optimization which would complicate the AD tape structure somewhat.

And yeah... development of super-nodes which use parallelization can happen independently (and in parallel, of course).

At least at the moment I don't see any single-core degradation in performance since nothing changes by design (except for nested AD). However, I cannot rule out that we will reach problems where we need to either do ugly ifdef things to code for single and multi-threaded cases or we decide to change the design to better accommodate multi-threaded performance at the cost of some single-core performance.

A hard choice which I would suggest to take (and it does affect other parallelization efforts) is the use of the Intel TBB as basis for all of our parallelization - but even that decision could be deferred if we really wanted to (meaning that what we use for other parallelization things may differ). So I would tend to go with using the TBB as a default for parallelization, but even that can be kept optional since the TBB is designed to be a add-on parallelization framework.

betanalpha:
> To me it looks like the questions were mostly around whether this disallows other kinds of parallelism, and it doesn't seem to, so we're maintaining optionality there. 
> 
To be clear I don’t think that this proposal obstructs other
approaches to parallelism. 

My point is rather that other approaches to parallelism can 
yield many of the same benefits without affecting the 
underlying structure of the autodiff code and affecting 
non-parallel execution or requiring the maintenance burden 
of dual implementations awkwardly separated by ifdefs.

In my opinion the difference in how the map_rect code
and the threading code have evolved make the relative
trade offs pretty evident, ultimately favoring map_rect 
approaches.

wds15:
We already have a `STAN_THREADS` ifdef and I would likely not need to introduce any more of these. Given that the TBB is written with *optional* parallelism in mind we should be able to code things up in a way such that in absence of `STAN_THREADS` the Intel TBB is instructed to only use a single core... which is what we essentially already do in the `get_num_threads` function.

So no additional ifdefs should land in our code (I will attempt to avoid that, of course... but I can't rule it out either).

bob-carpenter:
Counting ifdefs is just a weak proxy for the complexity of the code.   I'd be much more concerned about induced dependencies.   See below.

> On May 25, 2019, at 4:20 AM, wds15 <notifications@github.com> wrote:
> 
> We already have a STAN_THREADS ifdef and I would likely not need to introduce any more of these. Given that the TBB is written with optional parallelism in mind we should be able to code things up in a way such that in absence of STAN_THREADS the Intel TBB is instructed to only use a single core... which is what we essentially already do in the get_num_threads function.

This sounds like the TBB is going to be involved even in non-threaded code.  Is that right?  That's what I mean by dependencies in the code.

wds15:
Yes, the intention is to use the TBB even in the non-threaded version of our code - otherwise we would have to write the parallel reduce and parallel for functions proposed in two flavours (with and without threading turned on).

To start with the TBB would be used in very few localised places, but in the long run I think we should see the TBB just like boost. The TBB will make writing super-nodes much easier as that thing is meant to be added to a serial code-base and turn things into parallel code by using their *threading building blocks*.

So the TBB would show up to start with in the parallel* functions and maybe very few more places. Things would not run without the TBB. However, everything is modular so that you could rip out the TBB if you really wanted to.

seantalts:
Do you all (esp. @syclik in the reviewer role) think it would be helpful to split this design up into (at least?) two parts:

1) Intel TBB (which maybe doesn't need a full design doc but should have recorded answers to[ these questions](https://github.com/stan-dev/math/wiki/Checklist-for-adding-a-dependency-to-the-Math-library) somewhere)
2) Changes to Math autodiff
3) proposals for parallel_for and _map functions for Math and/or the Stan language

I feel like in some cases the later steps would be the only motivation for the earlier steps; but I think here we already have compelling use cases for 1 and 2 without looking at 3, right?

wds15:
Those questions for the TBB are answered [here already](https://github.com/stan-dev/math/wiki/Dependency-Checklist-for-the-Intel-TBB).

I thought that seeing everything together makes it easier to understand as you see the complete picture and how things fit together. We can do whatever is easiest.

This RFC is taking really long (4 weeks now) and I do have already very convincing benchmarks with a prototype. To further evaluate the prototype and to make it possible for others to run this with stan models I will probably do a fork of stan-math and start tagging useful versions as well. Maybe this way users willing to work on bleeding edge code can provide feedback so that we find out what options should be given to users. By playing with the code I have some intuition on this already, but confronting code with real models is the best way to know whats good and what not.

wds15:
The concatenation is per se not needed as the parallel map loop returns

`std::vector<f(*iter)>`

It's just that this output can be a ragged array so that it would need to be concatenated to be valid in the "Stan" world of containers. However, for the C++ world of things this is totally fine as a runtime return of a function I think. In that sense I don't quite see the need to a compile time concatenation.

In a way the parallel reduce is a compile time concatenation where we know that we want to do the sum. In that case we make this explicit by using the reduce algorithm of the TBB which allows us to speed things up enormously and it let's us keep the AD tree small which will only grow in size proportional to the number of reduce splits of the incoming data.

So I think it makes sense as proposed and we will not be able to combine the parallel map and parallel reduce by having a compile time thing for concatenation (the TBB use two distinct algorithms for that for a reason).

syclik:
> Do you all (esp. @syclik in the reviewer role) think it would be helpful to split this design up into (at least?) three parts:
> 
> 1. Intel TBB (which maybe doesn't need a full design doc but should have recorded answers to[ these questions](https://github.com/stan-dev/math/wiki/Checklist-for-adding-a-dependency-to-the-Math-library) somewhere)
> 2. Changes to Math autodiff
> 3. proposals for parallel_for and _map functions for Math and/or the Stan language
> 
> I feel like in some cases the later steps would be the only motivation for the earlier steps; but I think here we already have compelling use cases for 1 and 2 without looking at 3, right?

@seantalts, thanks for this.

1. Yes, the Intel TBB discussion is a technical one, not a design one. @wds15, sorry I haven't been clear on what is necessary to think about for this. I think you've been prototyping it and I haven't heard more since. (For example, the wiki page on the [dependency checklist](https://github.com/stan-dev/math/wiki/Dependency-Checklist-for-the-Intel-TBB) doesn't really discuss the build requirements.) What we need to think about is how the interfaces will build with this additional dependency and whether it's going to work on all 3 platforms with the compilers we've said we'd support. We can continue that discussion if you think it's time to discuss that.

2. Yes, definitely. I think the proposed technical design should be discussed separately from the design that links the Math library to the Stan language.

3. For this design, I think the first thing to do is to think through the parallelization from the Stan language.

I haven't been idle since this thing has appeared. I've been reading up on parallel design and different language constructs (mostly Python) for parallelism.

This proposal should really be about the language. This is allowing the Stan user to explicitly indicate what is parallelizable. We should really be evaluating the best constructs for specifying this.

Contrast this to the GPU effort. That's not about the Stan user specifying how their model should be parallelized. That is about improving speed for particular functions without the user needing to specify anything in the Stan program.

So... the short answer is yes to point 3. We should be looking at the best possible construct for the Stan language for specifying what can be parallelized.

bob-carpenter:
I'd suggest enlisting @VMatthijs to help with the language spec.  

There's Fortran on the one hand, with no side effects and easy loop parallelization, and then there's C++ on the other hand.  Hopefully Stan will be somewhere in between with side effects limited to exceptions, prints, and updates to the target log density.

VMatthijs:
I'd be happy to be involved in discussions about this. I'm guessing part of the choice to make here is between parallel map constructs and parallel for/foreach loops. For the loop constructs to be safe, we'd need to rely on a dependence analysis that wouldn't be very hard to implement given our current setup. Map constructs should automatically be safe seeing that the Stan language does not allow functions to side effect. The only exception to that is that functions ending in `_lp` can add to target. That is a commutative effect though, so the order in which the writes are performed do not matter. Still, we'd want to be sure to avoid race conditions by having a look at the data structure for storing target and how it is manipulated. Probably you're many steps ahead of me though and have  already figured all of that out.

seantalts:
> I've been reading up on parallel design and different language constructs (mostly Python) for parallelism.

Python is probably the worst language to look at for concurrency or parallelism constructs, unfortunately due to the GIL constraint contaminating everything. I’d recommend looking at the Haskell sparks, Clojure reducers and transducers, java fork/join, Scala parallel collections, erlang actors, and maybe a few others for inspiration.

syclik:
Lol. Thanks for the heads up. I wasn't trying to get inspiration for
design... just for evaluation. (I've also looked at javascript recently,
but less recently.)

One thing that was really clarified through this exercise is that this
proposal for parallelism (and the existing map_rect) is for Stan users to
specify parallel construction of the log prob computation. This is for the
Stan user to decide how to make it happen; the Math library just enables
that. This is different than what I think most Math developers would think
about when thinking parallelization. This proposal and the map_rect
construction isn't meant to be used in implementation of parallelized
functions. I think there's a lot of benefit in thinking about that
parallelism separately and we should just not disallow it. I think whatever
design we settle on would work; it'll be an implementation detail whether
there's an interaction there.

seantalts:
> (I've also looked at javascript recently, but less recently.) 
Javascript is another single-thread language with no native parallelism :P I think OCaml is maybe the only other "major" language without parallelism, so don't look to that either, haha.

> This proposal and the map_rect construction isn't meant to be used in implementation of parallelized functions.

Correct, and a great thing to highlight. Maybe that was a point of confusion for some.

Okay, so what's happening to this proposal? Should we break it up into those three pieces? Looks like we already have a PR for the TBB: https://github.com/stan-dev/math/pull/1180 Should we continue the TBB-related discussion there (or perhaps create an issue to track that discussion instead of the PR)? And then we could split the remaining parts of this doc into the autodiff system proposal vs. the new functions for the Stan language.  @wds15, do you have energy left for any of that? If not, does someone else want to help out?

wds15:
Dismissing this design proposal would be insane - really! I have by now demonstrated very impressive performance benchmarks (posted on discourse) on toy examples and I have in the meantime even worked with simple Stan programs where I use the proposed functions - and I also see nice speedups for simple Poisson inference. If it helps to see how this can be brought into the Stan language I can post these, but I wanted to create more realistic examples first (hierarchical Poisson and a regression case).

Splitting off the TBB discussion is something which is easy - just where? (Just as a note: We settled on this checklist for dependencies, I aligned with @bgoodri for rstan, I worked for full 2 month in wrangling the TBB make system and now I am again set back to zero).

However, I would like to ask that this design review continues by focusing on the changes to math (independent AD concept through `ScopedChainableStack` and nested AD refactor) and that  the functions proposed to be used ideally directly in the Stan language or as building blocks to create super-nodes inside of Stan-math.

The reason why I leave out the Stan language itself for the moment is our current feature freeze of the Stan 2 parser. The thing is that exposing these functions proposed here would require either higher order functions (we decided not to add those to the existing Stan 2 parser) or would need features which we do not have in Stan 2 - like closures and/or lambdas. However, if we would have closures and lambda's then the functions can be made to work very easily. In case we only have closures, but no lambdas then its still very straightforward. This would become apparent from the end-to-end Stan benchmark models where I emulate that stuff using C++ (the unit tests included in the design doc should give you a similar type of information already).

Thus the key thing in this proposal is

- math internals around independent AD + refactored nested AD => allows easy parallelization (with the TBB, for example)
- A parallel map and a parallel reduce from the TBB tailored for used with an adapter to the Stan language in a Stan3 world **and** for use as building blocks to create super-nodes.

My apologies for some confusion around parallel for and map (the TBB notation is mixed here). I only ever meant a parallel map (and NOT a for). These two functions are meant to be useful *inside* Stan math to create super-nodes and probably with some adaptor code *directly* in Stan code (the adapter code should be easy to create in the Stan parser). So these functions target **both**, the Stan math library will have a parallelized map and reduce and the Stan language should be able to adapt these things with concepts like closures and/or lambdas to the Stan language. For example, the parallel reduce can be used as basis to implement parallel lpdf's, I think. The benefit of using them is that they take away the management of the AD things to keep in mind when running parallel - and what is proposed enables nested parallelization (I don't want to encourage lot's of nested parallel things, no, that's likely inefficient, but it will work and this property will allow us to piece different things easily together).

It's great that people think now about parallelism. I would like to re-emphasize that the ideas in this design will do the simple things first - and I have demonstrated that these give real benefits to users as simple Poisson model likelihoods get serious speedups. What more do you want to see?

In short:

- The TBB stuff can be split off
- The rest in this design proposal should continue to be reviewed
- Discussions on how to integrate it with the Stan language can happen at a later stage when we have clarity on Stan 3. For this work here to make sense we can assume that some adapter code can be generated from the Stan 3 parser to make use of the parallel map and reduce directly.

I hope this gives sufficient clarity of the intent so that we can continue. I haven't not heard anything yet on the critical design pieces here (independent AD, refactored nested AD). The only concern I heard was added complexity which isn't really the case since the data structures don't change at all if you do not use these new functions and they only barely change if you do (I explained above).

seantalts:
@syclik does what Sebastian said here sound good to you? This is your review still, I'm just trying to facilitate and get us through our first serious design doc review. 

If the TBB is already approved then we can assume its use in this doc; otherwise we should pull it out. I think I'd personally have an easier time following this if the AD stuff was a separate discussion from a new parallel_map function. Indeed, it feels like TBB was a major choice and the AD stuff is the next major choice and that once that exists, parallel_map is quite easy to implement non-destructively. I think it was possibly the combination of all three at once that lead to a lot of chaos in this thread.

wds15:
As discussed, I have compiled a Stan example which uses the proposed new parallel reduce function. Please refer to the thread on [discourse](https://discourse.mc-stan.org/t/proposed-parallelism-rfc-stan-language-bits/9477). In case there a things not clear, please let me know.

I hope we can now proceed with the review of this RFC.

wds15:
Here is an older post which also describes how the parallel reduce should be used in the Stan language:

https://discourse.mc-stan.org/t/parallel-reduce-in-the-stan-language/8188

This is an earlier description and is not completely runnable. However, in that thread @paul-buerkner gives a really good example why we need a flexible user-specified reduce function. Giving our users flexibility is the key advantage of Stan and Paul's example is a great example for that.

designs/0003-parallel_autodiff.md
```diff
@@ -0,0 +1,705 @@
...
+base in a user-friendly way (addressing point 1+2) and that
+parallelism will scale much better as vectorization is combined with
+parallelism in a coherent approach (point 3).
+
+So ideally we can get more users in need for parallelism to actually
+use it and I also expect that vectorized parallelism will greatly
+enhance the problem set where parallelism actually speeds up a Stan
+program. For example, `map_rect` right now is not able to speed up
+programs which can take advantage of vectorization, but still have to
+process medium to large data sets. A parallel reduce with automatic
```
syclik: The second part of this statement is supporting, but not really part of the design. This is implementation, right? Meaning, this doesn't actually get into the language part.

designs/0003-parallel_autodiff.md
```diff
@@ -0,0 +1,705 @@
...
+
+- Parallel accumulation of independent log-likelihood terms with very
+  good scalability. The good scalability is achieved by combining
+  automatic sharding with vectorization.
+- Parallel evaluation of map where each iteration is independent.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The proposed signatures for the parallel map/reduce with automatic
```
syclik: It would be really useful to include:
1. a `map_rect` example in the Stan language
2. the proposed `parallel_map` example in the Stan language

Then move the C++ lower as a secondary focus. It's how we implement it, not design.

designs/0003-parallel_autodiff.md
```diff
@@ -0,0 +1,705 @@
...
+320ms on 1 core down to 170ms on 2 cores and 150ms on 3 cores. With 4
+cores execution time increases again. This was executed on a macOS
+system and shows greatly the potential of the approach, I think.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The overall strategy to implement all of this is based on
+
+- Introduce the Intel TBB which provides a work stealing queuing
```
syclik: It seems like Intel TBB is an orthogonal concern, isn't it? If so, it's important, but it's not critical to the design discussion.

Also, is the design incapable of working with plain threads?

seantalts:
Okay, let's try this some more! After the Math meeting it seemed like the best route for this was to split it into a few pieces:
1) Adding Intel's TBB library ([dependency checklist wiki](https://github.com/stan-dev/math/wiki/Dependency-Checklist-for-the-Intel-TBB), [thread 1](https://discourse.mc-stan.org/t/general-purpose-parallel-for-each-with-the-intel-tbb/7067/4). [thread 2 on building under windows](https://discourse.mc-stan.org/t/building-intel-tbb-on-windows-with-gcc-4-9-3/7346/9), [thread 3 on benefits](https://discourse.mc-stan.org/t/benefits-of-parallelization-with-a-threadpool-of-the-intel-tbb/9704), [PR with tests passing but waiting on licensing issues via @breckbaldwin](https://github.com/stan-dev/math/pull/1180))
2) A redesigned independent autodiff stack ("Step 2: Independent AD" in this design doc)
3) Refactoring nested AD to use independent AD ("Step 3: Refactor of Nested AD")
4) parallel map and reduce functions for Math library users ("Guide-level explanation")
5) Stan language design and concerns (not in this design doc, will get its own someday)

It seems like it'd be possible to go ahead and review step 2 and 3 by themselves (they are pretty related so I would probably chunk them together). @syclik, would it help if @wds15 and I wrote a separate design doc for those? If we did that, we'd be asking the reviewer to go ahead and assume that it's useful (because we wouldn't be including the end use-case as illustrated here in step 4) and that it could be made easier to init with Intel TBB (step 1, 95% orthogonal but TBB provides a nicer way to initialize independent AD stacks). If so I'm happy to work on that with @wds15. 

Another option would be to wait on the licensing issues of TBB to be resolved and get that merged if you think it would be better to know for sure one way or the other about TBB before considering the rest of these ideas.

wds15:
If splitting it like this helps, then sure. I thought RFCs are for bigger changes to give people an overview, but whatever generates progress is great.

We could consider to move forward with the TBB *at risk*. So we would not merge it to develop until the license stuff is sorted out, but still assume we can use the TBB. In case we do not use the TBB, then it will be another threadpool thingy. Reasons are:

- I did quite some research on the licencse stuff. The most striking reason to me is that the FreeBSD project includes the TBB in their distribution (and that project is very cautious about what they include), all details I wrote on the forum: https://discourse.mc-stan.org/t/licenses-of-dependencies/9295/28

- The use of the Intel TBB as a threadpool shows really good end-to-end performance on Intel **and** AMD, see the forum post https://discourse.mc-stan.org/t/benefits-of-parallelization-with-a-threadpool-of-the-intel-tbb/9704

seantalts:
Those seem pretty compelling for TBB use to me! I'm assuming @syclik is still reviewing the licensing and performance issues around TBB and the 4 other design reviews here and everything like it in the Math library, though. I think we can wait to hear back from @syclik to see if he does indeed want us to split this up or if he can review it as-is knowing what he knows now.

syclik:
Sorry for the delay. To update status:
- Licensing. Yes, I'm waiting to hear back from the lawyer. We (Bob and myself) asked a bunch of questions... we can post the questions online somewhere. We're waiting for the lawyer to get back. So... we don't know if we want to move forward with dependenices that are Apache 2.0 licensed.
- Performance. I've seen enough where I think we're fine with Intel TBB. So the preliminary answer is that it's fine for performance.

Splitting it apart is worth doing. @wds15 and @seantalts, I think for 2 and 3, these should be Math-level designs that are at the C++ level. I guess what I'm trying to say is that I'd expect to see what you'd see in a software architecture design. Especially if it's attempting to change our autodiff stack.

> It seems like it'd be possible to go ahead and review step 2 and 3 by themselves (they are pretty related so I would probably chunk them together).

Yes! If there are independent things to review and consider, it'd probably make sense to split, but if it's really one thing, then keep it together. 

And just to be clear, we want a document that describes the design meaning the software architecture and the interfaces at the C++ level. Ideally, we can evaluate the proposal at that level so that we, the Math devs, agree that this would be good moving forward. Once that's accepted, we can look at one or more implementations that make this proposal a reality, but that's a separate phase from the design phase.

bbbales2:
I think it's fair to ask for the language level stuff now. Like @syclik is saying I don't think they're separate.

But I think Sebastian has basically done this already, going from: https://discourse.mc-stan.org/t/parallel-reduce-in-the-stan-language/8188

It looks like the interface there is:

```
real map_accumulate(first, last, func, params, real_data, int_data)
```

where func is user provided and takes the form:

```
real func(subset_first, subset_last, params, real_data, int_data)
```

I think that's good. It can be done efficiently with nested autodiff in the forward pass cause the outputs are scalars everywhere. And most of the awkward datapacking/whatever will go away with closures.

I do not like the map function proposal without the accumulate/reduce. Without parallel reverse mode, I don't think we can make that always efficient, so it should wait.

I don't really have any comments on the autodiff stack implementation. Is that sorta language level stuff what you are looking for @syclik?

seantalts:
> I think it's fair to ask for the language level stuff now. Like @syclik is saying I don't think they're separate.

We talked about this at a Math meeting and decided we had enough to talk about just within the Math library and that the Stan language stuff was sufficiently orthogonal given good building blocks in Math.

syclik:
> We talked about this at a Math meeting and decided we had enough to talk about just within the Math library and that the Stan language stuff was sufficiently orthogonal given good building blocks in Math.

@bbbales2, we discussed it at the 7/18/2019 Math meeting: https://discourse.mc-stan.org/t/math-meeting-notes-from-7-18-2019/9888. The most relevant bullet point: "We will look at the Math contributions stand-alone from the design-doc perspective." That's what we discussed, but there wasn't a "decision" that you would have known about.

@seantalts, please be courteous to those that weren't at the meeting. Rather than telling other developers that something was decided on their behalf, you could mention who was at the meeting, point to the discussion or the notes, or at least add more color to the nature of what was discussed. That's something that I've insisted on since we started the meetings. See: https://discourse.mc-stan.org/t/monthly-math-development-meeting/8615/1

> important discussions that happen at the meeting will be communicated back online on these forums. If they don’t make it back here, it wasn’t important. No one should be penalized for not being present at a meeting.

seantalts:
I talked to Ben in person as well :) I don't think it's discourteous to mention that you decided this in a meeting. Of course, pasting links is always nicer than not! Didn't realize that was written up anywhere. Glad you're around to manage the design doc that you're reviewing even if it's a week later :)

seantalts:
Ben was just trying to help move this along as it's been more than 4 months now and unfortunately the github issue wasn't updated with the decisions you made in that meeting. One way to avoid this would be to actively manage the design doc and facilitate its resolution.

SteveBronder:
[Rendered Version](https://github.com/wds15/design-docs/blob/parallel_autodiff/designs/0003-parallel_autodiff.md)

SteveBronder:
[Rendered Version](https://github.com/wds15/design-docs/blob/parallel_autodiff/designs/0003-parallel_autodiff.md)

wds15:
Thanks for putting the link there...but we should probably start with a fresh pr to restart things. I will do that once I think it’s ready. For that I wanted to do a last edit.

SteveBronder:
I'm stuck on a train today so I took your drawing and did it in tikz. Not sure if I got it totally right but you can fixup the [code here](https://gist.github.com/SteveBronder/0df546b460a22cff14a3eaa463df3edd).

image

wds15:
Let's try to reopen this a-fresh.

PR #7, opened 2019-08-02
Ragged

bob-carpenter:
Ragged structures


syclik:
@bob-carpenter, I just looked at the ragged structures (and not the closures) doc.

It looks great! Really great.

One question: will we continue to have arrays be declared the way it currently is? As a concrete example:

```
real x[2, 3];
```

and

```
real[{ {3}, {3} }] x;
```

should be equivalent. It seems like we'd want to keep the old syntax because it makes rectangular things pretty easy, but I can see how we might want to move the brackets before the variable instead of its current position.

bob-carpenter:
I want to deprecate

```
real x[2, 3];
```

and replace with 

```
real[2, 3] x;
```

so that the character sequence for the type is contiguous, i.e., `real[2, 3]`.  

And yes, the intention is that `real[2, 3] == real[{3}, {3}]` and we keep convenient rectangular type declarations.

jgabry:
This looks great. Thanks for writing this up @bob-carpenter!

bob-carpenter:
Definitely want to keep rectangular definitions.  I'll make that
clearer.

Specifically, I want to deprecate

  real x[2, 3];

in favor of keeping the types/sizes contiguous:

  real[2, 3] x;

We're going to need this for tuples/structs anyway.

- Bob

wds15:
This looks great. Two questions:

- How do you envision to declare regular arrays of matrices/vectors in the future?
- Will it be possible to declare a ragged array dimensions using another ragged array of ints?

Sebastian

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+A *ragged container* is nothing more than a container whose elements
+may vary in size.
+
+### Ragged arrays of reals
+
+The simplest ragged container is a ragged array.  A simple non-trivial
+example is: 
+
+```
+real[{3, 4}] x = {{a, b, c}, {d, e, f, g}};
```
rok-cesnovar, line 50:
```suggestion
array[{3, 4}] real x = {{a, b, c}, {d, e, f, g}};
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+```
+
+This array is ragged because the sizes of `x[1]` and `x[2]` are
+different.  If `x[i1, ..., iN]` is the same for all valid indexes,
+then `x` is said to be *rectangular*.
+
+We write the unsized type of `x` the same way as for rectangular two-dimensional arrays,
+
+```
+x : real[ , ]
```
rok-cesnovar, line 91:
```suggestion
x : array[ , ] T
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+```
+
+Declarations for ragged arrays are made up of the sizes of their
+elements.  Declaring `x` of type and size `real[{3, 4}]` declares `x`
```
rok-cesnovar, line 95:
```suggestion
elements.  Declaring `x` of type and size `array[{3, 4}] real` declares `x`
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+
+Providing an out-of-bounds index will raise the same exceptions (and
+thus cause the same behavior in interfaces) as other out-of-bounds
+errors.
+
+A ragged three-dimensional array may be defined similarly; each
+element in the sizing is the sizing of a ragged two-dimensional array.
+
+```
+real[{ {2, 3}, { 1, 2, 3 } }] y
```
rok-cesnovar, line 107:
```suggestion
array[{ {2, 3}, { 1, 2, 3 } }] real y
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+y[1] == {{a, b, c}, {d, e, f, g}}
+```
+
+and the second is
+
+```
+y[2] == {{h}, {i, j}, {k, l, m}}
+```
+
+The type of both `y[1]` and `y[2]` is `real[, ]`, the type of a
```
rok-cesnovar, line 126:
```suggestion
The type of both `y[1]` and `y[2]` is `array[, ] real`, the type of a
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+two dimensional and whose second element is three dimensional.
+
+### Ragged arrays of vectors
+
+In addition to arrays, the values of ragged structures may be vectors
+or matrices.  For example, a ragged array of vectors may be declared
+and defined as
+
+```
+vector[{3, 2}] v = { [a, b c]', [e, f]' };
```
rok-cesnovar, line 142:
```suggestion
array[2] vector[{3, 2}] v = { [a, b c]', [e, f]' };
```

rok-cesnovar: Though the array here is optional. It does make it a bit more clear what is going on.

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+value of `v[1]` is the size three vector `[a, b, c]'`, whereas the
+value of `v[2]` is the size two vector `[e, f]'`.  Both `v[1]` and
+`v[2]` are of type `vector`.  All 
+
+All size functions (`rows()`, `cols()`, etc.) work as expected.
+
+Row vectors are handled identically, e.g.,
+
+```
+row_vector[{3, 2}] w = { [a, b c], [e, f] };
```
rok-cesnovar, line 155:
```suggestion
array[2] row_vector[{3, 2}] w = { [a, b c], [e, f] };
```
Again, the array here is optional. It does make it a bit more clear what is going on.

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+```
+
+A ragged array of matrices is declared using a sequence of sizes.  For
+example,
+
+```
+matrix[{ {2, 3}, {3, 2}, {1, 2} }] v
+ = { [[a, b, c], [d, e, f]],
+     [[g, h], [i, j], [k, l]],
+     [[m, n]] };
```
rok-cesnovar, line 191:
```suggestion
array[3] matrix[{ {2, 3}, {3, 2}, {1, 2} }] v
 = { [[a, b, c], [d, e, f]],
     [[g, h], [i, j], [k, l]],
     [[m, n]] };
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+
+Because all constrained types resolve at run time to real, integer,
+vector, row vector, or matrix types, ragged containers of constrained
+types may be declared in exactly the same way.
+
+For instance, to declare a ragged two-dimensional array of
+probabilities, we can use
+
+```
+real<lower = 0, upper = 1>[{3, 4}] x = {{a, b, c}, {d, e, f, g}};
```
rok-cesnovar, line 221:
```suggestion
array[{3,4}] real<lower = 0, upper = 1> x = {{a, b, c}, {d, e, f, g}};
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+```
+
+We can declare a ragged array of simplexes as
+
+```
+simplex[{3, 2, 2}] theta
```
rok-cesnovar, line 227:
```suggestion
array[3] simplex[{3, 2, 2}] theta
```

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+[unresolved-questions]: #unresolved-questions
+
+Whether this gets implemented for the deprecated Rdump format of I/O.
+(The JSON format should not need to change other than that loops need
+to be size-bounded rather than constant-bounded.)
+
+Ragged containers will play nicely with unsized local variables and
+function arguments.
+
+This proposal as written depends on allowing type declarations like:
```
rok-cesnovar, line 561: This was implemented in Stan 2.25 where we now have array[...] real x, array[...] vector[N] a, etc. so we can remove the text below.

designs/0003-ragged-containers.md
```diff
@@ -0,0 +1,575 @@
...
+
+*Optional*:  Because correlation matrices and covariance matrices are
+ typically square, a 3-dimensional array of 2 by 2 covariance matrices
+ may be declared as
+
+```
+cov_matrix[2][3] x
+  = { [[1, 0], [0, 1]],
+      [[3.1, -0.2], [-0.2, 3.1]],
+      [[15.2, 0.1], [0.1, 15.2]] };
```
rok-cesnovar, line 256:
```suggestion
array[3] cov_matrix[2] x
  = { [[1, 0], [0, 1]],
      [[3.1, -0.2], [-0.2, 3.1]],
      [[15.2, 0.1], [0.1, 15.2]] };
```

rok-cesnovar: Question: 

Do we also want to support

```
array[2] cov_matrix[{2, 2, 3}] x
  = { [[1, 0], [0, 1]],
      [[3.1, -0.2], [-0.2, 3.1]],
      [[15.2, 0.1, 0.1], [0.1, 15.2, 0.1], [0.1, 0.1, 15.2]] };
```

designs/0004-closures-fun-types.md
```diff
@@ -0,0 +1,608 @@
+- *Feature Name:* closures-fun-types
```
rok-cesnovar, line 1: This wasnt intended for this PR I guess?

designs/0003-ragged-containers.md
rok-cesnovar: Optional: 

Mention that 

`matrix[{2, 3}] u = [[a, b, c], [d, e, f]];` is a special case of `array[1] matrix[{2, 3}] u = [[a, b, c], [d, e, f]];`

which is why `matrix[{2,3}]` is the same as `matrix[2,3]`. At least to me, seeing it this way made it easier to understand why `matrix[2,3] = matrix[{2,3}]`.

rok-cesnovar (commented):
A couple of questions and suggestions to update the doc now that we support the `array[] T` syntax.

rok-cesnovar:
I also dont see any mention if Stan Math functions should support ragged arrays? 
For example

```stan
array[{3, 4}] real x = {{a, b, c}, {d, e, f, g}};
array[{3, 4}] real y = sin(x);
```

Should this work? If yes then I think adding tests for this in Stan Math should be mentioned somewhere in the docs.

bob-carpenter:
> I also dont see any mention if Stan Math functions should support ragged arrays?

I think that's an independent question.  It'd be great if we could extend unary function vectorization to them.  I would not try to do arithmetic, but reductions like `sum()` might make sense.    The trickier thing will be structured operations like `append_row`, which should work, but `append_col`, which wouldn't make sense.

rok-cesnovar:
Ok, that does make stuff a bit easier then for Stan Math, but a bit more for stanc as we need to keep track at whats ragged and whats not. Thanks.

WardBrian:
@bob-carpenter and I just discussed this and I'd like to suggest an alternate syntax that makes the dimensionality of the final array much clearer. Based on how we do function argument typing, I think we could have a syntax like:


```stan
array[3] int sizes = {3, 2, 3};

array[ , sizes] real = ...;
```

We could even optionally allow `array[3, sizes]`, with the 3 obviously being redundant.

Why? Because if I show you,  `array[sizes]`, you have no idea how many dimensions this array has without chasing down where sizes comes from, which might not always be right there. By contrast, `array[,sizes]`  tells you immediately that the array has 2 dimensions but is ragged. In general this would also work for things like  `array[,,,sizes_3d] real four_D_ragged;`

bob-carpenter:
I'd like to change the spec to work like @WardBrian suggests.

nhuurre:
@WardBrian I like that idea. It also clarifies the situation with `array[] vector[sizes]` (`array` is mandatory)

And since we're likely to get tuples before ragged arrays I suggest matrix dimensions are given as 2-tuples and not two-element arrays.
```diff
-matrix[{ {2, 3}, {3, 2}, {1, 2} }] v
+matrix[{ (2, 3), (3, 2), (1, 2) }] v
 = { [[a, b, c], [d, e, f]],
     [[g, h], [i, j], [k, l]],
     [[m, n]] };
```

PR #14, opened 2020-01-06
JSON Sampling output

rok-cesnovar:
I am requesting comments on my proposal to add JSON output to Cmdstan as previously discuss [here](https://github.com/stan-dev/cmdstan/issues/789).

See the [rendered version](https://github.com/stan-dev/design-docs/blob/ebd39a0a43d1abc5ab1cda700f2f290e31efb27c/designs/0007-json_writer.md).

### DIFF
--- a/designs/0007-json_writer.md
+++ b/designs/0007-json_writer.md
@@ -52,7 +52,7 @@ The main change would be that instead of only allowing csv output for Cmdstan ([
 },
 "param_names": ["lp__","accept_stat__","stepsize__","treedepth__","n_leapfrog__","divergent__","energy__","mu","tau","theta.1","theta.2","theta.3","theta.4","theta.5","theta.6","theta.7","theta.8","theta.9","theta.10","theta.11","theta.12","theta.13","theta.14","theta.15","theta.16","theta.17","theta.18","theta.19","theta.20"],
 "step_size": 0.372621,
-"inverse_mass_matrix_diag": [18.2285, 1.21924, 0.780185, 0.875131, 0.921772, 0.878005, 0.770609, 0.827646, 0.731635, 0.713899, 0.900615, 0.754151],
+"inv_euclidean_metric": [18.2285, 1.21924, 0.780185, 0.875131, 0.921772, 0.878005, 0.770609, 0.827646, 0.731635, 0.713899, 0.900615, 0.754151],
 // "inverse_mass_matrix" would be used if dense_e is used
 "samples":[[-51.044778,0.962712,0.37262,4.0,15.0,0.0,60.34343,11.304475,2.554103,-0.369945,-1.208247,-0.074997,0.867542,0.804843,-1.09118,-1.502441,-0.386866,-1.408202,0.207749,10.359596,8.218486,11.112924,13.520268,13.360128,8.517489,7.467084,10.316379,7.707781,11.835088],[-52.864594,0.84787,0.37262,3.0,15.0,0.0,59.504754,5.789448,0.370447,0.042525,-0.369192,-0.14859,-1.221288,-0.309616,-0.875135,-0.215068,1.375747,-2.241982,0.505575,5.805201,5.652681,5.734403,5.337025,5.674751,5.465256,5.709776,6.299089,4.958912,5.976736],[-52.617135,0.916438,0.37262,4.0,15.0,0.0,60.022217,8.755569,0.473294,1.049158,0.485332,-1.183938,1.728922,0.521784,-0.983077,1.422458,-0.363141,1.104313,0.740966,9.25213,8.985274,8.195218,9.573858,9.002527,8.290284,9.428811,8.583696,9.278235,9.106264],[-56.64873,0.828951,0.37262,2.0,7.0,0.0,60.040686,15.478846,0.379333,0.538912,-0.006475,-2.116285,1.562337,0.431793,-1.403904,1.046636,-0.575158,1.294478,1.623317,15.683273,15.476389,14.676068,16.071492,15.642639,14.946298,15.87587,15.260669,15.969885,16.094624],[-48.92103,0.318818,0.37262,4.0,15.0,0.0,65.581008,10.445837,2.136365,0.261096,1.208946,-0.539443,0.033684,0.03858,-0.283,0.35481,-1.177642,1.607905,0.750019,11.003635,13.028589,9.293387,10.517798,10.528259,9.841243,11.203841,7.929962,13.880912,12.048152],[-47.454484,1.0,0.37262,3.0,7.0,0.0,53.285885,6.43741,4.494301,1.440751,-0.143511,0.187546,0.226508,-0.645326,0.036783,0.990114,1.648384,-0.179351,-0.254901,12.912583,5.792424,7.2803,7.455409,3.537116,6.602726,10.887285,13.845745,5.63135,5.291806],[-48.093197,0.947498,0.37262,4.0,15.0,0.0,54.316045,9.039139,6.876012,-0.354503,-0.005195,-1.037007,0.06226,0.169926,0.836055,-0.536622,-1.369424,0.980715,0.330869,6.601566,9.003415,1.908665,9.467244,10.207553,14.787865,5.349314,-0.377043,15.782549,11.3142],[-46.536515,0.994792,0.37262,4.0,15.0,0.0,52.108556,8.749221,6.750503,0.395675,-0.169556,0.421857,-0.275193,-0.565822,-0.976979,0.95275,1.437867,-0.53873,-0.361132,11.420231,7.604628,11.596972,6.891527,4.929632,2.154119,15.180765,18.455552,5.11252,6.311393],[-46.140776,0.960326,0.37262,3.0,15.0,0.0,50.120721,10.701512,9.488804,0.849441,-0.312036,-1.669781,-0.212627,-0.235502,-1.058349,0.034213,-0.225973,0.890451,0.878977,18.761701,7.740656,-5.142714,8.683931,8.466877,0.659043,11.026155,8.55729,19.150836,19.041961],[-53.644293,0.626838,0.37262,4.0,15.0,0.0,59.601769,7.882549,0.421829,-0.294862,-0.189965,-0.006145,-0.42955,-0.552286,2.84409,-0.446171,-1.276383,0.909832,-0.934817,7.758167,7.802416,7.879956,7.701352,7.649578,9.082271,7.69434,7.344132,8.266343,7.488215]]
 }
@@ -77,15 +77,15 @@ Python's fastest library, `ujson` does not accept these special values. Pandas w
 [1] 1 Inf -Inf NaN
 ```
 
-`rjson` but only supports the standard JSON.
+`rjson`, another R package for reading JSON, only supports standard-compliant JSON.
 
 `rapidjson` library for C++ used in Cmdstan for JSON input, supports parsing the following values: NaN, Inf, Infinity, -Inf and -Infinity. For output, it uses the short names by default but can be easily extended to output strings or Javascript notation.
 
 To make JSON output as general as possible, we should, therefore, support at least two ways of handling special values:
 - Javascript notation (Infinity, -Infinity, NaN) which makes it compliant with Python's packages and rapidjson
-- string notation which makes it compliant with R's jsonlite package
+- quoted notation which makes it compliant with R's jsonlite package
 
-We can discuss if we should also support a "NA" notation. That would mean that we would output NA (or null) instead of NaN and Inf to produce a standard-compliant JSON. `[NA, null, 1]` is considered valid JSON. I have reached out to the authors of `jsonlite` if they would be willing to support the Javascript notation. If we get a positive answer, we could go with only the Javascript notation of special values.
+I have reached out to the authors of `jsonlite` if they would be willing to support the Javascript notation. If we get a positive answer, we could go with only the Javascript notation of special values.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
@@ -98,7 +98,7 @@ output
  file
  diagnostic_file
  format=csv or format=json
- special_values_encoding=js or special_values_encoding=string or special_values_encoding=na
+ special_values_encoding=default or special_values_encoding=quotes
 ```
 2. Move the RapidJSON library to the Stan repository
 
@@ -116,7 +116,9 @@ I have a prototype implementation that only writes out the samples, without the 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-The main drawback of going with a JSON output format is dealing with `Inf` and `NaN` value, as explained above. Given that it will be an optional output format, I do not see any real drawbacks, except for additional code to maintain in the codebase.
+The main drawback of going with a JSON output format is dealing with `Inf` and `NaN` value, as explained above. Given that it will be an optional output format, I do not see any real drawbacks, except for additional code to maintain in the codebase and additional documentation to maintain.
+
+The other drawback is that if we "pull the plug on output partway through", the output will not be a valid JSON object. At miminum it will be needing the concluding "]" and "}".
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
@@ -128,9 +130,9 @@ There are many alternatives to adding a JSON format. Here is a short rundown of 
 - XML: less readable, very inefficient for exporting arrays
 - any binary format (protobuf, thrift, avro, hdf5, ...): not human-readable but obviously faster. I think this is the next or complementary step to this initiative.
 
-The only realistic alternative I see is skipping the step of introducing a JSON output and going with binary output. But that will not enable some requested features like restarting.
+The only realistic alternative I see is skipping the step of introducing a JSON output and going with binary output. But that will not enable some requested features like restarting until we add the option of a binary input.
 
-The other alternative is in terms of the JSON writer library. We could use the boost JSON writer as we already ship boost with Cmdstan. I don't think adding a third library just for writing is necessary.
+The other alternative is in terms of the JSON writer library. We could use the Boost JSON writer as we already ship Boost with Cmdstan. I don't think adding a third library just for writing is necessary.
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
--- a/designs/0007-json_writer.md
+++ b/designs/0007-json_writer.md
@@ -6,18 +6,18 @@
 # Summary
 [summary]: #summary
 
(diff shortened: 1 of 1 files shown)


bob-carpenter:
`inverse_mass_matrix` should be `euclidean_metric`.

There's a dangling clause "rjson but only supports the standard JSON."

What does "NA" mean in JSON?  We want not-a-number, not undefined.  So we don't want R's version of "NA".

Can we make a PR ourselves to jsonlite perhaps to make it easier?  We should see if they want the change first, but that would indeed be the easiest.  

I still don't understand what supporting two versions would entail.  I'm concerned about us defining what amounts to a double standard (pun intended).

With `special_values_encoding=js or special_values_encoding=string or special_values_encoding=na` it looks like three different encodings.  I don't think we should use `na` as it's just too confusing with R.  Shouldn't "js" be "json", because we're referencing the "javascript object notation"?

What's the JSON/CSV line in the performance report?

Are you using a JSON lib to do the output?  Any idea why it's faster?  I can imagine buffering or faster double-precision to string routines.

Does JSON handle denormalized numbers?  We've had problems with this in I/O before.  We should probably take a stand on whether we support their use in Stan.

"I do not see any real drawbacks, except for additional code to maintain in the codebase."  Although it's not a drawback per se, this also complicates the doc and requires more doc to be maintained.  It's not free to add new functionality, even if it's optional, as it complicates the interface and makes it harder to use as the doc gets longer.

I believe protobuf comes with a protobuf to JSON converter.  The beauty of protobuf is that it should be a lot faster with binary I/O.

What is the restart proposal?  Imagine we're in the middle of a run and writing JSON when our cluster times out.  What happens to the output?  Restarting is also going to require more I/O wrangling, so won't be free (I'm all for it, though, as it's a feature we could use).  Restart is hard if you want to save the seed so as to get the same result as if you hand't restarted.  The state of our RNGs is bigger than their seed.

We defiitely don't want to produce rdump format.  The problem is that dump in R is uncontrolled---it uses arbitrary R in the output (for instance, producing `1:5` instead of `c(1, 2, 3, 4, 5)`.  


Would we change all the output commands to not only specify CSV vs. JSON, but also to choose among different encodings in JSON?   I'm concerned about having

riddell-stan:
I've read this (quickly).

Thank you for your work on this @rok-cesnovar .

My sense is that this is an incremental improvement on `stream_writer.hpp`. It's a regression, however, in that stream_writer.hpp already handles nan and inf in a predictable (insofar as CSV is a standard).

@syclik  said something about re-imagining the writer class at some point. What happened to that discussion?

Maybe we could do that and simultaneously settle on a protocol buffer schema that satisfies everyone? One plus for protobuf is that we can encode inf and nan values. With some work I suspect we could come up with a solution for NA values (a mask array?). If we had a group of 3 people invested in the problem I bet we could come to a solution that's much better than using JSON.

I'm invested in getting something better than stream_writer.hpp into Stan. (Last month I wrote a [protobuf version of stream_writer.hpp for httpstan](https://github.com/stan-dev/httpstan/blob/eeee0419dbea7bc69c86e12c27b86f5167caf111/httpstan/queue_writer.hpp); but I'm not happy with it). I'm happy to help work on this if others are interested.

bob-carpenter:
> On Jan 6, 2020, at 12:26 PM, riddell-stan <notifications@github.com> wrote:
> 
> I've read this (quickly).
> 
> Thank you for your work on this @rok-cesnovar .
> 
> My sense is that this is an incremental improvement on stream_writer.hpp. It's a regression, however, in that stream_writer.hpp already handles nan and inf in a predictable (insofar as CSV is a standard).

The fundamental problem isn't implementation, but that there's no standard for not-a-number and infinite values in JSON.  The standard packages in Python and R have gone different ways in their non-standard support.


> @syclik said something about re-imagining the writer class at some point. What happened to that discussion?
> 
> Maybe we could do that and simultaneously settle on a protocol buffer schema that satisfies everyone? One plus for protobuf is that we can encode inf and nan values. With some work I suspect we could come up with a solution for NA values (a mask array?).

By "NA", I take it you mean the R notion of undefined values?  If so, what's the use case given that Stan programs don't recognize undefined values?  I prefer the long-form data solution to the NA solution anyway as it doesn't require non-standard floating point values.  


> If we had a group of 3 people invested in the problem I bet we could come to a solution that's much better than using JSON.

We could certainly come up with something more performant, but I don't know how we could do much better with a human-readable format.

The big advantages aren't the format, but that it'll be easier to support other features like restarting.  The format would be an advantage if there was a standard that dealt with NaN and infinity.

Rok suggests that wouldn't be easy or possible with protobuf output.  I'm not an expert, so I'll leave it to you two to hash out.

> I'm invested in getting something better than stream_writer.hpp into Stan. (Last month I wrote a protobuf version of stream_writer.hpp for httpstan; but I'm not happy with it). I'm happy to help work on this if others are interested.

I would very much like a binary output version.  We got hung up last time
around the every-draw-needs-meta-information discussion, if I'm remembering correctly.  Also, protobuf explicitly states it shouldn't be used for large output files:

  Protocol Buffers are not designed to handle large messages. As a general rule of 
  thumb, if you are dealing in messages larger than a megabyte each, it may be 
  time to consider an alternate strategy.

  from: https://developers.google.com/protocol-buffers/docs/techniques#large-data

But maybe that's a conservative warning.  It also looks like reading is brittle for arbitrary-sized streams (much like our current brittle output!):

  The Protocol Buffer wire format is not self-delimiting, so protocol buffer 
  parsers cannot determine where a message ends on their own. The easiest way to 
  solve this problem is to write the size of each message before you write the 
  message itself. When you read the messages back in, you read the size, then read 
  the bytes into a separate buffer, then parse from that buffer.

  from: https://developers.google.com/protocol-buffers/docs/techniques#streaming

rok-cesnovar:
> `inverse_mass_matrix` should be `euclidean_metric`.

Thanks! I didnt spend much time on arguments as I am sure all of you have a better idea what should the names be. Will update the doc.

> There's a dangling clause "rjson but only supports the standard JSON."

I mean that rjson only supports what the JSON standard defines, which means no special values. Will fix in the doc.

> What does "NA" mean in JSON? We want not-a-number, not undefined. So we don't want R's version of "NA".

NA is undefined yes. This is just an idea how we could output standard-compliant JSON if there was a need (replace Inf, -Inf and NaN with NA or null that are allowed by the standard). I dont think this has much use. Just throwing it out there.

> Can we make a PR ourselves to jsonlite perhaps to make it easier? We should see if they want the change first, but that would indeed be the easiest.

I made an issue here: https://github.com/jeroen/jsonlite/issues/302 If I get positive feedback I will do just that.

> Shouldn't "js" be "json", because we're referencing the "javascript object notation"?

By specifying "js" the user specifies that special values are encoded the same way they are in Javascript (what python json package consumes). By string the user would specify that special values are encoded as strings "NaN", which is what jsonlite consumes. I am going to remove the NA suggestion, its a bad idea anyways.

> What's the JSON/CSV line in the performance report?

Just the ratio of execution times to see how much difference there is relatively to CSV exec. times.

> Are you using a JSON lib to do the output? Any idea why it's faster? I can imagine buffering or faster double-precision to string routines.

Yes, I used rapidjson for output. I would bet on faster double-precision to string. I am also using more precision in JSON (6 vs 5 used in csv), so we can rule out precision issues. But I have not done extensive performance tests. Also plan on benchmarking memory use.

> Does JSON handle denormalized numbers? We've had problems with this in I/O before. We should probably take a stand on whether we support their use in Stan.

It does on input, Dont know for output. Will check. Thanks!

Will split my answers in multiple posts.

bob-carpenter:
> NA is undefined yes. This is just an idea how we could output standard-compliant JSON if there was a need (replace Inf, -Inf and NaN with NA or null that are allowed by the standard). I dont think this has much use. Just throwing it out there.

We need to encode +inf, -inf, and NaN separately.  I think transcoding one or more of these as NA would be too confusing.  

> 
> Can we make a PR ourselves to jsonlite perhaps to make it easier? We should see if they want the change first, but that would indeed be the easiest.
> 
> I made an issue here: jeroen/jsonlite#302 If I get positive feedback I will do just that.

Awesome!

> 
> Shouldn't "js" be "json", because we're referencing the "javascript object notation"?
> 
> By specifying "js" the user specifies that special values are encoded the same way they are in Javascript (what python json package consumes).

My point was that "JSON" is an acronym for "javascript object notation", so isn't going with javascript the same as going with default JSON?  I understand that specifying "json" would be confusing, but then so is "js" if you know the relation between JS and JSON.  Perhaps another option?

> By string the user would specify that special values are encoded as strings "NaN", which is what jsonlite consumes.

NaN looks like a string to me.  Maybe "quoted" instead of "string"?


> I am going to remove the NA suggestion, its a bad idea anyways.

Thanks!


> What's the JSON/CSV line in the performance report?
> 
> Just the ratio of execution times to see how much difference there is relatively to CSV exec. times.

D'oh!

mike-lawrence:
When you say "The only realistic alternative I see is skipping the step of introducing a JSON output and going with binary output. But that will not enable some requested features like restarting.", I'm confused as to what barrier manifests in the binary formats that prevent restarting.

betanalpha:
The terminology used in the HMC code is “inv_metric” and I would recommend that here.

Inverse is important because it’s not a metric on the parameter space but rather the 
momentum space.  Euclidean is technically redundant here as all metrics we’d use in
Stan for a long time are actually Euclidean (yes even including what everyone calls
“Riemannian HMC”).

rok-cesnovar:
> Although it's not a drawback per se, this also complicates the doc and requires more doc to be maintained.

Agreed, will add to the doc.

> What is the restart proposal? Imagine we're in the middle of a run and writing JSON when our cluster times out. What happens to the output?

I meant the idea of doing restarts yes. There is no proposal in the sense of hashed out details. If the execution stops unexpectedly and Cmdstan doesnt exit cleanly, JSON will not be valid (would need a "] }" at the end). Will add that point to the doc.

> Would we change all the output commands to not only specify CSV vs. JSON, but also to choose among different encodings in JSON?

The different encodings would be handled by rapidjson. We would specify it when instantiating the rapidjson writer. We then just use its functions like writer.writeDouble(val). The rest is done by rapidjson.

> It's a regression, however, in that stream_writer.hpp already handles nan and inf in a predictable (insofar as CSV is a standard).

I am not sure I agree on the point that it would be a regression. I feel there is value in another human-readable output that would work on other features. It might also be faster, at least it seems so. And yes, the speed diff will be negligible when the log-prob evaluation takes a substantial amount of time.

> Maybe we could do that and simultaneously settle on a protocol buffer schema that satisfies everyone?

Yes, please! I am happy to work on the binary format also.

bob-carpenter:
See my previous comment with a link to protobuf's doc about sizing and big data.

The issue is that protobuf's not a self-terminating format, so you need to specify how big something is before you start generating.  If it stops partway through, you get ill-formed protobufs.

I already have a question out to Rok asking how JSON can support restarting, as I assume we'd also get unterminated and ill-formed JSON if we just pull the plug on output partway through.

mike-lawrence:
> The issue is that protobuf's not a self-terminating format, so you need to specify how big something is before you start generating.  If it stops partway through, you get ill-formed protobufs.

Ah, then I suggest not painting all binary formats with that brush. HDF5 is self-terminating for example (hence it's ability to write in single-writer/multiple-readers mode).

rok-cesnovar:
> Rok suggests that wouldn't be easy or possible with protobuf output. I'm not an expert, so I'll leave it to you two to hash out.

> When you say "The only realistic alternative I see is skipping the step of introducing a JSON output and going with binary output. But that will not enable some requested features like restarting.", I'm confused as to what barrier manifests in the binary formats that prevent restarting.

Yeah, the same could be done with protobuf, once we add protobuf to input. I meant more in the sense that its something we could use right now. Will improve the text, sorry.

> I already have a question out to Rok asking how JSON can support restarting, as I assume we'd also get unterminated and ill-formed JSON if we just pull the plug on output partway through.

You are correct. I noted that in the comment above. One option for restarts would be to offer a restart_file.json that would always hold the last state of whatever we need for restart. That would be very unefficient tho. This is something a PR for restarts would need to hash out.

rok-cesnovar:
I updated the docs based on the first round of comments. For now I went with "inv_euclidean_metric", but these names are probably something that needs to be hashed out separately. I definitely dont have enough background on most of them to have an opinion.

Hopefully I also answered all the questions so far. If not, do say so.

To re-iterate I am not against working on binary output/input. I am very much for! I don't see this as a JSON or binary design doc. My plan was to first work on this and then move to binary if someone else doesn't pick that up in the meantime.

bob-carpenter:
> On Jan 6, 2020, at 1:28 PM, Mike Lawrence <notifications@github.com> wrote:
> 
> The issue is that protobuf's not a self-terminating format, so you need to specify how big something is before you start generating. If it stops partway through, you get ill-formed protobufs.
> 
> Ah, then I suggest not painting all binary formats with that brush. HDF5 is self-terminating for example (hence it's ability to write in single-writer/multiple-readers mode).

That wasn't my intention---I thought we were specifically discussing protobuf.

I don't know anything about HDF5, but then I'm hardly a data serialization expert.

Multiple I/O formats are relatively orthogonal.  We just need people to design them and build them.

riddell-stan:
These problems are both solved. With regards to sizing, we split up the 
output into distinct messages (e.g., one message for each draw). 
Individual messages are relatively small, far below the limit. With 
regards to the delimiting, there's a standard practice for this. You 
prefix each message with its length in bytes. There's even a C++ tool 
for this which ships with protobuf: 
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/util/delimited_message_util.cc

(re: human readable: JSON is better, you're right.)

bob-carpenter:
> On Jan 6, 2020, at 3:35 PM, riddell-stan <notifications@github.com> wrote:
> 
> These problems are both solved. With regards to sizing, we split up the 
> output into distinct messages (e.g., one message for each draw). 


> Individual messages are relatively small, far below the limit. With 
> regards to the delimiting, there's a standard practice for this. You 
> prefix each message with its length in bytes. There's even a C++ tool 
> for this which ships with protobuf: 
> https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/util/delimited_message_util.cc

How are the messages broken into files?  What delimits when there are no more messages?  For restarting, will the data be readable if someone pulls the 
plug during sampling?


> (re: human readable: JSON is better, you're right.)

I don't think it's better---just satisfying a different goal that some people have.

riddell-stan:
> How are the messages broken into files?  What delimits when there are no more messages?  For restarting, will the data be readable if someone pulls the
plug during sampling?

You can put as many of these length-delimited messages into a file as you would like. The size restriction is just on a single *message*.

> How are the messages broken into files?  What delimits when there are no more messages?  For restarting, will the data be readable if someone pulls the
plug during sampling?

If a file contains a series of length-delimited messages, you detect when the messages are done by reaching the end of the file. (It would be a bit like a file containing a newline-delimited sequence of JSON objects, like a jsonlines file.) Restarting wouldn't be difficult. If someone pulls the plug you would observe fewer complete messages than you anticipated. Each message would still be readable. Again, this is essentially the same as if we used jsonlines (newline-delimited JSON).

The main argument for protobuf vs. jsonlines is that the encoding of doubles is so much more efficient. And Stan uses tons of doubles.

rok-cesnovar:
So for now we have 2 open questions: 

- do we skip JSON output and just move to working on binary input/ouput? My intention here was to add JSON as a complement not as an alternative to adding binary.

- If we dont skip working on JSON output, how many types of different encodings of Inf and NaN should we support. If the developers of the R package `jsonlite` will be willing to accept our pull request, we could go with a single encoding that would work with both Python and R.

riddell-stan:
In favor of binary input/output:
1. We're almost there. This spec can be re-cast as a protobuf spec. And we have a working implementation of a (different) protobuf serialization strategy in httpstan, so we know it can work.
2. It's potentially a big improvement in speed. Parsing binary doubles is virtually certain to be faster than parsing JSON string versions of doubles.

If we do stick with JSON let's consider getting rid of the stream_writer.hpp version so we only have to maintain one thing.

rok-cesnovar:
> 1. This spec can be re-cast as a protobuf spec.

Yes, but the binary doc needs to also thoroughly address the question of why we should be going with protobuf as opposed to any of the other numerous binary formats. And it also needs to specify how we are going to tackle binary input, creating .proto files with stanc3 etc. I do think that any design doc addressing binary should tackle both input and output.

> 2\. Parsing binary doubles is virtually certain to be faster

There is no question or doubt that binary is faster.

riddell-stan:
> Yes, but the binary doc needs to also thoroughly address the question of why we should be going with protobuf as opposed to any of the other numerous binary formats.

There's protobuf, avro, cbor, msgpack and maybe a few others. protobuf is the most mature, has good adoption, has the best tooling.

>And it also needs to specify how we are going to tackle binary input, creating .proto files with stanc3 etc.

I think we only need binary output, I think. We would have fixed proto files. (We're only talking about an alternative to stream_writer.hpp and the equivalent logger, right?)

bob-carpenter:
Conditioned on adding JSON support, then we need something that'll work with R and Python (and ideally Julia).  Patching jsonlite would be ideal if we could make one format that works both places.

I also see JSON as slightly complementary to protobuf.  I believe if we get protobuf, they have some kind of automatic conversion to JSON, but I have no idea whether the JSON would be useful.

bob-carpenter:
> 	• It's potentially a big improvement in speed. Parsing binary doubles is virtually certain to be faster than parsing JSON string versions of doubles.

Raw reading of double in C++ vs. converting from strings using their built-ins is on the order of 100 times faster.

bob-carpenter:
> And it also needs to specify how we are going to tackle binary input, creating .proto files with stanc3 etc.
> 
> I think we only need binary output, I think.

Surely we must need binary input to read what we write.  We'll need that
for any posterior inference we do.  

We'd also like to be able to read all the "meta" data like
adaptation state so that we could restart.

rok-cesnovar:
`jsonlite` developers have declined my feature request on the basis of [1, NaN, InF] not being valid JSON ([1, "NaN, "Inf"] is valid, though I find that very ugly/confusing). So I am going to close this doc, as I have come around on this and feel that having two separate encoding is going to be confusing to the users and was hoping we could go with one.

If/When we get binary I/O we would have the option of providing something like `scripts/toJSON.R` and `scripts/toJSON.py` if there is still a need for JSON then.

riddell-stan:
One thing I'd like to see in the spec: support for streaming the data. I think you could do this by using jsonlines instead of JSON. (Ignoring the inf/NaN issue for a moment.)

Also, I've created a post on the discussion board to summarize my thinking about serialization formats at the present moment: https://discourse.mc-stan.org/t/notes-on-stan-output-serialization-options-yaml-protobuf-avro-cbor/12837/2

One thing to note: this entire spec works fine if we replace JSON with YAML.

PR #15, opened 2020-01-24
Cmdstan3

mitzimorris:
Proposal for a new command-line utility which a conventional argument structure.


SteveBronder:
[Rendered Version](https://github.com/stan-dev/design-docs/blob/cmdstan3/designs/0005-cmdstan3.md)

mitzimorris:
spreadsheet comparing sampler controls available from RStan/PyStan and CmdStan -
https://docs.google.com/spreadsheets/d/1lU6vcY3GF_ftlFdCe2kDhr5FC4YJPWauC5bp2D9zz74/edit?usp=sharing

cells in yellow flag names that we'd like to change.
cells in green flag names we don't want to expose.

jgabry:
Moving the discussion from https://github.com/stan-dev/cmdstanr/issues/136 and https://github.com/stan-dev/cmdstanpy/issues/221 here. 

For the number of warmup and sampling iterations, I like @mitzimorris's proposal (in the spreadsheet) of `iter_warmup` and `iter_sampling` because it's nice to have the same `iter` prefix for all arguments for specifying number of iterations. But if we go with that style then I think we should be consistent and change `max_treedepth` to `treedepth_max` and any other analogous changes.

mitzimorris:
"max_treedepth" is consistent with "iter_warmup" - the concepts are "maximum" and "iterations", and "treedepth" and "warmup" specifiy which maximum value, which set of iterations, respectively.

also, it's consistent with current usage.

jgabry:
Ok yeah that’s fine then. 

So are people on board with `iter_warmup` and `iter_sampling`?

bob-carpenter:
Good luck---English compounds are a mess!  `iter_warmup` makes more sense
in terms of argument name sorting but "warmup iterations" is the way you'd
say it out loud, so I think either would be OK as an argument name.

riddell-stan:
I vote for English word order. "warmup_iters" seems more natural to me.

c.f. 
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions

bob-carpenter:
If we followed those .NET recommendations, it'd be `WarmupIterations` since
they have a no-abbreviations rule and use camel case. This Java-style
verbose camel case naming quickly gets out of hand in my experience.  And
one would use `NumPy.foo(x)` rather than `np.foo(x)` in Python.

Just as in natural language, it's hard to be prescriptive because there are
competing forces that need to be balanced:

* shorter is better than longer (`warm_iter` better than
`warmup_iterations`)
* easy to decode better than obscure (`warmup_iterations` better than
`warmup_iter`)
* English word order is better than not (`warmup_iterations` better than
`warmup_iter`)
* good sort order is better than not (`iter_warmup` keeps `iter_warmup` and
`iter_sampling` together, but we might want to put all warmup parameters
together, in which case it'd be the reverse)

This is why naming's hard!

In this case, I'm OK with any of the suggestions on the table.  I just
wanted to chime in on general principles.

mitzimorris:
agree with @bob-carpenter that naming is hard and I recognize that this is yet another go-around on the subject.

a non-representative sample (office 927) is happy with "iter_warmup" and "iter_sampling" so I've updated the spreadsheet accordingly.

jgabry:
I agree that these more advanced arguments should have more descriptive names, but I propose we first focus getting agreement on the most commonly used arguments (the ones at the top of @mitzimorris' spreadsheet) so we can beta release the CmdStanX interfaces. I'd rather not hold up those releases while debating what to rename the term buffer argument, but I do think it's worth holding them up to decide what to name the number of iterations and chains. Is that reasonable?

mitzimorris:
>  Is that reasonable?

completely reasonable, given that current CmdStanPy doesn't expose the adaptation schedule at all.

For CmdStan3 and CmdStanX wrappers:

- cores
- chains
- iter_warmup
- iter_sampling

Currently CmdStanPy has flag "save_diagnostics" - keep for now, decide on better names later?

jgabry:
Great. For `save_diagnostics` we already changed it to `save_extra_diagnostics` to differentiate from the diagnostics/“sampler params” that are always saved). Is that ok?

betanalpha:
These naming issues were discussed at length at the developer offsite last year
and documented in the interface roadmap draft,
https://discourse.mc-stan.org/t/interface-roadmap-last-draft-before-ratification-vote/10955.

To simplify the argument structure the executables should be split into separate
functionality, at the very least into fixed_param, MCMC with dynamic HMC, optimization
with L-BFGS, and ADVI executables.

Number of iterations:

iters by itself doesn’t communicate the fact that the arguments are specifying 
numbers of iterations.  Additionally “sampling” is an insufficient descriptor of
the main phase as all iterations correspond to MCMC samples.

We eventually settled on num_warmup_iters and num_main_iters for the roadmap.

Other sizes:

Both chains and cores would be less ambiguous if changed to num_chains and 
num_cores aligned with the above.

Seeds:

Specifying seeds for each chain independently does not guarantee independent 
PRNGs states.  The striding provided by sequential chain_id is meant to put each 
chain’s PRNG state into approximately independent regions, at least for the 
ecuyer1988 generator.  This can be better formalized with other generators,
for example the PCG family, https://www.pcg-random.org.  Regardless it would 
be unwise to have users specify seeds for each chain separately and put the 
burden of choosing sufficiently well-strides seeds on them.

Metrics:

The dynamic HMC algorithm does not store the metric but rather the inverse
metric.  Using “inv_metric_file” would make this more clear and less prone to 
errors.

Adaptation Schedule:

The suggested names are inappropriate in my opinion.  They currently read 
more as boolean flags plus “typical set” is ambiguous and all phases of 
adaptation tune the step size so “step size” is not an unambiguous reference 
to the final phase.

When restricting only to the nominal dynamic HMC implementation the three
phases correspond to

- Initial Buffer: adapting the step size only to give the chains time to get close to the
typical set, with no guarantee that they will reach the typical set

- Windowed: adapting the step size continuously while adapting the inverse metric at
the end of increasing windows of iterations

- Terminal Buffer: adapting the step size with the final inverse metric configuration fixed

I don’t think a condensed variable name will adequately describe any of those 
phases, which is why they were originally named after the phase instead of
the behavior during each phase.  More descriptive names like

adapt_init_phase_size
adapt_base_window_size
adapt_term_phase_size

would be welcome.

At the pre-offsite meeting we discussed removing these low-level configurations and 
instead  higher-level adaptation windows that set the low-level configurations to 
achieve specific goals, such as

adapt_full (stepwise and metric)
adapt_stepsize_only (only stepsize, fixed inverse metric at default or user specified values)
adapt_refinement (short stepsize and inverse metric adaptation phases)

andrewgelman:
That all makes sense.  It might also be worth discussing the warmup parameters with Aki and Ben Bales as they (and others) have been working on improved adaptation so maybe there are other inputs that might be coming soon.  Also we should make sure that the settings will enable short runs as that is often helpful for debugging models.
Andrew

peteroupc:
> Seeds: 
> 
> Specifying seeds for each chain independently does not guarantee independent PRNGs states. The striding provided by sequential chain_id is meant to put each chain’s PRNG state into approximately independent regions, at least for the ecuyer1988 generator. This can be better formalized with other generators, for example the PCG family, https://www.pcg-random.org. Regardless it would be unwise to have users specify seeds for each chain separately and put the burden of choosing sufficiently well-strides seeds on them. 

Note that even PCG's notion of "streams", namely its use of an additive constant, [does not produce independent random number sequences](http://pcg.di.unimi.it/pcg.php#flaws).  There are PRNGs that support independent sequences, but these vary depending on the PRNG.

- Some PRNGs give each seed its own non-overlapping and independent sequence of numbers.  This includes so-called "counter-based PRNGs" that hash a seed and an incrementing counter to produce random numbers.
- Some PRNGs (such as all linear PRNGs) have an efficient way to discard a huge number of outputs.  But this number can matter; for some PRNGs, some choices are better at producing independent sequences than others.

It's hard to determine whether a given strategy (for a given PRNG) produces independent random sequences without [testing it](https://github.com/peteroupc/peteroupc.github.io/blob/master/randomtest.md).  See also https://github.com/idontgetoutmuch/random/issues/18#issuecomment-593942197.

Also, besides using the same PRNG for all chains, it may be possible for some of the chains to use PRNGs of a different design from the others.

designs/0005-cmdstan3.md
```diff
@@ -0,0 +1,161 @@
...
+the remaining command line arguments consist of an
+unordered set of pairs of the form `<name>=<value>`
+which are parsed and validated by the `stan` utility
+and then passed in to the services method.
+
+Possible exceptions to this subcommand syntax are:
+
+- The `compile` subcommand, which would instead take a series of makefile options.
+- Utilities such as `stansummary` which operate on one or more files but don't need a compiled model.
+
```
rok-cesnovar: I am not sure if this is a topic to discuss here but would it be worth it to add a "machine friendly" mode that would print the stdout in such a way that it would be easier to parse in other applications. This would simplify cmdstanX interfaces as well as any other script for running cmdstan others are using.

designs/0005-cmdstan3.md
```diff
@@ -0,0 +1,161 @@
...
+            init_file=some_params.json init_radius=0.333 output_file=my_model.csv adapt_delta=0.95
+```
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Interaction with other features:
+
+- CmdStan3 makes calls to GnuMake and therefore will spawn a process on the underlying OS; must be robust to path and filenames
```
rok-cesnovar: What is the plan for the language in which this utility program that will spawn other processes will be written? Is the plan to stay with C++?

designs/0005-cmdstan3.md
```diff
@@ -0,0 +1,161 @@
...
+- CmdStan3 wraps the Stan services layer 
+
+### Implementation
+
+The processing steps:
+
+- Parse command inputs - requires ad-hoc parser for above syntax - if possible, leverage existing c++ command line options parsers.
```
rok-cesnovar: Would it make sense to also support a cmdstan run config file in which a user would specify all these command inputs and then use something like `stan run_cmd_file=model_run_1.dat`?

designs/0005-cmdstan3.md
mitzimorris: it would be a script - ideally python, but bash would be OK too.
in order to it to be Python, we need to have a robust installer for the toolchain.

designs/0005-cmdstan3.md
mitzimorris: excellent idea!

bob-carpenter:
@peteroupc:  We've never been able to detect any difference in output from choices of reasonable PRNGs, even when we just start each chain with a different seed.  We largely set things up how we did to get people to stop complaining about us not doing it the way we did and to make reproducibility easier with just a single seed.  Note that @betanalpha qualified everything with "roughly" because truly independent random number generation is another order of problem.

@betanalpha and @rok-cesnovar:  I'd very much like to balance interpretability with concision.  And getting a release out.  Let's look at some more examples.  

Do we want  to put argument types in uniformly?  So we'd have `data_list = <list>` in R and `data_dict = <dict>` in Python?  And how about `adapt_delta_prob` to indicate the argument is a floating point number in (0, 1)?

Turning to config, do we want `config = <file>`, `sampling_config = <file>`, `sampling_config_file = <file>` or `sampling_json_config_file = <file>`?  Or to indicate we're taking a string and not some object representing a file, how about `sampling_json_config_file_string = <file>`?  

Qualfiying arguments is a slippery slope and brings up the question of when to stop.  That's where balancing concision and interpretability comes in.  Plus, when the names get too long, they're uninterpretable again.

@rok-cesnovar The standard output stuff should just be used to let the user know what's going on, not something you want to parse and use somewhere else.  We've already decided that there are no backward compatibility requirements for it.  So please don't build anything that assumes the standard output will stay fixed.  Anything we are tempted to parse from standard output should have a fixed and parsable standard data format.

@andrewgelman  Pinging @bbbales2 and @avehtari  for you.   This is one of those problems where more hands makes heavier work until the whole thing grinds to a halt due to conflicting opinions and nobody empowered to make a decision.  Last time it ended in tears.

mitzimorris:
I would like to get as much consensus as is possible on this decision and continued discussion allows for new and better suggestions to be fielded.   argument naming is a trade-off between:

- keep existing names that are good enough and widely understood and accepted
- change names and uses that are inconsistent between interfaces to something that is consistent
- remove features that were exposed but have since been deemed unnecessary
- in a very few cases, change existing names.

For these decisions, what matters is what makes sense to both new and advanced users.  for extremely common arguments which are specified everywhere, i.e., warmup and sampling iters, given that most current users are using RStan,  that terminology is what we're going with.

in the short term,  CmdStanPy and CmdStanR will adopt the following naming conventions:

- iter_sampling
- iter_warmup
- cores
- chains

@jgabry - is this OK by you?

avehtari:
> Pinging @bbbales2 and @avehtari for you

Just in case if this meant that I'm supposed to say something: I agree with Bob's and Mitzi's justifications and I'm ok with Mitzi's proposal.

andrewgelman:
fine with me

bbbales2:
I agree with @betanalpha that metric_file should be inv_metric_file.

For what is currently the diagnostic_file, I'm not sure grad_uparam_file is better anymore. Maybe unconstrained_file? I really don't know. I don't even know if VI or optimizing use it.

The rest seems fine.

bbbales2:
Maybe hamiltonian_diagnostic_file? It's giving us info at the level where the Hamiltonian lives (unconstrained parameters, gradients of the hamiltonian, momentums). But then it should only be an option for sampling and not for optimizing/VI.

mitzimorris:
> Maybe hamiltonian_diagnostic_file? It's giving us info at the level where the Hamiltonian lives (unconstrained parameters, gradients of the hamiltonian, momentums).

I like this suggestion - "hamiltonians_file" is shorter.  

>  But then it should only be an option for sampling and not for optimizing/VI.

correct.  CmdStan3 is going to change how the arguments are handled - no more hierarchy/grouping of options - the command line parser will take a flat list of arguments.
for the advi method, we need to decide on an appropriate name for this argument, and then the CmdStan3 argument parser will do the right thing.

mitzimorris:
I've updated the spreadsheet and given everyone permission to comment.

> Adaptation Schedule:
> 
> The suggested names are inappropriate in my opinion.  They currently read
> more as boolean flags plus “typical set” is ambiguous and all phases of
> adaptation tune the step size so “step size” is not an unambiguous reference
> to the final phase.
> 
> When restricting only to the nominal dynamic HMC implementation the three
> phases correspond to
> 
> - Initial Buffer: adapting the step size only to give the chains time to get close to the
> typical set, with no guarantee that they will reach the typical set
> 
> - Windowed: adapting the step size continuously while adapting the inverse metric at
> the end of increasing windows of iterations
> 
> - Terminal Buffer: adapting the step size with the final inverse metric configuration fixed
> 
> I don’t think a condensed variable name will adequately describe any of those
> phases, which is why they were originally named after the phase instead of
> the behavior during each phase.  More descriptive names like
> 
> adapt_init_phase_size
> adapt_base_window_size
> adapt_term_phase_size
> 
> would be welcome.
> 
> At the pre-offsite meeting we discussed removing these low-level configurations and
> instead  higher-level adaptation windows that set the low-level configurations to
> achieve specific goals, such as
> 
> adapt_full (stepwise and metric)
> adapt_stepsize_only (only stepsize, fixed inverse metric at default or user specified values)
> adapt_refinement (short stepsize and inverse metric adaptation phases)

thanks - this is very helpful - in the interest of shorter names, I think
`adapt_init_phase` is preferable to current `adapt_init_buffer`

given that the next two phases do in fact adjust the inv_metric and stepsize, suggesting:
`adapt_metric_window` and `adapt_step_size` instead.

betanalpha:
> For these decisions, what matters is what makes sense to both new and advanced users. for extremely common arguments which are specified everywhere, i.e., warmup and sampling iters, given that most current users are using RStan, that terminology is what we're going with.

PyStan simply copied the RStan arguments, so the “specified everywhere”
in an accident of convention.  The transition to Stan3 was meant to correct
poorly chosen names to avoid ambiguity moving forwards, not preserve 
confusing conventions just because they are conventional.  What is frustrating 
to me here is all of the work and discussion put into the roadmap that is being 
ignored.

For example “iter_warmup” and “iter_sampling” are ambiguous arguments that 
could be interested as binary flags or the length variables that they actually are.  
These arguments are specifying the number of iterations and since number is 
more informative than “iteration”, “num_” is less ambiguous than “iter_”.

Moreover in the context of MCMC _all_ transitions/iterations of the Markov
chain are samples so “sampling” is not incompatible with “warmup”, which 
introduces ambiguity.  That ambiguity is only amplified by the current convention 
of specifying the total number of transitions/iterations/samples and then separately 
specifying how many are reserved for warmup.  If that behavior changes then 
the name of the “not-warmup” samples needs to be able to communicate that it is 
not the total number of transitions/iterations/samples and distinct from the total 
warmup transitions/iterations/samples.

Renaming is our one chance to establish names that will be reasonably 
meaningful to new users without establishing misleading conventions that
will only confuse any user who looks deeper into how Markov chain Monte 
Carlo is being implemented.

> Maybe hamiltonian_diagnostic_file? It's giving us info at the level where the Hamiltonian lives (unconstrained parameters, gradients of the hamiltonian, momentums).
> 
> I like this suggestion - "hamiltonians_file" is shorter.
> 
The “diagnostics_file” defines not a “Hamiltonian” but rather snapshots of the
evolution of the latent Hamiltonian dynamics.  “latent_dynamics_file”, 
“Hamiltonian_dynamics_file”, “unconstrained_dynamics_file” or some 
admixture is more precise and better communicates the purpose of the 
argument.
> correct. CmdStan3 is going to change how the arguments are handled - no more hierarchy/grouping of options - the command line parser will take a flat list of arguments.
> 
> for the advi method, we need to decide on an appropriate name for this argument, and then the CmdStan3 argument parser will do the right thing.
> 

As brought up in the roadmap discussions there is no compatible set of 
arguments that covers all four of the main algorithms (sampling, optimization,
ADVI, and fixed_param).  If these are split up into separate executables with 
separate sets of arguments then “Hamiltonian” becomes redundant for the 
dynamic HMC sampling and something like “latent_dynamics_file” becomes
more appropriate.

> given that the next two phases do in fact adjust the inv_metric and stepsize, suggesting:
> adapt_metric_window and adapt_step_size instead.

These are ambiguous names as the second window is not just adapting the 
inverse metric and the last window is not the only window where the step size
is adjusted.  I do not think that there are descriptive enough names that can be
given to these expert-level arguments that won’t be so long as to be unmanagable.
Again this is why they were originally given names that described the particular 
adaptation phases instead of the complexities of what goes on within those phases.

bbbales2:
I like any version of latent_dynamics_file, hamiltonian_dynamics_file, or unconstrained_dynamics_file. I think I like hamiltonian_dynamics_file the best.

I think the adapt_init_phase_size, adapt_base_window_size, and adapt_term_phase_size names are probably better long.

adapt_metric_window isn't quite right cause by default that is 25, and in a way there are 875 warmup draws by default used to pick the metric, so it's a bit weird. 25 is the base size of the doubling that produces that 875 number. Well it produces 775 then there's an extra 100 lol.

And @betanalpha is right that stepsize is changing the whole time, so adapt_step_size doesn't seem quite right.

I'd also be happy leaving them as init_buffer, window_size, term_buffer too. I assume these adaptation arguments are always gonna be a bit weird cause they're so specialized to Stan.

Those parameter names probably aren't holding any releases back though.

mitzimorris:
@jgabry and I had another look at the sample argument names spreadsheet -
spreadsheet comparing sampler controls available from RStan/PyStan and CmdStan -
https://docs.google.com/spreadsheets/d/1lU6vcY3GF_ftlFdCe2kDhr5FC4YJPWauC5bp2D9zz74/edit?usp=sharing

@betanalpha's arguments w/r/t adaptation are persuasive - one question - for names:

- adapt_init_phase_size
- adapt_base_window_size
- adapt_term_phase_size

for the sake of brevity, could "_size" be omitted from the names?

> The “diagnostics_file” defines not a “Hamiltonian” but rather snapshots of the
evolution of the latent Hamiltonian dynamics.  “latent_dynamics_file”

- latent_dynamics_file 

"Metric" will be changed to "inv_metric".

rok-cesnovar:
Do we really want to make this a new repository that would coexist in parallel with the existing cmdstan repository? While starting fresh definitely has its benefits, I am not sure I am sold on that aspect of this design doc. Once we make cmdstan3, we will tell users to use cmdstan3 and not what is currently cmdstan. Why have that repository there at all? We would not be fixing issues there, accepting PRs, etc. In my opinion it would only confuse the users. Given that the plan is to reuse quite a substantial part of the existing codebase (at least that what I think the plan is) this could be done on the current cmdstan repository without much pain.

This proposal would change the argument names and make a friendly layer on top that will call make, the compiled model, etc. But one would still be able to call make directly if they wanted to (but the wrapper would be recommended). Am I understanding that correctly? Also what if the users calls the model directly with this proposal?

Could we go this route:
1. refactor the names in the current code base and announce develop is now breaking backward compatibility. Obviously bump the major version on release time and delete bin/print and whatever is there just because we were trying not to break backward compatibility.
2. add the new "stan" utility to support everything currently supported by cmdstan.
3. add new features

1 and 2 would have to be completed before a new release. So starting with 1 immediately after the last cmdstan 2.x release would give plenty enough time to complete this.

I honestly really like the cmdstan3 idea. Giving the users a nice utility to use in the command line would be fantastic.

There are some benefits of this utility that could also be listed here: 
- sampling and other algorithms would be called through the stan utility executable which would open the possibility of auto-completing arguments. "stan sample mymodel outp" + pressing tab would autocomplete to `output_file=`.
- any potential change of the build system (but please lets not go in that rabbit hole in this discussion) would not break backward compatibility. Everything would be run through the utility executable.

Some possible additional ideas to add to the design doc:
- enable running multiple chains without the currently necessary bash wizzardry
- make the number of Stan threads an argument of sampling/etc, not an environment variable
- OpenCL platform/device IDs would be argument of sampling, not makefile variables

mitzimorris:
I think that we should proceed with CmdStan3 the way that we've gone with Stanc3, which is, as far as the users know, it's just "the new compiler" - I think this is pretty much in line with what you're saying, no?

in which case, working in the current repo, on branch feature/cmdstan3 would be fine.

> Given that the plan is to reuse quite a substantial part of the existing codebase (at least that what I think the plan is) this could be done on the current cmdstan repository without much pain.

not really.  the existing codebase is mostly a bunch of over-engineered and unmaintainable code around the argument structure.  the file `command.hpp` is a basically a case statement.

> refactor the names in the current code base and announce develop is now breaking backward compatibility. Obviously bump the major version on release time and delete bin/print and whatever is there just because we were trying not to break backward compatibility.

not sure what you mean by "refactor the names"?

bbbales2:
> not really. the existing codebase is mostly a bunch of over-engineered and unmaintainable code around the argument structure. the file command.hpp is a basically a case statement.

I think @rok-cesnovar is right here. If there's a problem with something in cmdstan, we should just replace it. Since the changes break backwards compatibility, we bump version numbers. Starting a new cmdstan3 repo just puts the version number in the package name.

mitzimorris:
> I think @rok-cesnovar is right here. If there's a problem with something in cmdstan, we should just replace it. Since the changes break backwards compatibility, we bump version numbers. Starting a new cmdstan3 repo just puts the version number in the package name.

agreed - see my previous comment - we can work in a branch in the CmdStan repo.

this design doc is about behavoir of the new interface, with the understanding that the CmdStanX wrappers will use it and hoping that it will also be more useable qua interface than the current CmdStan is.

rok-cesnovar:
> working in the current repo, on branch feature/cmdstan3 would be fine

That is fine with me. We can make PRs to that branch as opposed to develop so we can start preparing and doing things in chunks. That or a fork on someone's account works for me.

> not really. the existing codebase is mostly a bunch of over-engineered and unmaintainable code around the argument structure. the file `command.hpp` is a basically a case statement.

I am fine with replacing everything/anything once we decide to bump the major version. We should think about what else we could change as well (so we don't need to do another bump in 3 months): 
- threading and GPU interface (move to arguments)
- machine-friendly output mode
- CSV file format
- remove bin/print
- remove deprecated stuff from stanc, math, ...

 > not sure what you mean by "refactor the names"?

refactor of argument names

mitzimorris:
Closed this because discussion has gone stale. 
the spreadsheet:  https://docs.google.com/spreadsheets/d/1lU6vcY3GF_ftlFdCe2kDhr5FC4YJPWauC5bp2D9zz74/edit#gid=0

was useful in developing CmdStanPy and CmdStanR.

PR #17, opened 2020-03-10
Added an initial attempt at a design doc for reduce_sum

bbbales2:
[Rendered version](https://github.com/stan-dev/design-docs/blob/feature/reduce_sum/designs/0017-reduce_sum.md)

This is an attempt at a design doc for reduce_sum.

@wds15 @SteveBronder and @rok-cesnovar feel free to make changes directly to this branch. We should finalize this draft within a couple days and get a reviewer on it.

### DIFF
--- a/designs/0100-reduce_sum.md
+++ b/designs/0100-reduce_sum.md
@@ -42,9 +42,9 @@ For instance, the sum can be written:
 2. ```f({ x1, ..., xM }) + reduce({ x(M + 1), x(M + 2), ...})```, computing the first M terms separately from the rest
 3. ```f({ x1 }) + f({ x2 }) + f({ x3 }) + ...```, computing each term individually and summing them
 
-The first form uses only one partial sum and no parallelization can be done, the second uses two partial sums and so can be parallelized over two workers, and the last can be parallelized over as many workers as there are elements in the array of ```x```s. Depending on how the list is sliced up, it is possible to parallelize this calculation over many workers.
+The first form uses only one partial sum and no parallelization can be done, the second uses two partial sums and so can be parallelized over two workers, and the last can be parallelized over as many workers as there are elements in the array ```x```. Depending on how the list is sliced up, it is possible to parallelize this calculation over many workers.
 
-```reduce_sum``` is the tool that will allow us to automatically parallelize this calculation. ```x``` is called the sliced array, because it will be automatically sliced up for the parallelization.
+```reduce_sum``` is the tool that will allow us to automatically parallelize this calculation.
 
 For efficiency and convenience, ```reduce_sum``` allows for additional shared arguments to be passed to every term in the sum. So for the array ```{ x1, x2, ... }``` and the shared arguments ```s1, s2, ...``` the effective sum (with individual terms) looks like:
 
@@ -63,46 +63,46 @@ where the particular slicing of the ```x``` array can change.
 Given this, the signature for reduce_sum is:
 
 ```
-real reduce_sum(F func, T[] sliced_array, int grainsize, T1 s1, T2 s2, ...)
+real reduce_sum(F func, T[] x, int grainsize, T1 s1, T2 s2, ...)
 ```
 
 1. ```func``` - User defined function that computes partial sums
-2. ```sliced_array``` - Argument to slice, each element corresponds to a term in the summation
+2. ```x``` - Argument to slice, each element corresponds to a term in the summation
 3. ```grainsize``` - Target for size of slices
 4-. ```s1, s2, ...``` - Arguments shared in every term
 
 The user-defined partial sum functions have the signature:
 
 ```
-real func(int start, int end, T[] subset_sliced_arg, T1 arg1, T2 arg2, ...)
+real func(int start, int end, T[] subset_x, T1 arg1, T2 arg2, ...)
 ```
 
 and take the arguments:
 1. ```start``` - An integer specifying the first term in the partial sum
 2. ```end``` - An integer specifying the last term in the partial sum (inclusive)
-3. ```subset_sliced_array``` - The subset of ```sliced_array``` for which this partial sum is responsible (```sliced_array[start:end]```)
+3. ```x_subset``` - The subset of ```x``` (from ```reduce_sum```) for which this partial sum is responsible (```x[start:end]```)
 4-. ```arg1, arg2, ...``` Arguments shared in every term  (passed on without modification from the reduce_sum call)
 
-The user-provided function ```func``` is expect to compute the ```start``` through ```end``` terms of the overall sum, accumulate them, and return that value. The user function is passed the subset ```sliced_array[start:end]``` as ```subset_sliced_arg```. ```start``` and ```end``` are passed so that ```func``` can index any of the ```sM``` arguments as necessary. The trailing arguments ```argM``` are passed without modification to every call of ```func```.
+The user-provided function ```func``` is expect to compute the ```start``` through ```end``` terms of the overall sum, accumulate them, and return that value. The user function is passed the subset ```x[start:end]``` as ```x_subset```. ```start``` and ```end``` are passed so that ```func``` can index any of the tailing ```sM``` arguments as necessary. The trailing ```sM``` arguments are passed without modification to every call of ```func```.
 
 The ```reduce_sum``` call:
 
 ```
-real sum = reduce_sum(func, sliced_array, grainsize, s1, s2, ...)
+real sum = reduce_sum(func, x, grainsize, s1, s2, ...)
 ```
 
 can be replaced by either:
 
 ```
-real sum = func(1, size(sliced_array), sliced_array, s1, s2, ...)
+real sum = func(1, size(x), x, s1, s2, ...)
 ```
 
 or the code:
 
 ```
 real sum = 0.0;
-for(i in 1:size(sliced_array)) {
-  sum = sum + func(i, i, { sliced_array[i] }, s1, s2, ...);
+for(i in 1:size(x)) {
+  sum = sum + func(i, i, { x[i] }, s1, s2, ...);
 }
 ```
 
@@ -145,10 +145,10 @@ updating the model block to use ```reduce_sum``` gives:
 ```
 functions {
   real partial_sum(int start, int end,
-                    int[] subset_y,
+                    int[] y_subset,
                     vector x,
                     vector beta) {
-    return bernoulli_logit_lpmf(subset_y | beta[1] + beta[2] * x[start:end]);
+    return bernoulli_logit_lpmf(y_subset | beta[1] + beta[2] * x[start:end]);
   }
 }
 
@@ -189,7 +189,7 @@ and
 
 reduce_sum was designed specifically for the case that ```N >> P```.
 
-Regarding the first term (N), if the sliced_arg is of underlying type double or int, there is no autodiff copy performed.
+Regarding the first term (N), if ```x``` does not contain autodiff types, no autodiff copy is performed.
 
 M (the number of blocks of works that TBB chooses) will be decided by the number of work individual operations (N) and the grainsize. It should roughly be N / grainsize. If grainsize is small compared to the number of autodiff variables in the input arguments, then the overhead of copying disconnecting the nested autodiff stack from the main one will probably become evident. Experimentally, a large grainsize can be detrimental to performance overall, though, so grainsize will probably have to be implemented manually via user experimentation.
 
@@ -227,13 +227,13 @@ Repeat autodiff arguments are counted multiple times:
 The main drawback of this new function is the danger that this is replaced by some more advanced and easier to use parallelization mechanism. If that happens, it will be very easy to implement a failsafe backwards compatibility version of reduce_sum that leans on the fact that:
 
 ```
-real sum = reduce_sum(func, sliced_arg, arg1, arg2, ...)
+real sum = reduce_sum(func, x, s1, s2, ...)
 ```
 
 can be replaced by:
 
 ```
-real sum = func(1, size(sliced_arg), sliced_arg, arg1, arg2, ...)
+real sum = func(1, size(x), x, s1, s2, ...)
 ```
 
 where ```func``` was always provided by the user.
--- a/designs/0100-reduce_sum.md
+++ b/designs/0100-reduce_sum.md
@@ -37,10 +37,12 @@ real reduce_sum(F func, T[] sliced_arg, int grainsize, T1 arg1, T2 arg2, ...)
 where the function ```func``` has a signature like:
 
 ```
-real func(int start, int end, T[] sliced_arg, T1 arg1, T2 arg2, ...)
+real func(int start, int end, T[] subset_sliced_arg, T1 arg1, T2 arg2, ...)
 ```
 
-The array ```sliced_arg``` in each case represents the list of calculations to perform and sum. The user-provided function ```func``` is expect to compute the ```start``` through ```end``` terms of the overall sum, accumulate them, and return that value. The trailing arguments ```argN``` are passed without modification to every call of ```func```.
+The array ```sliced_arg``` in the reduce_sum call represents the list of calculations to perform and sum.
+
+The user-provided function ```func``` is expect to compute the ```start``` through ```end``` terms of the overall sum, accumulate them, and return that value. The user function is only passed the subset ```sliced_arg[start:end]``` of sliced arg (as ```subset_sliced_arg```). ```start``` and ```end``` are passed so that ```func``` can index any of the ```argN``` appropriately. The trailing arguments ```argN``` are passed without modification to every call of ```func```.
 
 An overall call to ```reduce_sum``` can be replaced by either:
 
@@ -57,7 +59,7 @@ for(i in 1:size(sliced_arg)) {
 }
 ```
 
-The argument ```sliced_arg``` is called the sliced argument because func only receives the elements of ```sliced_arg``` for which it is responsible. All the other ```argN``` arguments are shared completely between every call to func.
+The argument ```sliced_arg``` is called the sliced argument because func only receives the elements of ```sliced_arg``` for which it is responsible (in ```subset_sliced_arg```). All the other ```argN``` arguments are shared completely between every call to func.
 
 By requiring that the user provided function perform the calculation over a range of inputs, it is possible for the scheduler to lump smaller pieces of work together into big ones (and increase efficiency). This is done by adjusting the ```grainsize``` argument. The scheduler will try to cut up the N parallel terms into groups of size grainsize.
 
@@ -118,8 +120,8 @@ real parallel(int start, int end, int[] Y, int[] trials,
 
 And the model block call:
 ```
-target += reduce_sum(parallel, Y[1:N_subset], grainsize, trials,
-                     temp_Intercept, Xc[1:N_subset], b,
+target += reduce_sum(parallel, Y, grainsize, trials,
+                     temp_Intercept, Xc, b,
                      r_1_1, J_1, Z_1_1, r_1_2, Z_1_2,
                      r_2_1, J_2, Z_2_1,
                      r_3_1, J_3, Z_3_1, r_3_2, Z_3_2,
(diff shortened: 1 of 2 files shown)


designs/0100-reduce_sum.md
```diff
@@ -0,0 +1,169 @@
...
+real reduce_sum(F func, T[] sliced_arg, int grainsize, T1 arg1, T2 arg2, ...)
+```
+
+where the function ```func``` has a signature like:
+
+```
+real func(int start, int end, T[] sliced_arg, T1 arg1, T2 arg2, ...)
+```
+
+The array ```sliced_arg``` in each case represents the list of calculations to perform and sum. The user-provided function ```func``` is expect to compute the ```start``` through ```end``` terms of the overall sum, accumulate them, and return that value. The trailing arguments ```argN``` are passed without modification to every call of ```func```.
```
SteveBronder: Should we also provide a user function to slice matrices and vectors?

bbbales2: I'm for arrays of things only (now that I look at this I'm not sure we support ```T[]``` in the current implementation -- may only be ```real[]```). I don't think we should slice vectors and matrices. I forget why I think this, something about language consistency, but I settled on this last time we talked through this.

bbbales2: I guess the last conversation is roughly between: https://github.com/stan-dev/math/pull/1616#issuecomment-574861041 and https://github.com/stan-dev/math/pull/1616#issuecomment-575671089

wds15: Yes, that's what I recall from the last discussion we had on this. For now we only allow to slice arrays of T (T being whatever), but nothing else. Reasons were language consistency, memory locality in-efficiency for (column major) Eigen matrices and probably more things I forgot.

SteveBronder: Right now the implementation in stan math can accept both eigen vectors and std::vectors. I put that in so that it would be easier for brms etc. to use this.

idk, personally I'm fine with not accepting matrices since that feels like a whole bag of Qs on how to do that. Though it feels like taking in a vector would be nice

wds15: I agree in that slicing a vector would be nice, but I think we ended up saying that a `to_array_1d` is not too expensive to do. From my memory we did decide this matter for the moment, so in the interest of moving forward we should stick with that.

SteveBronder: Bob said `to_array` is rather expensive

https://github.com/stan-dev/math/pull/1616#issuecomment-575671571

bob-carpenter: `to_array` does an alloc and a copy.  That's just memory pressure and time as there's no autodiff implicated.

bob-carpenter:
I like this plan, though it'd be a lot easier on the user if they didn't have to do the slicing.  It'd require the system slicing the `std::vector`, which is an alloc and copy (but no autodiff).   But then the signature could be simply:

```
real f(T[] slice, T1 arg1, T2 arg2, ...)
```

In Stan, we can't make a generic wrapper to do this because we don't have variadic arguments or function arguments.

I'm not saying we should do this, but I wonder what other people think about this ease-of-use vs. efficiency tradeoff?  Not worth it because this whole thing's about efficiency?

Or is it the case that the user might need the start and end slices for getting into the other arguments?

bbbales2:
> I like this plan, though it'd be a lot easier on the user if they didn't have to do the slicing. It'd require the system slicing the std::vector, which is an alloc and copy (but no autodiff). But then the signature could be simply:

This is how it does happen.

Just for clarity, assume we call:

```reduce_sum(f, { 1, 2, 3, 4, 5 }, ...)```

And TBB breaks this into two pieces, one of them can be:

```
f({ 1, 2, 3 }, ...)
```

And the other:
```
f({ 4, 5 })
```

> Or is it the case that the user might need the start and end slices for getting into the other arguments?

Yup this is why the starts and ends are there. I'll try to clarify this later, or feel free to edit if you want.

**Comment by bob-carpenter** [comment]
When you say this is how it does happen, you mean that you don't pass the whole array being sliced, only the bit between start and end?  And you pass start/end for where it originated just in case you need it for one of the other args?  That wasn't clear to me from the quick pass I took over the proposal.  I realize it's not redundant, but if big things to slice are passed in through other arguments, won't this get expensive?  What we really need is something like tuples to pass structured objects, so that the slice would slice everything relevant.

Is this a TBB-specific piece of functionality or will it work with MPI, too?

And yes, please add an explanation of why there's not just an array/vector passed to the user function.

**Comment by wds15** [comment]
> I realize it's not redundant, but if big things to slice are passed in through other arguments, won't this get expensive? 

Yes, it will get expensive if you do pass in stuff in the other arguments which should be sliced. It's correct that tuples would be the best thing to handle this (so a big array of tuples gets sliced)...as a compromise we intend to allow an array of T... which means that T can be anything; like a multi-dimensional array or a matrix. The intention is that if you need to slice other stuff as well (because it's big), then you should stuff it somehow into the array which we slice and that array can have an arbitrary structure.

> Is this a TBB-specific piece of functionality or will it work with MPI, too?

We have implemented it with the TBB only ATM. The design does not prevent an MPI implementation - so when someone has the means to do the MPI bits, then it can be done; it's just that then you would have to re-engineer a scheduler which is currently nicely handled by the TBB.

> And yes, please add an explanation of why there's not just an array/vector passed to the user function.

Sure. The intent is that users can access their auxiliary data structures which may be needed to, for example, express blocking in hierarchical models.

**Comment by bbbales2** [comment]
> I realize it's not redundant, but if big things to slice are passed in through other arguments, won't this get expensive? What we really need is something like tuples to pass structured objects, so that the slice would slice everything relevant.

Yes this is a problem with the current design, good point. I'll add a section to clarify exactly how much extra autodiff overhead is created to support the parallelization (and avoid dealing with race conditions and such).

> And yes, please add an explanation of why there's not just an array/vector passed to the user function.

Yeah I missed that too. Makes sense.

**Comment by wds15** [comment]
@bbbales2 Do you mind if I reword "unresolved issues"? I don't quite see it as bad as you do it and I don't think we do have a reason to make it look as terrible. There is clear guidance from the Intel TBB folks on how to make good use of the TBB facilities so that they scale well the performance. In all examples I have seen so far, the scaling is not so great whenever the computational cost for a given unit is too small (which is the case for the `bernoulli_logit_lpmf` example). Whenever there is heavy computation involved, then things scale rather nice. Finally, when things are roughly equally coded, then map-rect and the new reduce-sum should come out similarly in their performance - with the added benefit of reduce-sum that it is much easier to code.

And at last... there is hopefully still room for improvement to reduce even further the additional "friction" introduced by the running things parallel - but that's for the future as it may require throwing the AD stack in the air and refactor it which is apparently hard to get everyone on board with.

**Comment by bbbales2** [comment]
> the scaling is not so great whenever the computational cost for a given unit is too small (which is the case for the bernoulli_logit_lpmf example). Whenever there is heavy computation involved, then things scale rather nice

The one thing I've seen that has actually scaled well has been some super artificial test code I wrote.

It's an unresolved issue until we have an explanation. I don't accept the Intel explanation cause that's not what I've seen. Increasing grainsize doesn't seem to get us much farther than a 2-3x speedup on 8 cores.

Like didn't someone hit 6x on some models before: https://discourse.mc-stan.org/t/linear-parallell-regression/5230/59 ? I dunno how many cores that was or whatever just there are some other possible models to experiment with.

**Comment by bbbales2** [comment]
There's something happening with memory or something somewhere. If we have to put a note in the docs that say, "don't expect to get more than 3x speedup or so, even when using > 4 cores", I'm fine with that.

It would be better to say, "if specific condition X, then expect to scale" or even better, "outside of specific condition Y, then expect to scale", which we'll only be able to confidently say with more testing.

**Comment by wds15** [comment]
> There's something happening with memory or something somewhere. If we have to put a note in the docs that say, "don't expect to get more than 3x speedup or so, even when using > 4 cores", I'm fine with that.

By Amdahl's law it is always hard to squeeze out more than 2-3x. Everything will depend on the model. I will port an ODE model to this framework... and that should scale amazingly well is my prediction. The hierarchical poisson example I started this all with is also not too artificial for my taste - and that scaled really well as well.

I don't mean to make things sound "all great", but we should be a bit more detailed and also give a prospect when to expect more than 2x.

BTW, a caveat of the current language - which we cannot avoid - is that we are forced to use the probability functions which do compute the normalization constants which are not needed in many applications. Nothing we can do about this at the moment other than asking users to write themselves optimized probability densities for their case which ignore the normalization (which I would not recommend to anyone to do that).

**Comment by bbbales2** [comment]
> I will port an ODE model to this framework... and that should scale amazingly well is my prediction.

Yes yes please do.

> The hierarchical poisson example I started this all with is also not too artificial for my taste - and that scaled really well as well.

Bring that one back into the fold.

I am not averse to models running fast. I just want a list of them.

We can start the docs:

Models that are easy to get good parallel performance from:
...

Models that are not so easy to get good parallel performance:
...

**Comment by wds15** [comment]
Poisson model results:

https://discourse.mc-stan.org/t/parallel-autodiff-v3/12474/24?u=wds15

code for it:

https://github.com/stan-dev/math/wiki/Parallel-reduce_sum-project-page

ODE stuff... is coming

**Comment by bbbales2** [comment]
Do you have the stan models and data in the new format? I think I converted one of them over once but I'm too lazy.

**Comment by bbbales2** [comment]
@wds15 @SteveBronder Umm, do we want to stick with reduce_sum? Reduce and sum seem kinda like the same thing? Maybe map_sum? parallel_for_sum? I'm happy to roll with reduce_sum just bringing it up explicitly before we finalize it.

@bob-carpenter I added details about the total amount of autodiff variables being created on the nested stacks to accommodate the threading calculations. Also added more things about start/end and the relation between sliced_arg in the reduce_sum call and the functor it calls.

**Comment by bob-carpenter** [comment]
"reduce" and "sum" are different.  sum is a kind of reduction.  So I agree that the "reduce" part is redundant.

**Comment by SteveBronder** [comment]
> @wds15 @SteveBronder Umm, do we want to stick with reduce_sum? Reduce and sum seem kinda like the same thing? Maybe map_sum? parallel_for_sum? I'm happy to roll with reduce_sum just bringing it up explicitly before we finalize it.

How about just `map_reduce`? We have `map_rect` so it feels like that makes sense

**Comment by bbbales2** [comment]
> How about just map_reduce?

That's just an overloaded term at this point. The map_reduce software thing has a pretty complicated reduce I think (more than just sum).

**Comment by wds15** [comment]
I renamed the thing already a few times - here is why I like what:

- `parallel` in the name is something which I don't think is a good idea. It just happens that we can run it in parallel, but it's nothing which is worth putting into the name - we haven't done that for `map_rect` either.

- naming it `map` something is attractive since we definitely do map an index sequence to units of execution... but this reduction operation is quite different from the syntax side as compared to `map_rect` such that it makes sense to distinguish it sufficiently.

`reduce_sum` is great from my end, because

1. We say with `reduce` what we do - run a huge reduction.
2. The specific reduce operation happens to be the `sum`. In principle there can be many reduction operations, so this is just a more detailed specification of what reduction we are doing. To me this is like `integrate_ode_bdf` / `integrate_ode_rk45` where again the `bdf` and `rk45` would say that we are doing ode integration - these just narrow down exactly what ode integration is done. The `sum` is the same here. It says what reduction is run.

For the above reasons I very much like `reduce_sum`.

**Comment by bbbales2** [comment]
Oh wow, looks like this has been sitting for four days.

I think this is ready for Official Reviewing. I'm happy to take questions/explain things/make fixes.

File: designs/0100-reduce_sum.md [hunk 2]
```diff
@@ -0,0 +1,214 @@
...
+real reduce_sum(F func, T[] sliced_arg, int grainsize, T1 arg1, T2 arg2, ...)
+```
+
+where the function ```func``` has a signature like:
+
+```
+real func(int start, int end, T[] subset_sliced_arg, T1 arg1, T2 arg2, ...)
+```
+
+The array ```sliced_arg``` in the reduce_sum call represents the list of calculations to perform and sum.
```
mitzimorris: I don't understand what you mean by _list of calculations to perform and sum_.  in the example below, the 2nd argument to `reduce_sum` is variable `Y`

designs/0100-reduce_sum.md
```diff
@@ -0,0 +1,214 @@
...
+or (modulo differences due to rearrangements of summations) the code:
+
+```
+real sum = 0.0;
+for(i in 1:size(sliced_arg)) {
+  sum = sum + func(i, i, { sliced_arg[i] }, arg1, arg2, ...);
+}
+```
+
+The argument ```sliced_arg``` is called the sliced argument because func only receives the elements of ```sliced_arg``` for which it is responsible (in ```subset_sliced_arg```). All the other ```argN``` arguments are shared completely between every call to func.
```
mitzimorris: again, I think that the argument `sliced_arg` is actually `unsliced_arg` (or `items`) and `subset_sliced_arg` is `sliced_arg` (or `slice`)

designs/0100-reduce_sum.md
```diff
@@ -0,0 +1,214 @@
...
+    r_6_1[J_6[n]] * Z_6_1[n] + r_6_2[J_6[n]] * Z_6_2[n] +
+    r_7_1[J_7[n]] * Z_7_1[n] + r_7_2[J_7[n]] * Z_7_2[n] +
+    r_8_1[J_8[n]] * Z_8_1[n] +
+    r_9_1[J_9[n]] * Z_9_1[n] +
+    r_10_1[J_10[n]] * Z_10_1[n] +
+    r_11_1[J_11[n]] * Z_11_1[n] + r_11_2[J_11[n]] * Z_11_2[n];
+}
+target += binomial_logit_lpmf(Y[1:N_subset] | trials[1:N_subset], mu);
+```
+
```
mitzimorris: why such a complicated example?  to illustrate a complex shared calculation?

designs/0100-reduce_sum.md
bbbales2: I think I'm explaining this in the wrong direction. I'll do a rewrite. It's hard to keep it all sorted.

designs/0100-reduce_sum.md
bbbales2: Yes, and that was easy to grab. We'll do a simpler one for the docs.

designs/0100-reduce_sum.md
bbbales2: > is this the vector which reduce_sum will slice up instead of making the user do the work ahead of time?

Yes, but again I think I wrote this a little weirdly here so I'll try to redo with a rewrite.

bbbales2:
@mitzimorris I did a rewrite of the guide. Can you take another look at it and see if it clarifies your questions?

mitzimorris:
@bbales2  - will give you comments.

discussed name "sliced_arg" with @bob-carpenter  - hate to have an argument over argument names, but this one is misleading - "sliced" here implies that what the user provides is already in the state of being sliced - quite the opposite - this is the complete/whole/unsliced vector.

bob-carpenter:
I get that the intention of the name "sliced_arg" is that it's the argument that will be sliced.   The ambiguity is that it could also mean the argument that's already sliced.

I also don't like adding `_arg` as it doesn't provide any information when attached to an argument variable.

How about `slices`?

bbbales2:
> is that it's the argument that will be sliced

Yeah, that's where it was coming from.

Things this argument is:

1. It will be sliced into pieces and passed of to the user-defined reduce function

2. It defines how many terms in the overall sum there are (one per term)

3. These arguments cannot be shared between different terms in the sum (no guarantee that term N + 1 will see the sliced arg of term N -- even if it's possible in certain ways of chunking up the work it would)

Maybe with that, I would call it the ```per_term_input```? ```per_term_array```? That hints that there is one for every term in the sum. I guess ```input``` and ```array``` seem a bit extraneous?

bbbales2:
```non_shared_arguments``` -- arguments really isn't extraneous because this corresponds to a list of things that are arguments to each term in a series. ```non_shared``` to emphasize no sharing between terms.

```per_term_arguments```?

**Comment by bbbales2** [comment]
```slices``` makes it sound like it's already been chopped up for parallelization, so I'm not totally into that.

**Comment by wds15** [comment]
How about `container`? This is the same nomenclature as we use in the `foreach` documentation: https://mc-stan.org/docs/2_21/reference-manual/foreach-loops.html

**Comment by bbbales2** [comment]
Well all the arguments could be containers, so I don't like that so much.

**Comment by bob-carpenter** [comment]
I don't like "container" because it's too generic.   Every vector and array is a container.

I had misunderstood and thought it was pre-sliced.  If it's to-be-sliced in a way determined by other args, that's different.  

The arg name doesn't matter too much as it's just a local variable for doc.  Unlke a function name, we can change it when someone has an idea that's better than `sliced_arg`.

**Comment by bbbales2** [comment]
> If it's to-be-sliced in a way determined by other args, that's different.

Check out the opening of the Guide-level explanation (https://github.com/stan-dev/design-docs/blob/feature/reduce_sum/designs/0100-reduce_sum.md) and see if that does a better job explaining it, cuz we gotta have a good explanation for the docs anyway.

It is quite a bit different than it was.

**Comment by bob-carpenter** [comment]
From @bbbales2 above:
> `slices` makes it sound like it's already been chopped up for parallelization, so I'm not totally into that.

From the doc:
> `sliced_arg` - An array of any type, with each element of the array corresponding to a term of the final summation (the length of sliced_arg is the total number of terms to sum)

I'd suggest using more formal type notation from the get-go.  We want to compute 

```
y = f(x1) + ... + f(xN)
```

where `y : real` is real valued, each argument `x1 : U[], ..., xN : U[]` is an array with elements of type `U`, and the function `f : U[] -> real`.   I like the convention of `y = f(x)` and am less keen on `y = f(arg)`.  The name "arg" doesn't add any information.

The informal doc quoted above makes it sound like `slice_arg : U[ , ]` with value `sliced_arg = { x1, ..., xN }`.  But later it's clear that the intention is `slice_arg : U[]`, with value `slice_arg = c(x1, ..., xN)` (that is, a concatenation of the `xn` arrays).  That's way nicer for users, because they won't have to deal with raggedness.

The doc for `reduce_sum` could say:

<blockquote>
2.  `x` --- the concatenation of arguments `x1, ..., xN`
</blockquote>

There's also an extraneous "4-" in the proposed doc for `grainsize`.

The modulo arithmetic order comment goes both way, so I wouldn't bother mentioning it.

I really like the examples about replacing the top-level call.  Rather than saying "overall call", I'd repeat it as it's not that large.

scalar function -> real function

index from 1, not 0 for Stan docs

Don't call a function "parallel".  Functions should be named after what they compute if possible rather than how they compute it.

Can that multilevel example be simplified with names, say demographic ones that we are all familiar with?  Also, it's not purely hierarchical.   All that brute-force indexing makes me want to figure out if I can do this with arrays rather than variables named things like `r_2_1` and `Z_7_2`.

**Comment by bob-carpenter** [comment]
I missed the formatting.  For naming, I'd just go with something simple and easy to follow.

```
/**
 * Return `f(begin1, end1, x1, u1, ..., uK) + ... + f(beginN, endN, xN, u1, ..., uK)`
 * where `x = concat(x1, ..., xN)` and each `xn = xN[beginN:endN]`, 
 * with sizes of `x1, ..., xN` chosen to roughly satisfy `size(xn) = slice_size`. 
 *
 * @param f function to apply
 * @param x argument to slice
 * @param slice_size target for size of slices
 * @param u1, ..., uK arguments to pass whole
 */
real reduce_sum(F f, T[] x, int slice_size, T1 u1, ...,TN uK);
```

**Comment by bbbales2** [comment]
> I'd suggest using more formal type notation from the get-go

So the way you wrote it there, each ```f``` is a partial sum (or a reduce -- I think at this moment I might like the phrase partial sum better though).

What bugs me about this is that in my head each ```f``` is a term in a likelihood evaluation. But when we're writing code the function we pass in is responsible for more than one likelihood term.

Like, the code is written in terms of ```f : U[] -> real``` but I think about the code in terms of ```f : U -> real```, and so I am tempted to start explaining things in terms of ```f : U -> real```, but that is misleading because eventually the thing we have to write is ```f : U[] -> real```.

I will try to add some of that function notation -- I think it adds clarity.

**Comment by bob-carpenter** [comment]
First, there's nothing log density specific here.  This could be used for anything.

A traditional map function is like our map_rect, and it produces

```
map(f, {x1, ..., xN}) = { f(x1), ..., f(xN) }
```

What we've done is add a so-called reduction (specifically a sum) to produce

```
reduce_sum(f, {x1, ..., XN}) = sum(map(f, {x1, ..., xN})) = f(x1) + ... + f(xN)
```

Eigen doc uses the term "reduction" this way. The compound operation is sometimes called "map-reduce", because it applies a reduction (`sum`) to the result of mapping a function over a sequence of arguments,

```
reduce_sum(f, {x1, ..., xN}) = map_reduce(f, {x1, ..., xN}, sum) 
```

You can generalize all of this to folding binary operations (that's how `sum` is defined), and then the map and reduction can be rolled together.  But I don't think that perspective's going to be very helpful in this discussion.

**Comment by bbbales2** [comment]
@mitzimorris I updated the text of reduce sum again to reflect Bob's comments. Could you have a look?

File: designs/0100-reduce_sum.md [hunk 5]
```diff
@@ -0,0 +1,261 @@
...
+For instance, the sum can be written:
+
+1. ```f({ x1, x2, x3, ... })```, summing over all arguments, using one partial sum
+2. ```f({ x1, ..., xM }) + reduce({ x(M + 1), x(M + 2), ...})```, computing the first M terms separately from the rest
+3. ```f({ x1 }) + f({ x2 }) + f({ x3 }) + ...```, computing each term individually and summing them
+
+The first form uses only one partial sum and no parallelization can be done, the second uses two partial sums and so can be parallelized over two workers, and the last can be parallelized over as many workers as there are elements in the array of ```x```s. Depending on how the list is sliced up, it is possible to parallelize this calculation over many workers.
+
+```reduce_sum``` is the tool that will allow us to automatically parallelize this calculation. ```x``` is called the sliced array, because it will be automatically sliced up for the parallelization.
```
mitzimorris: could we please call this argument `x` ?
if you omit words "is called the sliced array, because" everything is much clearer.

designs/0100-reduce_sum.md
```diff
@@ -0,0 +1,261 @@
...
+Given this, the signature for reduce_sum is:
+
+```
+real reduce_sum(F func, T[] sliced_array, int grainsize, T1 s1, T2 s2, ...)
+```
+
+1. ```func``` - User defined function that computes partial sums
+2. ```sliced_array``` - Argument to slice, each element corresponds to a term in the summation
+3. ```grainsize``` - Target for size of slices
+4-. ```s1, s2, ...``` - Arguments shared in every term
```
mitzimorris: same as above - change `sliced_array` to `x`

designs/0100-reduce_sum.md
```diff
@@ -0,0 +1,261 @@
...
+The user-defined partial sum functions have the signature:
+
+```
+real func(int start, int end, T[] subset_sliced_arg, T1 arg1, T2 arg2, ...)
+```
+
+and take the arguments:
+1. ```start``` - An integer specifying the first term in the partial sum
+2. ```end``` - An integer specifying the last term in the partial sum (inclusive)
+3. ```subset_sliced_array``` - The subset of ```sliced_array``` for which this partial sum is responsible (```sliced_array[start:end]```)
```
mitzimorris: `subset_sliced_array` should be `x_slice`

designs/0100-reduce_sum.md
```diff
@@ -0,0 +1,261 @@
...
+
+1. ```func``` - User defined function that computes partial sums
+2. ```x``` - Argument to slice, each element corresponds to a term in the summation
+3. ```grainsize``` - Target for size of slices
+4-. ```s1, s2, ...``` - Arguments shared in every term
+
+The user-defined partial sum functions have the signature:
+
+```
+real func(int start, int end, T[] subset_x, T1 arg1, T2 arg2, ...)
```
mitzimorris: `subset_x` should be `x_subset`

bbbales2:
Done!

bbbales2:
I'll go update the docs now.

bbbales2:
This okay to merge? I just gave it the design-doc number 17 since this is pull request #17.

PR #22, opened 2020-05-07
Stan CI and repo structure refactor

rok-cesnovar:
This design doc is aimed at upgrading and reorganizing the Stan Continous Integration and Github repository structure.

It's not a typical design doc as it does not propose anything new to the Stan language or backend. But it would have a positive effect on our test coverage and improve the developer experience. The results of these steps will (if accepted) have an effect on most developers, so I figured these will be the best place to do this discussion.

The issues of improving performance testing and numerical testing will be addressed in a separate design-doc that will follow this one in the next days.

### DIFF
--- a/designs/0005-stan-CI-and-repo-structure.md
+++ b/designs/0005-stan-CI-and-repo-structure.md
@@ -3,11 +3,6 @@
 - RFC PR: 
 - Stan Issue: 
 
-This design doc is aimed at upgrading and reorganizing the Stan Continous Integration and Github repository structure.
-Its not a typical design doc as it does not propose anything new to the Stan language or backend. But it will effect our test coverage and hopefully improve the developer experience. The results of these steps will (if accepted) have an effect on most developers, so I figured these will be the best place to do this discussion.
-
-The issues of improving performance testing and numerical testing will be addressed in a separate design-doc that will follow this one in the next days.
-
 # Summary
 [summary]: #summary
 
--- a/designs/0005-stan-CI-and-repo-structure.md
+++ b/designs/0005-stan-CI-and-repo-structure.md
@@ -1,0 +1,143 @@
+- Feature Name: Stan Continous Integration and Repository Structure
+- Start Date: 2020-05-05
+- RFC PR: 
+- Stan Issue: 
+
+This design doc is aimed at upgrading and reorganizing the Stan Continous Integration and Github repository structure.
+Its not a typical design doc as it does not propose anything new to the Stan language or backend. But it will effect our test coverage and hopefully improve the developer experience. The results of these steps will (if accepted) have an effect on most developers, so I figured these will be the best place to do this discussion.
+
+The issues of improving performance testing and numerical testing will be addressed in a separate design-doc that will follow this one in the next days.
+
+# Summary
+[summary]: #summary
+
+This design doc is aimed at upgrading and reorganizing the Stan Continous Integration (Stan CI) and Github repository structure. I propose a list of changes on how to upgrade Stan CI to improve the test coverage, speedup testing and simplify the structure of the repositories and the CI.
+
+There were a lot of changes in recent times in the Stan ecosystem (most importantly stanc3) and in the ecosystems that are closely related to it (most importantly R 4.0 and RTools 4.0) that should be accounted for in our CI and Github repositories. This document is aimed at improving that.
+
+I took the liberty to deviate from the typical structure of the rest of the design docs as this topic is not a typical thing we usually discuss in design docs. I first describe the current status and its defficiencies and then propose a list of changes that could be done to improve the current status. Each change can be treated as separate and the final document will only include the proposals that will be approved.
+
+# Current status
+[current-status]: #current-status
+
+First a summary of all the tests run on Math/Stan/Cmdstan/Stanc3 repositories. Times in square brackets lists the typical run time of the listed test. If not specifically mentioned the tests are run on any operating system.
+
+In Math the following tests are run on Jenkins:
+
++ `test/unit` on Linux with STAN_MPI on [20 minutes or 1 hour on the Linux box]
++ `test/prob` (distribution tests) with `O=0`, `N_TESTS=300` [1.5-2 hours]. 
+On merge to develop, `test/prob` also runs `ROW_VECTOR` tests which run for 3-3.5+ hours or 6 hours on the Linux box.
++ `test/unit` with `STAN_OPENCL` that run on Linux or Mac [~1.5 hours]
++ tests with `thread` or `map_rect` in their names with `STAN_THREADS` [~2 minutes]
++ `test/unit` on Windows [~2 hours]
++ tests with `thread` or `map_rect` in their names with `STAN_THREADS` on Windows [~2 minutes]
++ upstream Jenkins tests (see Stan tests)
+
+On merge to develop we also run
+- `test/unit` on Linux with `STAN_THREADS` on [~20 minutes or 1 hour on the Linux box]
+- `test/unit` on Mac with `STAN_THREADS` on [~1 hour]
+
+In Stan the following tests are run on Jenkins:
+- `src/test/unit` [~0.5 hour]
+- `src/test/unit` on Windows [~40 minutes]
+- `src/test/integration` [~45 minutes]
+- upstream Cmdstan tests (see below)
+- `src/test/performance` on Mac [~5 minutes]
+
+Stan repo also runs Travis Linux tests for src/test/unit with clang and g++.
+
+In Cmdstan the following tests are run on Jenkins:
+- `src/test/interface` on Linux with `STAN_MPI` on [~5 minutes]
+- `src/test/interface` on Mac [~5 minutes]
+- `src/test/interface` on Windows [~5 minutes]
+- performance tests which are run on changes in Stan/Math [~10 minutes]
+
+Stanc3 tests (these are not part of the Math/Stan/Cmdstan structure currently):
+- `dune tests` [~10 minutes]
+Tests that the pushed .expected files match the actual state.
+- model end-to-end tests [~10 minutes]
+Compiles a list of models with the develop cmdstan. See the list in the beginning of a log file [here](https://jenkins.mc-stan.org/blue/rest/organizations/jenkins/pipelines/stanc3/branches/PR-517/runs/5/nodes/46/steps/77/log/?start=0).
+- TFP test [1 minute]
+Tests for the TFP backend.
+- On request we can also run a test to compile all "good/integration" tests with stanc3 + develop cmdstan. [NA]
+
+# Proposals
+[proposals]: #proposals
+
+The order of proposals is not in order of importance, though the first few ones are those where I expect the least amount of controversy. If anyone has any additional proposals please comment with them and we can include them here.
+
+##### Proposal 1: Replace stanc2 tests in stan-dev/stan repository with stanc3
+  
+\
+We are currently still using stanc2 in the upstream Math tests and in the Stan repository tests. This should be removed and replaced with equivalent tests that use stanc3. This would also mean we need to somehow "connect" the Stan and stanc3 repositories. I dont think we need to add stanc3 as a submodule of Stan. We should instead somehow mark the latest stanc3 master git commit hash in a separate file or in one of the makefiles. Each merge to master in stanc3 should then update it. This would trigger Stan tests on each merge in Stanc3, which would be another benefit of this.
+
+##### Proposal 2: Move stanc2 to a separate (new) repository and add it as a Cmdstan submodule
+\
+Stanc2 will be phased-out in the near future and its source files will not be used in the next Rstan release (They will probably still be in the StanHeaders package if I am not mistaken). We will still support stanc2 optionally in cmdstan for a few releases to debug stanc3 bugs. Together with the move we should also close or transfer all language issues in stan-dev/stan.
+
+##### Proposal 3: Use "jumbo" tests on Jenkins in Stan Math
+\
+Jumbo tests are multiple smaller tests merged in one file. We could merge test files in subfolders, for example test/unit/math/prim/fun, in a single file or a few files per subfolder. We would keep existing small files for local testing but Jenkins would run `./runTests.py --jumbo test/unit` which would create and use these "jumbo" test files. 
+
+Preliminary tests have shown that by doing jumbo files (in chunks of 15 files) in most but not all `test/unit` subfolders has cut the test time on Jenkins in half. Even 50% faster tests would be great. But we can potentially get even more. For some more details see [here](https://github.com/stan-dev/math/pull/1863#issuecomment-625178238). We could use this on Linux and Mac tests and if Proposal 5 is accepted, could also use this on Windows.
+
+##### Proposal 4: Refactor `make generate-tests` to reduce the number of generated files in `test/prob`
+\
+Tests files in `test/prob` are generated using the [generate_tests.cpp](https://github.com/stan-dev/math/blob/develop/test/prob/generate_tests.cpp) file. This creates roughly 2700 test files. Compiling these files represents 90%+ of the distribution tests execution time in Stan Math. And the distribution tests are one of the bottlenecks. 
+
+We could improve this by grouping existing test files as in Proposall 3, but in this case this would be done in the generation phase, not after.
+
+##### Proposal 5: Move Windows testing to the RTools 4.0 toolchain
+\
+RTools 4.0 was released in April 2020 and we have already seen Stan users moving to it. This means that we should be testing with the toolchains available in RTools 4.0 to make sure everything works there. Testing both 4.0 and 3.5 is obviously not feasible, if for nothing else, due to the lack of Windows testing worker machines. But we should also not leave 3.5 users in the dark as we know many will not upgrade if not forced to.
+
+I propose moving all Windows Jenkins tests to run with RTools 4.0 and adding a separate worker for Windows with RTools 3.5 (WinRT35). For Jenkins tests this means speedup in compile times for the Math and Stan tests. The WinRT35 worker would be used to make sure that everything still works at the Cmdstan level by compiling/running a set of models. WinRT35 could be run in Cmdstan with free instances on Appveyor/Azure or Github Actions.
+
+This means that Windows users of Math and Stan on the C++ level would be forced to switch to using RTools 4.0 if they want to work in a supported (tested) environment. It is very likely, everything will keep working for them for the forseable future (until the move to `C++17`), it will just not be as severely tested. I do believe that the number of Windows users of Cmdstan with RTools 4.0 in the near future is far larger then those of Stan Math or Stan in `C++`.
+
+##### Proposal 6: Reorganize Stan Math tests
+\
+The current structure of tests have the following defficiencies (in my opinion):
+- the default test should be full unit test with no flags (MPI, OpenCL, threads). The only tests that currently run with no flags are on Windows. The default and most common use of Stan is without these flags so that should also be the default and first test.
+- regular threading tests (not merge to develop) do not run reduce_sum tests
+- tests with OpenCL and MPI should only run tests files that are affected by these flags. STAN_OPENCL and STAN_MPI have zero effect in tests where they are not used. Where they are used, things will not compile if the flag is not set.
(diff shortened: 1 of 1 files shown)


alashworth:
Everything in this proposal is great, thanks for writing it. Here are my thoughts:

P1: Here is just an idea if people want something more than doing this via Makefile and shell script. As stanc3 is a separate project, one way to handle this dependency is having the CI server get dependencies through a binary artifact repository (i.e. package manager). I like repurposing `https://conan.io/` and BinTray because it is free and open-source with first-class support for C++, but there are other options if we have servers to self-host things.

The other extreme is a monorepo, which might make sense because Stan math and the Stan compiler are really extremely coupled, and it does not make sense to develop on one without the other. This introduces build challenges since Dune isn't cross-language. Bazel is an option here. See P7.

P2. No comment.

P3. This is great. How did you refactor to remove the many multiple definitions in the unit tests? This was the main blocker for me doing this automatically with CMake.

P4. I tried to do jumbo compilation of the distribution tests, but the amount of template instantiation combined with COFF limitations on Windows caused the speedup to not be that great. 

P5. Please.

P6. Yes. I spent a very long time doing parts of this in CMake. It would make things much more straightforward. Especially for MPI support, there is specific ifdefs in GTest to hack around the current test organization. (sidenote: I would really like to make MPI support platform agnostic. Like OpenCL, this should work on any platform with an MPI implementation, including Windows, so there is no more special treatment for any one platform).

P7. Well, you know this was a controversial issue previously. But I think for projects of Stan's size, where patches are maintained for all the dependencies, a Google-style monorepo with a single build tool like Bazel would make sense. It's not really a separate project if a change in one repository means orchestrating similar merges in another.

bob-carpenter:
I like Proposal 7.  Someone was trying to combine math, stan, and cmdstan at one point.  That should be OK because they all stay in synch.  stanc3 could be part of that---it didn't exist when the monorepo plans were first floated.

> It's not really a separate project if a change in one repository means orchestrating similar merges in another.

stanc(3) isn't the problem so much as the interfaces, which are out of synch with the main project and all using different versions.

The thinking around stanc3 is that Stan math is only one backend.  With a TensorFlow backend, we get the same coupling as with a Stan math back end---when TensorFlow adds a new function, we have to update the Stan compiler if we want to use it.  Same for when Stan math gets updated.  Same for when Eigen and Boost get updated.  Stan math often updates w.r.t. changes in Eigen, Boost, or C++ itself.  Sometimes that's adding new functions and sometimes new coding styles.

> if you are interested in building Stan math with CMake/Conan or Bazel and want to see what it would look like, I could put something together over the weekend. 

That'd be great.  Many people have tried to do this and fallen short of enough functionality to replace our makefiles on all the platforms on which we deploy.

alashworth:
A few months ago I had something that was 95% of the way there targeting Stan 2.20. For reference, the thread is here: https://discourse.mc-stan.org/t/request-for-comments-wip-stan-cmake-ified-monorepo/10143 and the repository link is here https://github.com/alashworth/stan-monorepo

mitzimorris:
> stanc(3) isn't the problem so much as the interfaces, which are out of synch with the main project and all using different versions.

just to be clear, by interfaces you mean PyStan and RStan, which are at releases 2.18 and 2.19 respectively.   I think that JuliaStan wraps CmdStan, as do the new CmdStanX wrappers CmdStanPy and CmdStanR.

@alashworth - does cmake run on all platforms and is it installed or installable everywhere, or will this add to the toolchain install requirements?

alashworth:
1. CMake is cross-platform. 
2. It is not installed by default. My original proposal was to download a complete set of build tools at compile-time using Conan, a "user-space" cross-platform package manager, and this would handle all dependencies at the expense of disk space. Think Python virtual environment or container. Of course, how do you install Conan? It's a Python package... the one install requirement that handles all the other toolchain requirements.

bob-carpenter:
Nothing's installed on Windows by default, even make.  So that's not such a big concern.  

I think it'd be great to have a package manager that installed Stan and all of its dependencies.  Other people may need to specialize that, but it'd be a huge help for most of our users.  That's another thing we've talked about off and on for years, but never sorted out.

rok-cesnovar:
> combined with COFF limitations on Windows caused the speedup to not be that great.

Oh, I didn't really think about Windows here as distribution tests on CI run on Linux. We will have to be careful there for Windows.

> How did you refactor to remove the many multiple definitions in the unit tests?

Some regex and namespacing. These would have to be fixed first if this approved.

> My old repo where I did this work is out of date

I followed your work and would love your help if we decide to go this route!

My idea was that Stan with Stan Math is the "backend" repo. While the others are interfaces. I was not sure how RStan/Pystan would "take out" what they need if cmdstan would be monorepo. If that is not an issue, that changes things.

The first phase would always be merging and retaining makefiles. Then evaluate CMake/Bazel/etc. 
based on your work @alashworth

SteveBronder:
I like all these! For P7 whats the effort look like? @alashworth if you can slam it out over a weekend I'd say that's a very reasonable amount of time to do that. I think (1) is a definite we should do. Maybe until stanc2 is officially deprecated have the tests run on both 

If we can merge the jumbo files up into one large file we could run something like `gcov` for code coverage. Or maybe there's some way to do gcov with many smaller files

rok-cesnovar:
> For P7 whats the effort look like?

The first thing is merging a submodule which isnt a huge deal. Then transfer issues (not sure what happens with PRs).

I would like to separate switching the build system from this. will make a P8 :)

alashworth:
Just to reiterate: if you just want to see what 80% looks like, minus the (targets old version of Stan, no Stanc3, MPI test support missing, etc), you can look here https://github.com/alashworth/stan-monorepo

mitzimorris:
> My idea was that Stan with Stan Math is the "backend" repo. While the others are interfaces.

Would CmdStan still have sub-module "Stan" which would be the merged Stan/Stan math repos?  This is fine.  The current CmdStan/Stan interface is pretty clean, as evidenced by the fact that it hasn't required much work to update CmdStan for each new release.   CmdStan does depend directly on the Stan makefiles, so if you move to cmake, then you need to move to cmake everywhere.  and this will then make the CmdStanX wrappers dependent on cmake.

from the proposal:

>  The question is would we lose Stan Math users if they would be forced to use Math as part of a slightly larger repository.

good question.  do we know how many folks are using just Stan Math?

rok-cesnovar:
> good question. do we know how many folks are using just Stan Math?

I have no idea, honestly.

I am going to add 7A: monorepo with current makefile build system and 7B: switch build system. 7B depends on 7A which depend on 7.

If anyone has comments on other proposals also comment please. And tag anyone interested.

syclik:
> The question is would we lose Stan Math users if they would be forced to use Math as part of a slightly larger repository.

I don't think you'd lose Stan Math users if this happened. If we think that's a concern, we can programatically bundle a tar.gz that includes what's necessary.

The bigger impact is the social implications and code contracts associated with having a single repo. Since there's no folder-level permissioning, it would make sense to have one set of processes for issues, PRs, merging, documentation. I'm sure there's a version of this that could be great. I'm sure there's also a version of this that makes the development unstable.

Since this is talking about Stan, Stan Math, and stanc, perhaps there's a common standard that we can all share. For Stan Math, the invariant that we strive for is that every merge into `develop` works. In Stan Math, there's also no ability to cross repos to fix things, but that could happen if we combine repos. We also have the standard GitHub PR merge permissions set of requiring at least one approval and zero disapprovals. In some senses, it's easier to deal with a fix when we can go across these boundaries. But we need to start being extra vigilant about it... we don't want PRs to start having the equivalent of [legislative riders](https://en.wikipedia.org/wiki/Rider_(legislation)) attached to PRs that have side effects that aren't explicitly documented and hard to figure out without a full inspection of the code. It's hard enough to keep that in check within the Stan Math repo.


@rok-cesnovar, something specific to mention from the Math part:
> I think we dont need additional tests on merge to develop.

This one is tricky. We've had problems with this before, but that's because of how we've been testing this.

Our current behavior:
1. Each PR is tested by taking the PR branch and merging the current `develop` into the branch.
2. When a PR is merged, then `develop` is tested.
3. When `develop` is updated after a PR, we do not retest all the existing PRs with a merge from the new `develop`.

The reason we test `develop` after a merge is because we don't know if the PR actually passes tests when the latest version of `develop` is merged into the branch.

We could change this... then it would look like:
1. Each PR is tested by taking the PR branch and merging the current `develop` into the branch.
2. When a PR is merged, `develop` is NOT tested.
3. When `develop` is updated after a PR, we RETEST all the existing PRs with a merge from the new `develop` (or some subset needs to be retested).

Either way is fine, but just removing the test on a merge isn't really stable enough. We added that in after some trial and error and some really difficult debugging because what we assumed worked didn't.

rok-cesnovar:
> This one is tricky. We've had problems with this before, but that's because of how we've been testing this.

Yeah my wording is not the best here. I meant in the sense that we would perform the exact same tests on merge but no additional ones. We currently run STAN_THREADS tests on Mac and Linux on merge only.

syclik:
Ah. Yes, probably don't need to duplicate those.

rok-cesnovar:
@syclik if I understand your comment correctly, you are essentially worried for cases where developers without a good enough knowledge of some parts of the codebase would approve PRs. Not sure if that is what you meant.

If you did, I think you are correct that is definitely a valid concern. I am however not sure merging Stan Math to Stan would be an issue in these terms. Neither would a subsequent merge of Stan in a Cmdstan repo. 

This is all `C++` land and the overlap of developers across these repositories is big. Offcourse not everyone has a deep knowledge in everything, but that is also true for Math itself, which has a larger codebase than Stan and Cmdstan. 

What would change things is including stanc3 directly, which I dont think we want to do. In Cmdstan/stanc3 I would only like to somehow mark merges to master in stanc3 in the stan repo. Because what we need is better tracking of history and a simple way of checking out Cmdstan at any given point in time. This is what we dont have right now.
It can also be a submodule, if we find a way that doesnt force `C++` devs to install the OCaml environment.

rok-cesnovar:
I updated the design doc based on the comments so far:
  - split P1 (use of stanc3 in stan repo tests) in two competing proposals: 1A (stanc3 becomes a stan submodule) and 1B (stanc3 git hash in a stan makefile)
  - clarified the text on merge to master tests
  - added P8: merge stan to cmdstan

I decided to not include the change to build system for now as I think we should first decide if merging is what we want and then if we do that we can evaluate the build systems based on work and know-how from @alashworth. 

Otherwise, the merger discussion will always derail to "what build system to use". I think merging has enough of its own benefits to be considered. And a richer build system is also something that would be really great to have, but with careful consideration on what that means to the user. We more or less know its a win for the developer. If we get to a merged repo it will also be much easier for devs and users to test the changes.

alashworth:
Some notes on merging repos: 

1.  issues and pull-requests: you might want: https://github.com/alashworth/github-issues-import
2.  if you want to edit history (necessary if you want to have git log and bisect working in a monorepo with different folder structure), try https://github.com/newren/git-filter-repo. If you are willing to cull large files, you can shrink Stan math from 1 GB to 100 MB.

rok-cesnovar:
Thank you, that is very valuable info!

syclik:
> @syclik if I understand your comment correctly, you are essentially worried for cases where developers without a good enough knowledge of some parts of the codebase would approve PRs. Not sure if that is what you meant.
> 
> If you did, I think you are correct that is definitely a valid concern. I am however not sure merging Stan Math to Stan would be an issue in these terms. Neither would a subsequent merge of Stan in a Cmdstan repo.

Yes. I agree that we can merge them and figure it out, but I think we should think about it before we do it. We need more clarity on who is supposed to do what in terms of approvals and reviews. We want it to be as friendly to developers, long-time contributors and new contributors alike, but we also need the repo to stay stable. The question really is... who proposes these processes (I think it should probably go in the design docs?), who approves them, do the developers need to all agree to the new change in processes, if devs don't follow along who is enforcing changes? Rather than just forcing a new set of procedures on people, it'd be good to have buy-in. Especially since we're an open-source project... no single person can dictate how this works.

> @alashworth 
> issues and pull-requests: you might want: https://github.com/alashworth/github-issues-import

Just a heads up, GitHub now allows for transfering of issues. GitHub Doc on [transferring an issue](https://help.github.com/en/github/managing-your-work-on-github/transferring-an-issue-to-another-repository). And a screenshot (I highlighted it in blue on the bottom right):
image

rok-cesnovar:
> but I think we should think about it before we do it.

Agreed, that is why I started this discussion. We have been discussing this on and off and I wanted to take some more action on this. Its not something that would be done over night. 

I also said I will not force this, as this particular change will be difficult to get consensus on as it changes the status quo rather dramatically and not incrementally.

But I also know that the current structure has a lot of drawbacks. I am not sure it has that many benefits. But it is the current norm.

> Rather than just forcing a new set of procedures on people, it'd be good to have buy-in.

Please take this as discussion on the changes as if they were posted on Discourse. I want to make the changes that will be agreed upon with the developers. We dont have a formal TWGD or whatever it was called who called the final shots. I am listing changes that we could do, will give enough time so anyone can reply and we can. And then start work on what was agreed upon.

syclik:
> Please take this as discussion on the changes as if they were posted on Discourse. I want to make the changes that will be agreed upon with the developers.

Sorry if I made it sound like this discussion was the step of "forcing a new set of procedures"... this discussion is the opposite of that! This discussion is our chance to think through the procedures and get buy-in before it happens.

One of the reasons I asked about the design docs on discourse was because I don't know what the expectations are. And I agree that the current structure has a lot of drawbacks. Just because it's the norm doesn't mean it has to be that way! We just have to respect the open-source nature of the project and the distributed nature of development as we change things around.

alashworth:
Yeah. GitHub functionality to transfer issues unfortunately only works one issue at a time (as of September 2019) when I last looked at it. So maybe it's changed.

bob-carpenter:
CmdStan is very small and shouldn't get in the way of anything RStan or PyStan's doing.  Having said that, it looks like CmdStan hasn't been written with the same attention to detail on namespaces.  For example, I see `stan_summary_helper.hpp`, the first thing I looked at, just defining functions in the top-level namespace. 

The language doc (reference manual, user's guide, functions reference) also gets updated when functionality gets added to the language (that is, in stanc3).  

Modulo adding namespaces to cmdstan, would there be any issues with creating a monorepo with all of the following?

* docs
* math
* stan
* stanc3
* cmdstan

That'd be a complete cmdstan build.  We'd have to sort out directory structures, but I don't think there's any namespace overlap among math and stan (though it looks like cmdstan may be a problem).   That is someone could download this repo, and just link directly to a math `.hpp` file without even cutting out the rest.   Or they could just use `src/stan/math`.  That's how we include Eigen in math.

The community standards for review, etc., are the same across the math, stan, and cmdstan repos.  Trying to formalize them in a document has been more challenging.    If someone wants to try to write up what those standards are, we can all try to agree explicitly.  I wouldn't want lack of agreement on a written form of procedures to hold the monorepo back.

rok-cesnovar:
I like this idea. We would have to be nice to C++ devs by providing a way to use this without an Ocaml environment, but I am all for this. More than willing to lead the effort if we (hopefuly) get buy in from other devs.

Math to stan, stanc3 to stan, stan to cmdstan and docs to cmdstan seems like a nice incremental way to get there.

What currently happens a lot is stuff gets implemented in math but never gets to stanc3. Or gets added to the language but not to the docs.
 
Also, just today I decided to fix a small issue in Stan: https://github.com/stan-dev/stan/pull/2917
Simple file relocation turned into 3 PRs due to a namespace change.

mitzimorris:
I very much like this idea.

> We would have to be nice to C++ devs by providing a way to use this without an Ocaml environment, but I am all for this. More than willing to lead the effort if we (hopefuly) get buy in from other devs.

in addition to the Ocaml requirements imposed by stanc3, the docs add requirements for R bookdown package, pandocs, and latex.

what will the policy be on tests required to submit a PR?

regarding CmdStan, one possibility would be to leave current CmdStan alone, and instead plan on adding CmdStan3.

alashworth:
The docs are their own kettle of fish. When I was trying to complete all the CI tests on Windows and MacOS, I ran into some problems because for whatever reason the MikTeX distribution I was using on Windows was not able to parse the beta distribution documentation. 
 
Speaking of CI, this is just my opinion, but I wish all CI tests ran on all three operating systems. That includes docs, and MPI. Yes, there are Windows servers that use MPI. MS-MPI works. It should be feasible to achieve this dream once we soup up the build system.

(also, and this is a separate issue, I would really be interested in either a) creating a minimal set of unit tests that runs in 10 minutes or less on a typical 4-core laptop, or b) creating a public server that caches object files ala https://github.com/mozilla/sccache ).

bob-carpenter:
> The docs are their own kettle of fish.

Not to mention a can of worms.  But they are the things that are most requiring of change when things in the language change.

> Speaking of CI, this is just my opinion, but I wish all CI tests ran on all three operating systems.

I don't think anyone will disagree.  It's just waiting on someone who can configure it.  We don't really need to generate the docs everywhere, as we don't expect users to be generating them.  I don't know when you last looked at the docs, but everything's getting generated through bookdown now, which means R and pandoc.

mitzimorris:
> When I was trying to complete all the CI tests on Windows and MacOS, I ran into some problems because for whatever reason the MikTeX distribution I was using on Windows was not able to parse the beta distribution documentation

yup, we carved the docs out into their own repo in order to lighten the testing burden on individual devs.   we need documentation which docs to change and how to change them - i.e. add docs on the docs to the dev process docs.

HaoZeke:
I'm a bit new to the ecosystem, but I personally dislike monorepos. The stan ecosystem is heavy enough as it is (given that `stan` pulls `math` in as a submodule) and a monorepo only makes dealing with issues and PRs more of a mess. Also, adding `math` to other libraries would become more complicated for C++ developers (nothing impossible, but a non-trivial annoyance none the less)

rok-cesnovar:
> in addition to the Ocaml requirements imposed by stanc3, the docs add requirements for R bookdown package, pandocs, and latex.

Yes, if they wish to build them. This should be done so that this is optional. Much like you need doxygen to build math docs but can work on Math without doxygen installed.

> what will the policy be on tests required to submit a PR?

We can do this based on what folders are changed. In Math we, for example, don't run any tests if a PR only touches .md files or the /doxygen folder.
So we could do:
- change in math -> full tests
- change in stan -> test on algorithms+compiling and cmdstan tests
- change in cmdstan ->  cmdstan interface tests

> Also, adding `math` to other libraries would become more complicated for C++ developers. 

> More specifically, for cross-domain applications, it is harder to convince people to agree to use stan, or a part of the stan repo than "the math library used by stan".

We could, at least I would advocate for that, still have Math releases. So a tar file with only source files of what is currently Math.

@HaoZeke I think what we had to do to upgrade the cpplint version in Math (thank you by the way), highlights one of the reasons I think we should be merging repos.
We had to do a PR in Math https://github.com/stan-dev/math/pull/1871 and a PR in Stan https://github.com/stan-dev/stan/pull/2916 and then do a custom Jenkins run with manually setting branches.

If we had cpplint in cmdstan (why dont we?) that would be a third PR. Not the end of the world, but not pleasant either.

rok-cesnovar:
Thanks to everyone that have commented so far. I think the consensus opinion on P7 and P8 from those that commented is "Yes, but very careful". Obviously there is more to discuss there and let's continue with that.

I would like to also get your opinions on some of the less intrusive changes.
Specifically:
- Proposal 1A: Use stanc3 in the stan-dev/stan repository tests and make stanc3 a submodule of stan
- Proposal 1B: Use stanc3 in the stan-dev/stan repository tests
- Proposal 2: Move stanc2 to a separate (new) repository and add it as a Cmdstan submodule
- Proposal 5: Move Windows testing to the RTools 4.0 toolchain
@ahartikainen I would really like to hear your thoughts here if you have time.
- Proposal 6: Reorganize Stan Math tests

I personally prefer 1A over 1B. If we get consensus on these specific proposals we can mark them approved.

ahartikainen:
> * Proposal 5: Move Windows testing to the RTools 4.0 toolchain
>   @ahartikainen I would really like to hear your thoughts here if you have time.

Moving RTools40 sound reasonable. I think CmdStanPy users should move to use RTools40 and this is going to be the default compiler used in `cmdstanpy.install_cxx_toolchain`.

Setting up Github Actions is probably easy solution, if you don't need to use it with PR from forks. E.g. run tests only when `develop` is updated.

rok-cesnovar:
> run tests only when `develop` is updated.

Thanks. Running Rtools 3.5 tests on push to develop is enough yeah. Updated the doc.

rok-cesnovar:
Note to self: when these changes are complete, update https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms#supported-cpp-versions-and-compilers

PR #24, opened 2020-05-23
[WIP] tuples and structs rough proposal

bob-carpenter:
So far just moved this over from the Wiki.  So no need to review until I get it into the proper form.  Comments always welcome of course!

### DIFF
--- a/designs/tuples_structs.md
+++ b/designs/tuples_structs.md
@@ -1,0 +1,366 @@
+- Feature Name: Tuples and Structs
+- Start Date: 2020-04-23
+- RFC PR: ??
+- Stan Issue:
+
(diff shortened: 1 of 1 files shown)


WardBrian:
I know this has been open for a while, but one suggestion:

A common feature of other languages with tuples is unpacking through assignment, e.g.

```
x, y = (1, 2)
```
leaves x with value 1 and y with value 2. This sort of thing is not particularly difficult on the implementation side (you can just treat it as two assignments in serial) but makes programming with tuples much more elegant. 

Bonus points if we include the convention from OCaml, Python, etc of having `_` serving as an 'ignore this' name, if only by convention

WardBrian:
Closing in favor of #46

PR #27, opened 2020-09-26
Added regression syntax design doc

bbbales2:
I've been thinking about some sort of regression syntax in Stan. It's kind of a follow up to [Speeding up hierarchical models](https://discourse.mc-stan.org/t/speeding-up-hierarchical-models/14783). When I say regression syntax I'm talking about the interfaces (brms/rstanarm/lme4/glmer/lm/...) that take in something like the R regression syntax (`y ~ 1 + x + (1 | group) + ...`) and do something with it.

It's probably something that has come up before, but there's just so much there it's hard to talk about it without writing a lot down. So I tried to write down the bare minimum justification and implementation of a regression syntax in Stan.

Stan is written to do regressions and more, so at face value this doesn't seem that important but I'm trying to hit three points:

1. Make it possible to avoid design matrix libraries outside of Stan
2. Package up big computation to send it off to a GPU/accelerator
3. Build design matrices if you need them

Now that I got that written out I have some questions hoooray:

1. @jgabry @paul-buerkner @bgoodri does avoiding external design matrices buy us anything from an interface-writing perspective? I thought it would when I wrote this, but I've never actually written a high level regression interface and so I have my doubts.

2. @SteveBronder @t4c1 @rok-cesnovar is this syntax enough to pack things up into big, GPU-friendly calculations?

3. @dpsimpson is it actually useful to build out design matrices in the way I've written here? I remember the sparse stuff you talked about we do linear algebra on them, but maybe this syntax isn't enough to build the matrices you need and you're left having to pass them in from the outside anyway.

Rendered markdown is [here](https://github.com/bbbales2/design-docs/blob/feature/0026-regression-syntax/designs/0026-stan-regression-syntax.md)

### DIFF
--- a/designs/0026-stan-regression-syntax.md
+++ b/designs/0026-stan-regression-syntax.md
@@ -1,0 +1,256 @@
+- Feature Name: Stan Regression Syntax
+- Start Date: 2020-09-21
+- RFC PR:
+- Stan Issue:
+
+# Summary
+[summary]: #summary
+
+The goal is to add a regression syntax to Stan similar to that offered by the high level interfaces (rstanarm and brms, and historically lm/glmer/lme4, etc.).
+
+# Motivation
+[motivation]: #motivation
+
+Adding a regression syntax directly to Stan should make it easy to:
+
+1. Develop high level Stan interfaces that expose regression-like syntaxes without depending on external design matrix or sparse matrix libraries
+
+2. Express likelihood evaluations in a way that can be automatically parallelized and sent to GPUs/other accelerators
+
+3. Manipulate design matrices directly in Stan to do things like integrate out random variables
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The basic multilevel regression syntax from rstanarm/brms/inla/lme4/etc. is written like:
+
+```response ~ linear_model```
+
+The response variable is the variable being modeled, the `~` indicates this expression is going to form a regression, and the `linear_model` is a representation of all the covariates, factors, and variables that go into predicting the response on the left.
+
+Right off the bat this syntax is a little different than Stan code because there is no distribution. For instance, a normal likelihood term in Stan looks like:
+
+```y ~ normal(mu, sigma)```
+
+The regression model above leaves the distribution undefined, and usually the `linear_model` only defines a certain part of the model. In terms of Stan syntax this might look like:
+
+```response ~ ?(linear_model, ?)```
+
+where the `?` terms are undefined.
+
+Different regression software packages make different assumptions about the choice of likelihood and link functions and how extra parameters of the model are defined. This is true too for priors, which are either inferred automatically or specified separately from the linear model.
+
+Stan already provides a way to express priors, likelihoods, and link functions though, and so the key to a regression syntax in Stan is a mechanism to support the `linear_model` part of the regression.
+
+## Linear Models
+
+brms breaks down the linear part of the model into a few terms:
+
+```linear_model = population_level_terms + group_level_terms + smooth_terms```
+
+This proposal is going to ignore the `smooth_terms` part of the regression (not because they aren't interesting, just because it is a big enough project to work on the first two terms).
+
+The responses in multilevel regression models are arranged into groups. That is why there are both population level terms and group level terms -- the population level terms account for the overall effects and the group level effects account for the group level variation.
+
+A basic regression might be:
+```
+y ~ 1 + x + (1 | group) + (x | group)
+```
+
+The `1 + x` term are the population level effects. The `1` indicates that there should be a population level intercept in the model and the `x` term indicates a population level slope term should be fit.
+
+The `(1 | group) + (x | group)` terms are the group level effects. The `(1 | group)` term indicates there should be a group level intercept (a different intercept term for every group) and the `(x | group)` term indicates there should be group level slopes (a different slope term for every group).
+
+The regression implicitly defines population level intercept and slope parameters, intercept and slope parameters for each group, and any hierarhical parameters to go along with these. The implicitly defined variables is something that a Stan syntax would not have; all parameters are explicitly defined in Stan.
+
+Though the statistical dichotomy in the problem are the two types of terms, the population level effects and the group level effects, in terms of computation, evaluating a linear model like this splits into a dense matrix vector multiply and a sparse matrix vector multiply. The population intercept and continuous population effects can be handled efficiently in a dense matrix vector multiply, and the population level factors variables and all the group level terms can be handled efficiently in a sparse matric vector multiply. This is how the linear model is formulated in the [lme4 paper](https://cran.r-project.org/web/packages/lme4/vignettes/lmer.pdf).
+
+There are numerous other syntaxes in various regression packages in R. For instance, `(1 + x | group)` implies a correlated prior on the group level slopes and intercepts in lme4, but `(1 + x || group)` does not. `x - 1` means do not fit a population level intercept. For a syntax in Stan, where parameters must be explicitly defined, this extra syntax will not be necessary.
+
+One thing not mentioned so far are the higher order groupings. Terms like `(x | group1:group2)` indicate that the slopes should vary according to the unique combination of `group1` and `group2` labels on each response (so a combined grouping of `group1` and `group2`). The aims of this proposal can be met with a syntax that supports the terms in the regression:
+
+```
+1 + x + (1 + x | group1) + (1 + x | group1:group2)
+```
+
+## Possible Implementation
+
+The above syntax are already supported in Stan. `brms` and `rstanarm` already use Stan code directly. The problem is, while this syntax is fairly routine Stan code, it does not make it easy to meet the goals laid out in the beginning.
+
+1. Interface developers are still dependent on design matrix libraries (that may differ across languages)
+
+2. While perhaps efficient for a CPU, the code is unlikely to be efficient for a GPU or accelerator
+
+3. The cleanest way to express the sparse matrix-vector multiply in Stan are for loops which do not make it easy to get design matrices out.
+
+For clarity, the example model from above can be written in Stan syntax. Here is the regression model:
+```
+y ~ 1 + x + (1 + x | group1) + (1 + x | group1:group2)
+```
+
+Assume there are `N` responses, `G1` unique values of group 1, and `G2` values of group 2.
+
+The data for the Stan model would look like:
+
+```
+int N; // number of responses
+int G1; // number of unique members of group 1
+int G2; // number of unique members of group 1
+vector[N] y; // responses
+vector[N] x; // covariates
+int<lower=1, upper=G1> group1_idx[N]; // group 1 membership of the responses
+int<lower=1, upper=G2> group2_idx[N]; // group 2 membership of the responses
+```
+
+The parameters would look like:
+
+```
+real intercept; // population level intercept
+real slope; // population level slopes
+vector[G1] group1_intercepts; // group 1 intercepts
+vector[G1] group1_slopes; // group 1 slopes
+matrix[G1, G2] group12_intercepts; // group1:group2 intercepts
+matrix[G1, G2] group12_slopes; // group1:group2 slopes
+```
+
+The linear model can either be evaluated in a loop:
+
+```
+vector[N] linear_model;
+for(n in 1:N) {
+  linear_model[n] = intercept + slope * x[n] +
+    group1_intercepts[group1_idx[n]] + group1_slopes[group1_idx[n]] * x[n] +
+    group12_intercepts[group1_idx[n], group2_idx[n]] + group12_slopes[group1_idx[n], group2_idx[n]] * x[n];
+}
+```
+
+or it can be written in a vectorized form:
+
+```
+vector[N] linear_model = intercept + slope * x +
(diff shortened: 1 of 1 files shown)


seantalts:
Tagging @enetsee who is also interested in this topic!

dpsimpson:
Short version. If we stick to just lme4 stuff initially, there are two
cases;
- Don't marginalize out anything, in which case probably dense is enough
- Marginalize out stuff, in which case we need to be able to do sparse
matrix stuff (like in the lme4 paper for the Gaussian case, or similarly
with Laplace approximations).

This is an extremely cool idea!! and it would be very nice to our Python
users (but also to our R users, who currently translate from brms and then
modify, which is messy).

And obviously we can add cool things like GPs (dense) and spatial models
(often sparse), and smoothing splines (often sparse), and time series
models (sparse). It would be ideal if we could keep as much of the brms
notation as possible for these complex models...

t4c1:
So this is just new syntax for something that is already possible? While I am more familiar with the existing syntax I guess that is fine if that syntax is closer to majority of Stan users.

I have to say I do not see what is the point of the design matrix stuff. To me it just looks like forcing sparse algebra, which will probably be slower that dense version. Are there any advantages in terms of usability?

Anyway, the dense should work fine with GPUs. With kernel generator it won't be hard to implement - we can just write something very similar to the vectorized form of the model. Although the CPU implementation will be hard to optimize, due to the version of Eigen we are using not supporting indexing with vectors. On the other hand we do not have any GPU implementations for sparse operations yet.

dpsimpson:
From my point of view, this isn’t a speed thing. It’s shifting something
that is currently done separately for each interface (ie a thing that is
only available in RStan) into the language so that there is consistent
access to it.

It also lets us build a brms-like higher lever interface for all of the
versions of Stan, which should help users immensely.

A thing that is annoying is that that means the new syntax will need to
consistently support factors, which is the thing that model.matrix in R
does really well.

enetsee:
@seantalts - thanks for the nudge!

@bbbales2, as Sean mentioned, I'm very interested in this topic and is a use case I had in mind when we were looking at the structure of the compiler. Rather than adding this as _new_ syntax as Stan, the idea was to have an alternative frontend just for regression models which elaborated to core Stan or the intermediate representation we use in the compiler (i.e. we have an additional step where we recover anything that was left implicit in the terser language of regression).

There is some really interesting work by Andy Gordon on a 'regression calculus' [here](https://www.microsoft.com/en-us/research/publication/fabular-regression-formulas-as-probabilistic-programming/) (which is just a fancy way of describing a core language for regression along with its meaning) which I was keen on extending to add BRMS like functionality.  I spoke with Andy Gordon about this briefly at StanCon in Cambridge but I haven't had too much time to follow up of late. 

There was also a presentation at StanCon of another front end language, [Chronikis](https://github.com/adobe/Chronikis)  which was targeted at time series models and again, elaborated to Stan.

In terms of motivation, I think we have to recognize that syntax is important: it facilitates the expression of recurring patterns and allows us to restrict expressiveness. With specialized front-end we can also introduce stricter semantics allowing us to pick up errors more easily. The very fact that BRMS has a large user base is evidence of the utility of this approach.

I would be very keen to get involved with this from a language design perspective.

paul-buerkner:
I agree this is a very cool idea, and I am happy to provide my expertise with brms in making this possible. Essentially, I do see that a little bit like (a subset of) brms but directly in Stan with potentially much more opportunities to optimize the code on a lower level (brms does try to write optimized Stan code but of course has only access to the top-level of the Stan language itself). Additional advantages would be that we would get methods such as `posterior_predict` and `log_lik` (using our R names right now) for these models directly in Stan, which is otherwise a little ugly to do in the generated quantities block. 

There are obvious challenges of course but also a few perhaps not as obvious ones, for example, how to deal with predicting new data (which is a quite relevant aspect in brms users' workflows) or how to handle predictions for new levels of the groupings factors. Ideally, we could even achieve some sort of marginalization not only for estimation but also for predictions (e.g., as discussed in https://github.com/paul-buerkner/brms/issues/489). Some of these features may go to far for the Stan regression syntax but would definitely improve its usability a lot.

designs/0026-stan-regression-syntax.md
```diff
@@ -0,0 +1,256 @@
...
+
+```linear_model = population_level_terms + group_level_terms + smooth_terms```
+
+This proposal is going to ignore the `smooth_terms` part of the regression (not because they aren't interesting, just because it is a big enough project to work on the first two terms).
+
+The responses in multilevel regression models are arranged into groups. That is why there are both population level terms and group level terms -- the population level terms account for the overall effects and the group level effects account for the group level variation.
+
+A basic regression might be:
+```
+y ~ 1 + x + (1 | group) + (x | group)
```
SteveBronder, line 57: For the `1` in the random effects part what would the generated name be? Just like `[group name]_intercept`?

designs/0026-stan-regression-syntax.md
```diff
@@ -0,0 +1,256 @@
...
+
+```
+covariate .* (parameters | group1, group2)
+```
+
+where `parameters` is now a 2d variable. To get the design matrices, the `parameters` variable could be removed.
+
+So if the following returns a length `N` vector:
+```
+covariate .* (parameters | group)
```
SteveBronder, line 169: Is this always going to be a `.*`? If so I think this would be easier to work with in the compiler and C++ code if we directly put the covariates in the () like 

```
(covariate, parameters | group)
```

designs/0026-stan-regression-syntax.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+In this case `coefficients_matrix` would need compressed to a vector to make the sparse matrix vector product make sense, whereas `dmat2(coefficients_matrix)` can mean the effective matrix vector product without worrying about flattening `coefficients_matrix`. An extra function `flatten` could be provided to do this reordering if necessary.
+
+With this syntax, the regression from above could be rewritten:
+```
+vector[N] mu = intercept + slope * x +
+    (group1_intercepts | group1_idx) + x .* (group1_slopes | group1_idx) +
+    (group12_intercepts | group1_idx, group2_idx) +
+    x .* (group12_slopes | group1_idx, group2_idx);
```
SteveBronder, line 207: Looking at this I'm not really sure exactly how we'd optimize this on the backend. I kind of like something similar to your original proposal for a `hiearchy()` function for something like 

```
vector[N] mu = intercept + slope * x + hierarchy(f, group1_idx, group2_idx, group12_intercepts, x, group12_slopes)
```

Where `f` is a function with a signature like `vector f(real intercept, matrix x, vector slope)` that would be applied to each group (in this case the group is the combo of groups 1 and 2).

I wonder if we could do anything useful if we have a `group_index` type in the stan language? Or could we give matrices an index? Maybe something like 

```
group_index[N] group_idx1;
group_index[N] group_idx2;
// Anything after first two dims for a matrix are indices?
matrix[N, M, group_idx1, group_idx2] X;
// Or maybe a tag?
matrix[N, M] X index(group_idx1, group_idx2);
```
Though I think the issue with what I have here is that we would have to pay for a sort / unsort if the data was not presorted so we could iterate over the groups nicely

SteveBronder (commented):
Couple little comments. My main Qs are mostly around how we would use this syntax to optimize things on the backend. @t4c1 if you check out the lmer paper they go into details on why the sparse matrix representation can be more efficient for large numbers of random effects. Though I do wonder if there is a way we can keep this dense while being efficient

https://cran.r-project.org/web/packages/lme4/vignettes/lmer.pdf

andrewgelman:
Interesting!  I'd love to be involved in some of this discussion.  Also could someone tag Bob Carpenter (I'm not sure how to do that), as he and Matt Hoffman and I were talking a lot about these issues back in 2011.  Also Jonah/BenG/Aki, because this stuff would be directly relevant to our forthcoming book on applied regression and multilevel models.
Andrew

SteveBronder:
@bob-carpenter

designs/0026-stan-regression-syntax.md
bbbales2: Yes, that seems good. I am nervous the `.*` there was actually more specialized than a `.*` in the wild might be, if that makes sense. The Fabular thing that @enetsee linked above does something like:

```
(covariate {parameters} | group)
```

Not opposed to something else either.

designs/0026-stan-regression-syntax.md
bbbales2: It would depend on the package, but I'd expect something like that (and for it to be a vector with the number of elements of the group).

designs/0026-stan-regression-syntax.md
bbbales2: > I'm not really sure exactly how we'd optimize this on the backend

I imagined this would be something recognizable by the compiler. Like, "oh, I have a bunch of design matrices, better lump them together and turn them into a big `hierarchy` statement".

But also I think @t4c1 likes expressions and I could see that working too, so it kinda all comes together at runtime. As long as it gets into a big `hierarchy`-like statement it's good.

> group_index type in the stan language
> we would have to pay for a sort / unsort if the data

Is the goal here to express a sparseness or to express the ordering of the matrix to make accesses efficient in some way?

bbbales2:
> And obviously we can add cool things like GPs (dense) and spatial models
(often sparse), and smoothing splines (often sparse), and time series
models (sparse). It would be ideal if we could keep as much of the brms
notation as possible for these complex models...

@dpsimpson Yeah I like the `y ~ gp(x)` sorta thing that brms does.

Separately I would like some sort of Stan library system with a package manager. I think that's the only practical way we're going to get all the fancy design matrices and stuff that would need built for splines, approx GPs, etc in Stan models. Pumping that stuff through Stan math would be too slow development-wise. I think if we had a more formal/centralized way for people to share their code, we'd see a lot of interesting code getting shared (and avoid everyone having to reinvent things, digging in forums for code snippets, etc.).

> So this is just new syntax for something that is already possible?
> Are there any advantages in terms of usability?

@t4c1 Yeah, having the design matrices and being able to evaluate/use them efficiently is a plus. Part of the goal is usability and part of the goal is to make it easier for the compiler/runtime to recognize that a bunch of calculations are related. A little of both.

> There is some really interesting work by Andy Gordon on a 'regression calculus'
> another front end language, Chronikis

@enetsee The Fabular thing was interesting. Yes this is very much what I want to see in Stan. I like how they talk about breaking the regressions out into different levels. I also like the variables being explicit, but I'm probably not considering some advantage of them being implicit.

I had a look at Chronikis. I was kinda thinking about domain-specific Stan packages (alternative frontends like you're saying, and I talked to @imadmali about one). The problem with front end languages is I don't know compiler stuff and so the extent of compiling I want to do is Python format-strings sorta code generation.

The discourse thread at the top (this [one](https://discourse.mc-stan.org/t/speeding-up-hierarchical-models/14783)) has sortof a brute-force giant function that we could use for evaluating the right hand side of a hierarchical regression. This would be something we could compile a higher level language to.

The downsides to this are it's not terribly easy to read the syntax, we don't get the design matrices themselves, and there's no hierarchy of regressions like in the regression calculus paper. I don't want to clutter the Stan language up with too much extra syntax, but I also wouldn't want to limit it to a super-conservative syntax either (cuz the intention is lots of people happily use Stan and the interface languages -- not filter everyone from Stan to interfaces).

> There are obvious challenges of course but also a few perhaps not as obvious ones

@paul-buerkner Could we talk about the workflow stuff on this sometime? I am rather scared of forgetting something here. Maybe in a couple weeks after the 2.25 release (on the 19th)? @enetsee or @andrewgelman or anyone else if you want to come to this, that's great. I'll update the design doc with whatever we talk about.

paul-buerkner:
Happy to have a chat about this. Just let me know here or via email.

Am So., 4. Okt. 2020 um 23:57 Uhr schrieb Ben Bales <
notifications@github.com>:

bbbales2:
@paul-buerkner @enetsee @andrewgelman are you free any time in the next couple weeks to talk about the not-so-obvious challenges here?

paul-buerkner:
I would be flexible on most days as well. Feel free to suggest a time.

bbbales2:
@paul-buerkner let's plan on before the Stan meeting then. That's 2PM UTC Thursday (10AM EST) if I'm doing my math correctly.

@enetsee if you want to come and that doesn't work, pick another time. Sounds like we're both flexible. I'll e-mail a call link before then.

paul-buerkner:
If you want to talk this or next week, would it also work after the Stan meeting? The hour before the Stan meeting is not ideal those two weeks.

bbbales2:
After is good too, so let's say 4PM UTC, noon EST

andrewgelman:
Hi, just curious what happened with this?

bbbales2:
@andrewgelman had a meeting and I never updated the design doc.

There were some things there worth writing down, but the thing that was nagging me when I was trying to update this were unresolved problems with defining parameters that might have different sizes as a function of data.

Problems like, you have an interaction:

```
parameters {
  matrix[N_ages, N_educs] alpha0;
}
```

And then you have a model where in your data you don't have at least one of all the ages and educations.

I sorta got stuck here. And if you don't solve this problem you still leave a lot of annoying shuffling of stuff to the outside.

bbbales2:
I'm not sure I talk about this problem in the design doc, but I think I am assuming that it's solved in some way, and that seems like a big assumption.

PR #28, opened 2020-09-29
parallel_map design doc

andrjohns:
I've written up a design document for the ```parallel_map``` functionality that I've been working on. It provides a new means of within-chain parallelisation that has the user-friendliness of ```reduce_sum```, but without the restriction of returning a sum.

Feedback on the design and implementation are all very much welcome!

Rendered markdown [here](https://github.com/andrjohns/design-docs/blob/feature/parallel_map/designs/0028-parallel_map.md)

Stan Math branch [here](https://github.com/stan-dev/math/compare/develop...andrjohns:feature/map)

### DIFF
--- a/designs/0028-parallel_map.md
+++ b/designs/0028-parallel_map.md
@@ -24,7 +24,7 @@ for (int i = 0; i < 10000; ++i) {
 }
 ```
 
-There are two key components to parallelising this loop via ```parallel_map```: an indexing function and an computation/application function.
+There are two key components to parallelising this loop via ```parallel_map```: an indexing function and a computation/application function.
 
 ## Index Function
 The indexing function is the workhorse of the framework, as it allows ```parallel_map``` to manipulate only the variables that are needed by a given thread. This is in contrast to ```reduce_sum```, which will only 'slice' one argument and copy the rest in full to all threads. For the above loop we want to loop over both arguments (```a``` and ```b```), this means that the indexing function will be:
@@ -34,7 +34,7 @@ auto index_fun = [&](int i, const auto& fun, const auto& x, const auto& y) {
 };
 ```
 
-As you can see, this approach allows for a great deal of flexibility in the indexing of arguments. For example, if in the loop above, ```a``` was a vector of size 10000, and ```b``` was a scalar to broadcast to each element of ```a```, the indexing function would instead be:
+As you can see, this approach allows for a great deal of flexibility in the indexing of arguments. For example, if in the loop above ```a``` was a vector of size 10000 and ```b``` was a scalar to broadcast to each element of ```a```, the indexing function would instead be:
 ```
 auto index_fun = [&](int i, const auto& fun, const auto& x, const auto& y) {
   return fun(x[i], y);
@@ -56,7 +56,7 @@ auto index_fun = [&](int i, int j, const auto& fun, const auto& x, const auto& y
 ```
 
 ## Apply Function
-Returning to the original loop, once the indexing function has been specified, the next step is to specify the function to be applied to the indexed arguments. This function should be specified assuming that the arguments have already been indexed (i.e., assuming that the inputs are scalars, rather than vectors/matrices that need to be indexed). For the above loop, this function would be:
+Returning to the original loop, once the indexing function has been specified the next step is to specify the function to be applied to the indexed arguments. This function should be specified assuming that the arguments have already been indexed (i.e., assuming that the inputs are scalars, rather than vectors/matrices that need to be indexed). For the above loop, this function would be:
 
 ```
 auto app_fun = [&](const auto& x, const auto& y) {
@@ -124,6 +124,6 @@ This framework is similar to (and heavily inspired by) ```reduce_sum```, as both
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-- Should we change the name from ```parallel_map```? Sebastian raised the point that we don't preface any other parallel functionality with "parallel"
+- Should we change the name from ```parallel_map```? Sebastian raised the point that we don't prefix any other parallel functionality with "parallel"
 - Is it going to be feasible to handle the index function in Stan? In c++ this needs to have an ```auto``` return type so that it can be used for copying ```varis```, ```vars```, and applying the computation function. How will this work with Stan where a return type needs to be specified for functions?
 - Any other ideas for improving performance and usability are also very welcome
--- a/designs/0028-parallel_map.md
+++ b/designs/0028-parallel_map.md
@@ -1,0 +1,129 @@
+- Feature Name: parallel_map
+- Start Date: 2020/09/29
+- RFC PR: (leave this empty)
+- Stan Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The proposed ```parallel_map``` functionality provides a means of within-chain parallelisation that does not require the sum of iterations be returned (i.e., ```reduce_sum```) or any complex argument-packing (i.e., ```map_rect```).
+
+# Motivation
+[motivation]: #motivation
+
+While the ```reduce_sum``` framework has been very popular for its ease of use, it's limited to situations where the goal is to return the sum of independent computations. Many users have models with large functions/computations that return vectors or matrices in full, rather than their sum. For these cases, users seeking parallelisation are limited to the ```map_rect``` functionality, which has a non-zero barrier to entry due to the need for argument packing and unpacking. The ```parallel_map``` framework aims to provide a means of parallelising these operations in a manner which is easy for users to understand and implement.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Target Loop
+As an example, let's say the user wanted to calculate the log of the beta function from two vectors of size 10000:
+```
+for (int i = 0; i < 10000; ++i) {
+  result[i] = lgamma(a[i]) + lgamma(b[i]) - lgamma(a[i] + b[i])
+}
+```
+
+There are two key components to parallelising this loop via ```parallel_map```: an indexing function and an computation/application function.
+
+## Index Function
+The indexing function is the workhorse of the framework, as it allows ```parallel_map``` to manipulate only the variables that are needed by a given thread. This is in contrast to ```reduce_sum```, which will only 'slice' one argument and copy the rest in full to all threads. For the above loop we want to loop over both arguments (```a``` and ```b```), this means that the indexing function will be:
+```
+auto index_fun = [&](int i, const auto& fun, const auto& x, const auto& y) {
+  return fun(x[i], y[i]);
+};
+```
+
+As you can see, this approach allows for a great deal of flexibility in the indexing of arguments. For example, if in the loop above, ```a``` was a vector of size 10000, and ```b``` was a scalar to broadcast to each element of ```a```, the indexing function would instead be:
+```
+auto index_fun = [&](int i, const auto& fun, const auto& x, const auto& y) {
+  return fun(x[i], y);
+};
+```
+
+Further, it's not required that the indexing function return scalars, *as long as the result of the computation/application function is a scalar*. This means that given two matrices, for example, the indexing function could be:
+```
+auto index_fun = [&](int i, const auto& fun, const auto& x, const auto& y) {
+  return fun(x.row(i), y.col(i));
+};
+```
+
+Additionally, ```parallel_map``` supports looping over two indexes. If ```a``` and ```b``` were both matrices then the index function would use two indexes (row & column):
+```
+auto index_fun = [&](int i, int j, const auto& fun, const auto& x, const auto& y) {
+  return fun(x(i, j), y(i, j));
+};
+```
+
+## Apply Function
+Returning to the original loop, once the indexing function has been specified, the next step is to specify the function to be applied to the indexed arguments. This function should be specified assuming that the arguments have already been indexed (i.e., assuming that the inputs are scalars, rather than vectors/matrices that need to be indexed). For the above loop, this function would be:
+
+```
+auto app_fun = [&](const auto& x, const auto& y) {
+  return lgamma(x) + lgamma(y) - lgamma(x + y);
+};
+```
+
+## parallel_map Call
+The final step before applying these functions to the inputs is to declare the output/result container that each iteration of the loop should be evaluated into. A limitation of the framework is that this result container needs to be declared beforehand and passed to the ```parallel_map``` call, as while it is simple to deduce the return type it is not (yet) possible to deduce the size of the return type.
+
+Once the result container has been declared, we can then call ```parallel_map```:
+```
+parallel_map(app_fun, index_fun, std::forward<result_type>(result), grainsize,
+             std::forward<a_type>(a), std::forward<b_type>(b))
+```
+
+Alternatively, if the loop to be parallelised is over two indexes (i.e., matrices), then both a row-wise and column-wise grainsize need to be provided to ```parallel_map```:
+```
+parallel_map(app_fun, index_fun, std::forward<result_type>(result), row_grainsize,
+             col_grainsize, std::forward<a_type>(a), std::forward<b_type>(b))
+```
+
+Once ```parallel_map``` has finished executing the loop, the result container (```result``` above) will contain the results from each iteration.
+
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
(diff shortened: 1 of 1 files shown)


designs/0028-parallel_map.md
```diff
@@ -0,0 +1,129 @@
...
+```
+for (int i = 0; i < 10000; ++i) {
+  result[i] = lgamma(a[i]) + lgamma(b[i]) - lgamma(a[i] + b[i])
+}
+```
+
+There are two key components to parallelising this loop via ```parallel_map```: an indexing function and a computation/application function.
+
+## Index Function
+The indexing function is the workhorse of the framework, as it allows ```parallel_map``` to manipulate only the variables that are needed by a given thread. This is in contrast to ```reduce_sum```, which will only 'slice' one argument and copy the rest in full to all threads. For the above loop we want to loop over both arguments (```a``` and ```b```), this means that the indexing function will be:
```
bbbales2, line 30: Can you give more examples of the indexing function? Is this something we should adopt for `reduce_sum` as well?

Perhaps write some Stan psuedo-code? Like parallelize this code:

```
vector[N] mu = ...;
vector[G] effects = ...;
int[N] idxs = ...;
vector[N] y = ...;
vector[N] lp;
for(i in 1:N) {
  lp[i] = normal_lpdf(y[i] | mu[i] + effects[idxs[i]]);
}
```

The problem with the basic `reduce_sum` version of this is that every chunk of work ends up copying all the partials in `effects`. If we could get around that with fancy indexing that would be nice.

Anyway I like the indexing. Indexing is the biggest limiter to `reduce_sum` I think, so if there's away around this that's great.

designs/0028-parallel_map.md
```diff
@@ -0,0 +1,129 @@
...
+
+# Summary
+[summary]: #summary
+
+The proposed ```parallel_map``` functionality provides a means of within-chain parallelisation that does not require the sum of iterations be returned (i.e., ```reduce_sum```) or any complex argument-packing (i.e., ```map_rect```).
+
+# Motivation
+[motivation]: #motivation
+
+While the ```reduce_sum``` framework has been very popular for its ease of use, it's limited to situations where the goal is to return the sum of independent computations. Many users have models with large functions/computations that return vectors or matrices in full, rather than their sum. For these cases, users seeking parallelisation are limited to the ```map_rect``` functionality, which has a non-zero barrier to entry due to the need for argument packing and unpacking. The ```parallel_map``` framework aims to provide a means of parallelising these operations in a manner which is easy for users to understand and implement.
```
bbbales2, line 14: > large functions/computations that return vectors or matrices in full

Returning anything but scalars and I'm gonna complain :P. Precomputing gradients with lots of outputs is not efficient with reverse mode. To do that efficiently will require parallel reverse mode.

Of course, if you want to do mess with parallel reverse mode I'm all for it!

bbbales2 (commented):
Just noticed this. Feel free to ping in the future. Couple questions!

designs/0028-parallel_map.md
andrjohns: Oh good point on the pseuco-code. I'll update the doc proper later this week, but in the interim, given the Stan model:
```
data {
  int[N] idxs;
  vector[N] y;
}

parameters {
  vector[N] mu;
  vector[G] effects;
  vector<lower=0>[N] sigma;
}

model {
  vector[N] lp;

  for(i in 1:N) {
    lp[i] = normal_lpdf(y[i] | mu[i] + effects[idxs[i]], sigma[i]);
  }
}
```

This would be expressed with ```parallel_map``` via:
```
functions {

  // Return is TBD as need to co-ordinate with Stanc3 to have this
  // transpile to an auto return
  TBD index_fun(int i, F fun, vector y, vector mu, vector effects,
                vector idxs, vector sigma) {
    return fun(y[i], mu[i], effects[idxs[i]], sigma[i]);
  }

  // effects will already have been indexed, so don't
  // need idxs array here
  real apply_fun(vector y, vector mu, vector effects) {
    return normal_lpdf(y | mu + effects, sigma);
  }
}

data {
  int[N] idxs;
  vector[N] y;
  int grainsize;
}

parameters {
  vector[N] mu;
  vector[G] effects;
}

model {
  vector[N] lp;

  parallel_map(apply_fun, index_fun, lp, grainsize,
               y, mu, effects, idxs, sigma);
}
```

andrjohns: > Can you give more examples of the indexing function? Is this something we should adopt for reduce_sum as well?

I'm a bit biased because I think the approach is neat, but I think it would expand the use-cases for ```reduce_sum``` by quite a bit. I've had a few situations where I've wanted to use it, but the parameters were too large for the copy-cost to be worth the parallelisation.

designs/0028-parallel_map.md
andrjohns: > Precomputing gradients with lots of outputs is not efficient with reverse mode

Benchmarking has shown that it's definitely not great in all cases, but it does provide some speed-up. I still need to add some more robust benchmarking to this proposal though.

> Of course, if you want to do mess with parallel reverse mode I'm all for it!

Haha would not know where to start! I'd be interested in tinkering with it though. Any chance there's a writeup somewhere of where the parallelism needs to go in? Or if you're feeling particularly generous could we set up a hangouts to go over the current autodiff and what needs to be built out for parallel reverse mode?

andrjohns: Just found @wds15 [design doc](https://github.com/wds15/design-docs/blob/parallel_autodiff/designs/0003-parallel_autodiff.md) and [math branch](https://github.com/stan-dev/math/compare/parallel-ad-tape-3) for the parallel autodiff, will start reading up on those

designs/0028-parallel_map.md
bbbales2: > but I think it would expand the use-cases for `reduce_sum`

Well I was thinking deprecating reduce sum if this is just better.

The thing `reduce_sum` does that this doesn't is work with ranges so that the parallel work can be larger. Maybe this can work with ranges too?

You'll need to make copies of the shared arguments or do atomic operations on the adjoint accumulates.

I think because the indexing is custom, you'd need compiler help to figure out what to copy. For instance:
```
return fun(y[i], mu[i], effects[idxs[i]], sigma[i]);
```

1. `effects[idxs[i]]` could access the same element of effects at two different `i` s, so there's a race condition on the reverse pass
2. Before evaluating this statement, you won't know what to copy for each chunk of work
3. So somehow I think you'd need to parse this and put compiler stuff in there to know what to copy. Like:

`fun(deep_copy(y[i]), deep_copy(mu[i]), deep_copy(effects[idxs[i]]), deep_copy(sigma[i]))`

andrjohns: That's actually the main attraction of the indexing function, that it's used for deep copying the vars and varis.

Going back to the c++ side of things, the indexing function is defined as:
```
auto [&](int i, const auto& fun, const auto& y, const auto& mu,
         const auto& effects, const auto& sigma, const auto& idxs) {
    return fun(y[i], mu[i], effects[idxs[i]], sigma[i]);
  }
```

This is then used to deep copy just the vars at that iteration, using:
```
    auto var_copier = [&](const auto&... xargs) {
      return std::tuple<decltype(deep_copy_vars(xargs))...>(
        deep_copy_vars(xargs)...);
    };

...

    tbb::parallel_for(
      tbb::blocked_range<size_t>(0, S, grainsize),
      [&](const tbb::blocked_range<size_t>& r) {
        // Run nested autodiff in this scope
        const nested_rev_autodiff nested;

        for (size_t i = r.begin(); i < r.end(); ++i) {

        ...
        auto args_tuple_local_copy = index_fun(i, var_copier, x...);
        ...
        }
      });
```

> effects[idxs[i]] could access the same element of effects at two different i s, so there's a race condition on the reverse pass

If the vars are deep copied, is there a race condition?

[Full implementation here](https://github.com/andrjohns/math/blob/b57ac53f1dcff54ee6c7acd3b36b1dda0ff877d7/stan/math/rev/functor/parallel_map.hpp)

designs/0028-parallel_map.md
bbbales2: > Benchmarking has shown that it's definitely not great in all cases

The problem with multiple outputs (and the reason I'm so allergic to it), is I'm scared it makes it possible for people to write slow code.

What we're doing is computing the Jacobian of a function `f` using reverse mode autodiff and then embedding that in a forward pass.

On paper, if you have a function with N inputs and M outputs, the Jacobian is an MxN matrix and it takes M reverse mode passes or N forward mode passes to compute it.

In this case, parallel_map is a bunch of evaluations of a little function `g`, so something like $f = [g_1, g_2, ...]$

Because each of the `g` functions in the parallel_map calculation don't share calculations, we can separate out the adjoints on the input, and we can do all these M reverse mode passes in parallel.

If each of the `g` functions has multiple outputs though, those will not be parallelizable and you'll need to do multiple reverse mode passes.

The caveat to this is if there is a function where the forward pass is slower than the reverse pass, you can get away with multiple reverse passes because they are cheap. The ODE solvers are currently like this, all the distributions are like this, and some of the linear algebra stuff has different complexity for the forward/reverse passes.

I don't like it because someone needs to guess and check whether their model will be faster and they can accidentally write code that is slower than the serial version (not due to parallel overhead -- there's just more work in computing all the Jacobians). This is how we ended up with the current single-output focused `reduce_sum` design.

> would not know where to start

The problem with parallel reverse mode is how to handle race conditions if multiple `chain` functions are being evaluated in parallel, so that's where to start thinking. Yeah we can have a meeting whenever. @wds15 probably would like in too.

The larger problem of parallel is a lot of models are memory bound, and so we won't get faster unless we do something parallel across multiple computers. It's not something that needs written at first, but I want to keep multiple-computers in mind for the design.

designs/0028-parallel_map.md
bbbales2: > If the vars are deep copied, is there a race condition?

Nope! If they're getting deep copied, that's that, you have your own vari and the vari s can be propagated back from there as necessary.

bbbales2: Ooo, I see, the `fun` in `index_fun` is deep copying. I thought the fun was the user's function. Yes that makes sense and also the auto return type makes sense now.

andrjohns: Yeah it's used in a few contexts:

Counting vars:
```
int nvars = index_fun(0, var_counter, x...);
```

Copying varis:
```
index_fun(i, vari_saver(&vari_map[i]), x...);
```

Deep-copying vars:
```
auto args_tuple_local_copy = index_fun(i, var_copier, x...);
```

Applying function to an iteration (```prim``` only, ```rev``` applies function to copied vars):
```
result[i] = index_fun(i, app_fun, x...);
```

designs/0028-parallel_map.md
andrjohns: Thanks for writing that all out, super helpful. I'll mull this over for couple days and try to get my head as far around it as I can. Maybe hang around after the general meeting? @wds15 work for you?

> The larger problem of parallel is a lot of models are memory bound, and so we won't get faster unless we do something parallel across multiple computers. It's not something that needs written at first, but I want to keep multiple-computers in mind for the design.

I don't know if we'll get that through the TBB, so that might need a different implementation all together

bbbales2: > Maybe hang around after the general meeting?

Works for me

> through the TBB

Yeah it wouldn't be the TBB.

wds15: My apologies for not yet having digged into this, but it sounds like the main points in my mind are already hammered on (in a good way):

- The index function concept is cool and if we can make it happen here we should think about putting it into `reduce_sum_v2` or whatever.
- The index function thing should ideally also work for ranges.
- What is unclear to me is how the index function stuff can actually work with the Stan language... and in case the index function thing is a major project on its own, then maybe the prospect of tuples makes the need for an index function obsolete (with compromises)?

I do think that the return type should be anything - even though users can easily slow down their program due to the large Jacobians. There are sufficient applications where the memory boundness is not a problem - like ODE solving, mixture models, joint GP's where multiple covariance matrices are constructed (if that is done in practice).

For truly parallel reverse mode AD, one would probably come up with something which leverages information from the compiler in order to know upfront how to manage the synchronization.

On Thursday after the Stan meeting should work.

wds15:
@bbbales2 Is the `reverse_pass_callback` thing something which could serve as an idea to dynamically build Intel TBB dependency graphs? I mean, if we can dynamically build out a dependency graph, then we can let the TBB figure out how to run it in parallel.

(other than another parallel idea, we should probably move this part of the discussion somewhere else)

andrjohns:
I like the dependency graph idea. If I'm understanding that (and the TBB doc) right, would that be having the ```chain()``` method return a function node, and then the ```grad()``` call collects the function nodes into a single graph and executes?

wds15:
I think the var objects would be nodes of the graph and the vari operandi would represent edges. The graph must represent the dependency structure of the reverse pass. Each node then performs a chain call...but I am not sure if that can fly given that the tbb graph thing may be far more heavy than the var and vari what we use now, but maybe this can work in some form.

andrjohns:
Just to make sure that I'm on the same page here, is the goal basically that we want to be able to evaluate the chain calls for the var stack in parallel, rather than sequentially iterating through it?

wds15:
Correct. This would mean full parallelism without any special user interaction...but while it is attractive I have doubts about its possibility(even though I suggest that which I do because it would be really cool).

bbbales2:
> graph thing may be far more heavy than the var and vari what we use now, but maybe this can work in some form.

> Just to make sure that I'm on the same page here, is the goal basically that we want to be able to evaluate the chain calls for the var stack in parallel, rather than sequentially iterating through it?

So what Sebastian said, we wouldn't ever do it at the level of a single var, but it might turn out that everything could be made parallel if it's set up right.

At a language level I'm a fan of always doing manual parallelization. That means explicit parallel-fors, etc. I don't think it's worth the hassle to have the compiler look at a for loop or a sequence of commands and try to work out the dependencies and parallelize them.

Now that I'm thinking about it, I keep harping on about parallel reverse mode, but maybe what we need is the ability to allocate nested substacks and autodiff those backwards in parallel? In this way the main stack remains serial, and at some point it tells nested stacks to be parallel.

> I think the var objects would be nodes of the graph and the vari operandi would represent edges.

It's weird, but something Tadej has pointed out, all vars are just vari pointers. The thing that is happening is there are two types of varis, ones for storing variables (nodes) and ones for representing operations (edges) and sometimes they are the same thing (the varis from scalar operations) and sometimes they are not (the varis from matrix operations).

> Is the reverse_pass_callback thing something which could serve as an idea to dynamically build Intel TBB dependency graphs?

I don't have a good enough picture of TBB to say. I hope whatever this ends up being is as reverse_pass_callback-y as possible.

wds15:
> Now that I'm thinking about it, I keep harping on about parallel reverse mode, but maybe what we need is the ability to allocate nested substacks and autodiff those backwards in parallel? In this way the main stack remains serial, and at some point it tells nested stacks to be parallel.

If you will, `reduce_sum` is nothing else than a language construct which tells us the independence structure of the AD tree. This allows us to do it all in parallel, since we know it's safe to do it.

Right now, we also have to do things in an eager way / meaning we have to do nested AD instead of doing a truly parallel reverse sweep over a large AD tree. However, if we would insert operands which can be "chained into" in parallel, then we can run the reverse sweep in parallel. With "chained into" in parallel I mean that the adjoint stored would receive additions to them from multiple chain calls at the same time. This is doable with thread local storage stuff from the TBB. This would imply that all input operands to a parallel region would be copied once into such a special operand... interesting. I need to think a bit about this before confusing others here more probably.

bbbales2:
Oh I see, you're talking about having nested stacks being nodes in any other stack, so there would be like a nested_stack_vari.

If the main stack is serial, then there would still need to be some extra mechanism to do multiple nested_stack_varis in parallel, so maybe those are nested in something else.

seantalts:
Just coming to this via the stanc3 issue. Could you add some words about how this would work with MPI (maybe to the reference-level section)?

bbbales2:
@seantalts we've branched the conversation a bit.

Part of this is a new indexing scheme for the inputs (this is where auto comes from, and if it works out maybe we do something like this for `reduce_sum`), and then part of it is a new parallelization mechanism.

On the new parallelization mechanism, there's a bunch of stuff over here: https://discourse.mc-stan.org/t/older-parallel-ad-tape-ideas/18534

PR #21, opened 2020-05-01
Static matrices

SteveBronder:
This is currently a Work In Progess design document for adding `static_matrix` and `static_vector` types to the Stan language.

You can see a full rendered version [here](https://github.com/SteveBronder/design-docs/blob/static-matrix/designs/0005-static-matrices.md). See the section on Prior art for references to past discussions on this.

@bob-carpenter and @t4c1 can you take a look at this? Feel free to make changes as you see appropriate

PR #29, opened 2020-12-10
[WIP] Static matrix design doc December update

bbbales2:
Note: This is a WIP design doc -- we'll merge it into the main design doc (https://github.com/stan-dev/design-docs/pull/21) later

@SteveBronder @rok-cesnovar @t4c1 I tried to do a rewrite of the design doc focused on setting up the language discussion again.

I think brevity is our friend here, so I didn't avoided examples and implementation details as much as possible. We can always link to the implementation stuff we have now.

[Rendered version](https://github.com/stan-dev/design-docs/blob/static-matrix-dec-update/designs/0005-static-matrices.md)

### DIFF
--- a/designs/0005-static-matrices.md
+++ b/designs/0005-static-matrices.md
@@ -18,7 +18,7 @@ Currently, a differentiable matrix in Stan is represented as an Eigen matrix hol
 
 At the Stan Math level a constant matrix is a `var_value<Eigen::Matrix<double, -1, -1>>` with an underlying pointer to a `vari_value<Eigen::Matrix<double, -1, -1>>`\*.  This means that accessors for the value and adjoint of `var_value` objects can access contiguous chunks of each part of the `vari_value`. Any function that accepts a `var_value<T>` will support static matrices.
 
-At the language and compiler level, a `matrix` can be substituted for a constant matrix if the matrix is only constructed once and never reassigned to.
+At the language and compiler level, a `matrix` can be substituted for a constant matrix if the matrix is only constructed once and it's subslices are never assigned to.
 
 ```stan
 const matrix[N, M] A = // Construct matrix A...
@@ -26,10 +26,9 @@ A[10, 10] = 5; // illegal!
 A[1:10, ] = 5; // illegal!
 A[, 1:10] = 5; // illegal!
 A[1:10, 1:10] = 5; // illegal!
-A = X; // illegal!
 ```
 
-Any function or operation in the stan language that can accepts a `matrix` as an argument can also accept a constant matrix.
+Any function or operation in the stan language that can accepts a `matrix` as an argument can also accept a constant matrix. Functions which can take in multiple matrices will only return a constant matrix if all of the other matrix inputs as well as the return type are static matrices.
 
 
 # Reference-level explanation
@@ -59,11 +58,7 @@ using ChainableStack = AutodiffStackSingleton<vari_base, chainable_alloc>;
 
 7. In the stanc3 compiler, support detection and substitution of `matrix/vector/row_vector` types for constant matrix types.
 
-8. Create a design document for allowing type attributes in the stan language.
-
-9. Add a `const` type attribute to the stan language.
-
-Steps (1) and (2) have been completed in the example branch with some work done on (3). Step (7-9) have been chosen specifically to allow more time to discuss the stan language implementation. The compiler can perform an optimization step while parsing the stan program to see if a `matrix/vector/row_vector`:
+Steps (1) and (2) have been completed in the example branch with some work done on (3). Step 7 has been chosen specifically to allow more time to discuss the stan language implementation. The compiler can perform an optimization step while parsing the stan program to see if a `matrix/vector/row_vector`:
 
 1. Does not perform assignment after the first declaration.
 2. Uses methods that have constant matrix implementations in stan math.
@@ -148,20 +143,5 @@ Discussions:
 [JAX](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-In-Place-Updates) is an autodiff library developed by google whose array type has similar constraints to the constant matrix type proposed here.
 
 [Enoki](https://github.com/mitsuba-renderer/enoki) is a very nice C++17 library for automatic differentiation which under the hood can transform their autodiff type from an arrays of structs to structs of arrays. It's pretty neat! Though implementing something like their methods would require some very large changes to the way we handle automatic differentiation.
-
-# Unresolved questions
-[unresolved-questions]: #unresolved-questions
-
-- What parts of the design do you expect to resolve through the RFC process before this gets merged?
-
-Whether the staged development process listed in the reference level explanation will suffice.
-
-If the restriction on not allowing assignment to the entire matrix such as `A = X` is too restrictive.
-
-- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
-
-Any intricacies of using the GPU via `var_value<matrix_cl<double>>` should be deferred to a separate discussions or can happen during the implementation.
-
-Methods involving changing the current `matrix` type in the Stan language
 
 *Interestingly, this also means that `var_value<float>` and `var_value<long double>` can be legal.
--- a/designs/0005-static-matrices.md
+++ b/designs/0005-static-matrices.md
@@ -25,7 +25,7 @@ A[10, 10] = 5;
 
 Besides that they can be thought of as standard matrix.
 
-At the Stan Math level a `static_matrix` is a `var_type<Eigen::Matrix<double, -1, -1>>` with an underlying pointer to a `vari_type<Eigen::Matrix<double, -1, -1>>`\*. This means that accessors for the value and adjoint of `var_type` objects can access contiguous chunks of each part of the `var_type`. Any function that accept a `var_type<T>` will support static matrices.
+At the Stan Math level a `static_matrix` is a `var_type<Eigen::Matrix<double, -1, -1>>` with an underlying pointer to a `vari_type<Eigen::Matrix<double, -1, -1>>`*. This means that accessors for the value and adjoint of `var_type` objects can access contiguous chunks of each part of the `var_type`. Any function that accept a `var_type<T>` will support static matrices.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
@@ -56,10 +56,10 @@ More templates can be confusing and will lead to larger compile time. The confus
 With `sparse_matrix` and `complex_matrix` this will now add _another_ `*_matrix` type proposal to the language. There's no way to get around this in the Stan language currently, though some small side discussions exist to support attributes on types such as
 
 ```stan
-@(sparse, static)
+@(sparse_type, static_type) // can't have static here because of C++ keyword
 matrix[N, M] X;
 
-@(complex, static)
+@(complex_type, static_type)
 vector[N] Y;
 ```
 
@@ -92,4 +92,4 @@ I'm interested in hearing about Stan language semantics and any difficulties I m
 
 Any intricacies of using the GPU via `var_type<matrix_cl<double>>` should be deferred to a separate discussions or can happen during the implementation.
 
-\*Interestingly, this also means that `var_type<float>` and `var_type<long double>` can be legal.
+*Interestingly, this also means that `var_type<float>` and `var_type<long double>` can be legal.
--- a/designs/0005-static-matrices.md
+++ b/designs/0005-static-matrices.md
@@ -18,7 +18,7 @@ Currently, a differentiable matrix in Stan is represented as an Eigen matrix hol
 
 At the Stan Math level a constant matrix is a `var_value<Eigen::Matrix<double, -1, -1>>` with an underlying pointer to a `vari_value<Eigen::Matrix<double, -1, -1>>`\*.  This means that accessors for the value and adjoint of `var_value` objects can access contiguous chunks of each part of the `vari_value`. Any function that accepts a `var_value<T>` will support static matrices.
 
-At the language and level, a `matrix` can be substituted for a constant matrix if the matrix is only constructed once and never reassigned to.
+At the language and compiler level, a `matrix` can be substituted for a constant matrix if the matrix is only constructed once and never reassigned to.
 
 ```stan
 const matrix[N, M] A = // Construct matrix A...
@@ -37,7 +37,9 @@ Any function or operation in the stan language that can accepts a `matrix` as an
 
 This feature will be added in the following stages.
 
-1. Replace `var` and `vari` with templated types. This has already been done in the example branch [here](https://github.com/stan-dev/math/compare/feature/var-template) (see [here](https://github.com/stan-dev/math/blob/d2967fe2bf6e0d4729d67a714ef40d95d907b18b/stan/math/rev/core/vari.hpp) for the `vari_value` implementation and [here](https://github.com/stan-dev/math/blob/d2967fe2bf6e0d4729d67a714ef40d95d907b18b/stan/math/rev/core/var.hpp) for `var_value`). `var` and `vari` have been changed to `var_value<T>` and `vari_value<T>` with aliases for the current methods as `using var = var_value<double>` and `using vari = vari_value<double>`. These aliases allow for full backwards compatibility with existing upstream dependencies.
+1. Replace `var` and `vari` with templated types.
+   - This has already been done in the example branch [here](https://github.com/stan-dev/math/compare/feature/var-template) (see [here](https://github.com/stan-dev/math/pull/1877/files#diff-130c5a75cc427d7d41715e9fca8281f4R177) for the `vari_value` implementation and [here](https://github.com/stan-dev/math/pull/1877/files#diff-ab13c40c3f03b20efbba9d70d55b4dcdR34) for `var_value`). `var` and `vari` have been changed to `var_value<T>` and `vari_value<T>` with aliases for the current methods as `using var = var_value<double>` and `using vari = vari_value<double>`. These aliases allow for full backwards compatibility with existing upstream dependencies.
+   - The `vari_value` specialized for Eigen types holds two pointers of scalar arrays `val_mem_` and `adj_mem_` which are then accessed through `Eigen::Map` types `val_` and `adj_`.
 
 2. Make the stack allocator conform to accept `vari_value<T>` objects. This is done by having `vari_type<T>` inherit from a `vari_base` class and defining the [`ChainableStack`](https://github.com/stan-dev/math/blob/d2967fe2bf6e0d4729d67a714ef40d95d907b18b/stan/math/rev/core/chainablestack.hpp) as
 
@@ -45,19 +47,23 @@ This feature will be added in the following stages.
 using ChainableStack = AutodiffStackSingleton<vari_base, chainable_alloc>;
 ```
 
-3. Additional PRs for the current arithmetic operators for `var` to accept `vari_value<T>` object types.
+3. Additional PRs for the current arithmetic operators for `var` to accept `var_value<Eigen>` types.
    - Since the new operations work on matrix types, `chain()` methods of the current operators will need to be specialized for matrix, vector, and scalar inputs.
 
 4. Add additional specialized methods for constant matrices.
- - All of our current reverse mode specializations require an `Eigen::Matrix<var>`, which then assumes it needs to generate `N * M` `vari`. Specializations will need to be added for `var_type<Eigen::Matrix<var>>`
+ - All of our current reverse mode specializations require an `Eigen::Matrix<var>`, which then assumes it needs to generate `N * M` `vari`. Specializations will need to be added for `var_type<Eigen::Matrix<T>>`
 
-5. In the stanc3 compiler, support detection and substitution of `matrix/vector/row_vector` types for constant matrix types.
+5. Generalize the current autodiff testing framework to work with constant matrices.
 
-6. Create a design document for allowing type attributes in the stan language.
+6. Add specializations for `operands_and_partials`, `scalar_seq_view`, `value_of` and other helper functions needed for the distributions.
 
-7. Add a `const` type attribute to the stan language.
+7. In the stanc3 compiler, support detection and substitution of `matrix/vector/row_vector` types for constant matrix types.
 
-Steps (1) and (2) have been completed in the example branch with some work done on (3). Step (5-7) have been chosen specifically to allow more time to discuss the stan language implementation. The compiler can perform an optimization step while parsing the stan program to see if a `matrix/vector/row_vector`:
+8. Create a design document for allowing type attributes in the stan language.
+
+9. Add a `const` type attribute to the stan language.
+
+Steps (1) and (2) have been completed in the example branch with some work done on (3). Step (7-9) have been chosen specifically to allow more time to discuss the stan language implementation. The compiler can perform an optimization step while parsing the stan program to see if a `matrix/vector/row_vector`:
 
 1. Does not perform assignment after the first declaration.
 2. Uses methods that have constant matrix implementations in stan math.
@@ -126,9 +132,9 @@ Delaying release of the `const` type to the language allows for a slow buildup o
 # Drawbacks
 [drawbacks]: #drawbacks
 
-More templates can be confusing and will lead to longer compile time. The confusion of templates can be mitigated by adding additional documentation and guides for the Stan Math type system. Larger compile times can't be avoided with this implementation, though other forthcoming proposals can allow us to monitor increases in compilation times.
+More templates can be confusing and will lead to longer compile times. The confusion of templates can be mitigated by adding additional documentation and guides for the Stan Math type system. Larger compile times can't be avoided with this implementation, though other forthcoming proposals can allow us to monitor increases in compilation times.
 
-Waiting for type attributes in the Stan language means this feature will be a compiler optimization until both type attributes are allowed in the language and an agreement is made on stan language naming semantics. This is both a drawback and feature. While delaying a `constant_matrix` type as a user facing feature, it avoids the combinatorial problem that comes with additional type names for matrices proposed in the language such as `sparse` and `complex` matrices.
+Waiting for type attributes in the Stan language means this feature will be a compiler optimization until both type attributes are allowed in the language and an agreement is made on stan language naming semantics. This is both a drawback and feature. While delaying a `constant_matrix` type as a user facing feature, it avoids the combinatorial problem that comes with additional type names for matrices proposed in the language such as `sparse` and `complex`.
 
 # Prior art
 [prior-art]: #prior-art
@@ -138,6 +144,8 @@ Discussions:
  - [Static/immutable matrices w. comprehensions](https://discourse.mc-stan.org/t/static-immutable-matrices-w-comprehensions/12641)
  - [Stan SIMD and Performance](https://discourse.mc-stan.org/t/stan-simd-performance/10488/11)
  - [A New Continuation Based Autodiff By Refactoring](https://discourse.mc-stan.org/t/a-new-continuation-based-autodiff-by-refactoring/5037/2)
+
+[JAX](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-In-Place-Updates) is an autodiff library developed by google whose array type has similar constraints to the constant matrix type proposed here.
 
 [Enoki](https://github.com/mitsuba-renderer/enoki) is a very nice C++17 library for automatic differentiation which under the hood can transform their autodiff type from an arrays of structs to structs of arrays. It's pretty neat! Though implementing something like their methods would require some very large changes to the way we handle automatic differentiation.
 
--- a/designs/0005-static-matrices.md
+++ b/designs/0005-static-matrices.md
@@ -8,11 +8,12 @@
 
 This proposes two things:
 
-1. A generalization of the C++ reverse mode autodiff type in Stan Math, from `stan::math::var`
-to `stan::math::var_value<T>`, where `T` is the underlying storage type of the autodiff
-variable. For `T == double`, this type is equivalent to the current Stan Math `var` type.
-The usefulness of the generalized reverse mode autodiff type comes from passing more
-general data structures as the template type, in particular the Eigen Matrix types.
+1. Adding a template argument to Stan's C++ reverse mode autodiff type
+to allow autodiff variables to have different types of internal storage, for instance
+a `double` or an `Eigen::MatrixXd` or something else. Using Eigen matrices as the
+internal storage would allow matrices in the Stan language to have a much nicer memory
+layout which then also generates more performant programs. This scheme can be fully
+backwards compatible with current Stan code.
 
 2. A design for exposing these autodiff variables at the Stan language level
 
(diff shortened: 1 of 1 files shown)


bbbales2:
I realized we called the test function for varmat `expect_ad_matvar`. Seems like it should be `expect_ad_varmat` lol.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+
+# Summary
+[summary]: #summary
+
+This proposes a generalization of the reverse mode autodiff type in Stan Math,
+`var_value<T>`, where `T` is the underlying storage type of the autodiff variable.
+For `T == double`, this type is equivalent to the current Stan Math `var` type.
+
+The usefulness of this type comes from setting `T` to more general data structures,
+in particular the Eigen Matrix types. Instead of dealing with matrices of vars,
```
syclik, line 14: Suggestion:

1. Move the "The usefulness..." sentence to the paragraph above and add a paragraph break after it. It'd be nice to have the usefulness of the proposed generalization right along with the introductory paragraph.

2. This allows the second paragraph to focus on the point about matrix of vars vs var_value of matrix. Suggested replacement to make it clearer (please feel free to ditch, change, etc.): "Instead of dealing with matrices of `var` (`Eigen::Matrix<var>`), this allows operations in the Math library to be written in terms of a generalized var of matrices (`var_value<Eigen::Matrix<double>>`)." And then continue with the next sentence.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+this allows operations in the Math library to be written in terms of vars that are
+themselves matrices.
+
+This makes working with matrix variables in Stan autodiff much faster. This design
+is fully backwards compatible with the curren Stan autodiff types and those remain.
+
+For the rest of this proposal, "autodiff" means "reverse mode autodiff".
```
syclik, line 21: Two things:
1. Typo: the period goes within the quotation mark... so the correction is: `"reverse mode autodiff."`
2. It'd be good to add that the rest of the proposal, we'll leave off the `stan::math` namespace. Suggested sentence: "For brevity, we will omit the `stan::math` namespace where it's clear from the context." (or something like that?)

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+
+With optimizations from the compiler the conditions for a `matrix` to be a
```
syclik, line 23: @bbbales2, some clarity here? I'm not sure what this paragraph means. Also, why is `matrix` written as code here, but not in other uses?

syclik (commented):
@bbbales2, just adding some comments to the first part of the doc. Feel free to take some of it or not.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Currently, an `N x M` matrix in Stan is represented as an Eigen matrix holding
+a pointer to an underlying array of autodiff objects, each of those holding
+a pointer to the underlying autodiff object implementation.
+
+The basic Eigen type used by Stan is `Eigen::Matrix<var, R, C>`. For brevity,
+this will be referred to as a "matvar" type (a matrix of vars). There are no
```
syclik, line 35: I think it'd help for readability to just use code styling for both `matvar` and `varmat`. It's done below in a couple spots and for consistency and being able to spot it in text, I think it might be useful.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+major differences between a matrix of vars or a vector of vars or a row vector
+of vars.
+
+This proposal will enable an `N x M` Stan matrix to be defined in terms of one
+autodiff object with two separate `N x M` containers, one for its values and
+one for its adjoints.
+
+This basic example of this type is `var_value<Eigen::MatrixXd>`. For brevity,
+these will be referred to as "varmat" types (a var that is a matrix). Again,
```
t4c1, line 44:
```suggestion
these will be referred to as "varmat" types (a var that contains a matrix). Again,
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+there are no major differences when the matrix is replaced by a vector or a row vector.
+
+The motivation for introducing the varmat type is speed. These types speed up
+computations in a number of ways:
+
+1. The values in a varmat are stored separately from the `adjoints` and can
+be accessed independently. In a matvar the values and adjoints are interleaved.
+With how memory loads work on modern cpus, that means reading the values requires
```
t4c1, line 52:
```suggestion
With how cacning and memory loads work on modern CPUs, that means reading the values requires
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+reading the values and the adjoints.
```
t4c1, line 53:
```suggestion
reading the pointers, the values and the adjoints.
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+
+2. To read all `N x M` values or adjoints of a varmat requires one pointer
+dereference and `N x M` double reads. Reading a similar matvar requires
+`N x M` pointer dereferences and `N x M` reads.
```
t4c1, line 57:
```suggestion
dereference and `N x M` double reads. Reading a similar matvar requires
2 pointer dereferences and `N x M` reads.
```
Just pointer to vari and in vari the pointer to the array in arena memory.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+
+3. Memory is linear in varmat, and in the best case interleaved in matvar
+(though because every element is a pointer, in worse case matvar memory accesses
+are entirely random).
+
+4. Because values and adjoints of a varmat are each stored in one place, only
+two arena allocations are required to build an `N x M` matrix. matvar requires
+`N x M` allocations.
```
t4c1, line 65: That is true, but it does not matter that much.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+With how memory loads work on modern cpus, that means reading the values requires
+reading the values and the adjoints.
+
+2. To read all `N x M` values or adjoints of a varmat requires one pointer
+dereference and `N x M` double reads. Reading a similar matvar requires
+`N x M` pointer dereferences and `N x M` reads.
+
+3. Memory is linear in varmat, and in the best case interleaved in matvar
+(though because every element is a pointer, in worse case matvar memory accesses
+are entirely random).
```
t4c1, line 61: Another advantage of linear memory layout is that vectorized memory access instructions can be used to access it. This is a precondition for Eigen doing any vectorization.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+2. It is easier to assign individual elements of a matvar to another variable -- this
+can be done by writing a new pointer over an old pointer. Assigning elements to a varmat
+can be done, but it requires either matrix copies (slow, and not implemented), or
+adding extra functions on the reverse pass callback stack (implemented and discussed later)
+
+# Math Implementation
+[math-implementation]: #math-implementation
+
+At this point versions of `var_value<T>` types have been implemented for `T` being an
+Eigen dense type, an Eigen sparse type, an Eigen expression, a double, or a GPU type.
```
t4c1, line 91:
```suggestion
Eigen dense type, an Eigen sparse type, an Eigen expression (a limited number of expression types are supported at the moment), a double, or a `matrix_cl` (matrix with data on OpenCL device - usually a GPU).
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+varmat are saved into the memory arena before being overwritten, and a call is pushed
+onto the chaining stack that will restore the values back from the arena and set the
+respective adjoints of the varmat entries to zero.
+
+The assumptions that makes this work is after a subset of a varmat is overwritten,
+that part of the varmat cannot be used in any further calculations, and so the adjoints
+of that part of the matrix will be zero in the reverse pass at the time of assignment.
+
+This means that assignment into a varmat would invalidate any views into that
+varmat's memory.
```
t4c1, line 165: I would not say invalidates. Just the change is visible in views. Also I don't think you explained anywhere what views are.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+The initial varmat implementation of this model achieves a per-gradient speedup of around two.
+
+Now that varmat variables can be assigned, it is possible that the compiler silently replace every
+matvar in Stan with a varmat, and none of the standard Stan use cases would be affected.
+
+There are four major implementation strategies available at the Stanc3 level.
+
+1. Implement a new Stan type to differentiate explicitly between matvar and varmat types
+
+2. Use matvar types whenever there is any function in the Stan program that does not support varmat
```
t4c1, line 192: Once all the functions support varmat this is the same as 4.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,240 @@
...
+The advantage of #2 is that it would allow programs that can be written using only varmat functions
+to be very fast (and not even worry about any conversion to matvar or not). The problem with this
+implementation is that, the difference between `varmat` and `matvar` would again need documented
+at the user level and some notation would need added to the user manual to indicate if a function
+is or is not available for optimization.
+
+The advantage of #3 is that varmat and matvar types could be implemented without any extra
+documentation for the user. Similarly to #2, if the user uses functions only available for matvar,
+there will be some slowdown as conversions between varmat and matvar are handled, but unlike #2
+the slowdown might not be catastrophic. In this case it would not be as crucial to carefully
```
t4c1, line 217: Which of 2/3 is slower is just speculation. Also I bet it will depend on the model a lot. I would prefer this removed unless you support it by benchmarking multiple models.

designs/0005-static-matrices.md
bbbales2: I think that was a danging sentence from the editing process. Removed.

designs/0005-static-matrices.md
bbbales2: Yeah, I will do this.

I originally wanted to avoid `varmat` because `varmat` is not real code and `var` s are and I don't highlight them.

designs/0005-static-matrices.md
bbbales2: I want to leave the pointer point to point number 2

designs/0005-static-matrices.md
bbbales2: This one is matvar, not varmat, so each vari we access requires a pointer dereference and we can't assume varis are stored near each other or anything.

designs/0005-static-matrices.md
bbbales2: True, but I'm also not sure I should delete it. Can though if you think it's distracting. It's certainly not as important as the others.

designs/0005-static-matrices.md
bbbales2: I still like invalidates, but it's not a segfault or anything to read the view. Perhaps a segfault would be better though since it would prevent inadvertent bugs, so I'm not sure. I changed it.

designs/0005-static-matrices.md
bbbales2: I don't think we'll ever reach the 100% conversion point.

designs/0005-static-matrices.md
bbbales2: Hmm, it is, but also I don't want to say 2 and 3 are the same, cause they're not. The reason to do 3 would be to avoid at least one of:

1. catastrophic slowdowns in 2 where using a weird function with no varmat implementation in one place means the whole program is slow

2. documenting all functions that do and do not support varmat and to what extent (only certain arguments, etc.)

bbbales2: In the end I added the assumption that the matvar parts are cheap. So like, if someone needs one `lbeta` call and that isn't a `varmat` function, then all the matrix operations and stuff don't get converted back to `matvar` -- only the single, assumed relatively cheap function gets wrapped up.

designs/0005-static-matrices.md
bbbales2: I added a note that 2, 3, and 4 all become the same thing if varmat is supported everywhere.

bbbales2:
@syclik @t4c1 thanks for comments, I made the edits (or comments)

designs/0005-static-matrices.md
t4c1: I think pointers are relevant here as well - they are extra data that needs to be transfered.

designs/0005-static-matrices.md
t4c1: if we are counting memory transfers than it is `2 x N x M`, because values are transferred into the cache as well.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+[here](https://github.com/bbbales2/computations_in_glms/blob/master/benchmark_model/mrp_`varmat`.stan).
+
+The initial `varmat` implementation of this model achieves a per-gradient speedup of around two.
+
+Now that `varmat` variables can be assigned, it is possible that the compiler silently replace every
+`matvar` in Stan with a `varmat`, and none of the standard Stan use cases would be affected.
+
+There are four major implementation strategies available at the Stanc3 level.
+
+1. Implement a new Stan type to differentiate explicitly between `matvar` and `varmat` types
```
rok-cesnovar, line 200:
```suggestion
1. Implement a new Stan type to let the user explicitly determine whether to use `matvar` and `varmat` types
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+4. Use `varmat` types everywhere
+
+The original proposal was leaning towards #1 because it most naturally corresponds to the Math
+implementation. However, this leaves the problem of explicitly declaring `varmat` to the user. To
+avoid duplicating the user's manual and function reference for `varmat` and `matvar` functions, there
+would also need to be some automatic, under the hood conversions between the types. This automatic
+converting would make it so that if a `varmat` is forgotten anywhere, the entire program will be
+slowed down. This sort of explicit typing shouldn't be too much of a problem in development models
+or generated Stan code, but in large, user-written models this would be error prone to implement
+and maintain.
```
rok-cesnovar, line 216: 1 also duplicates the list of signatures (longer error messages). Though that can be assumed from the duplicating function reference I guess.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+The initial `varmat` implementation of this model achieves a per-gradient speedup of around two.
+
+Now that `varmat` variables can be assigned, it is possible that the compiler silently replace every
+`matvar` in Stan with a `varmat`, and none of the standard Stan use cases would be affected.
+
+There are four major implementation strategies available at the Stanc3 level.
+
+1. Implement a new Stan type to differentiate explicitly between `matvar` and `varmat` types
+
+2. Use `matvar` types for any function in the Stan program that does not support `varmat`
```
rok-cesnovar, line 202:
```suggestion
2. Use `matvar` if there is at least one function in the Stan program that does not support `varmat`
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+parts of the calculation can still try to take advantage of `varmat` types. The difficulty here
+is compiler implementation which would need to be more advanced to detect when it could use
+`varmat` and when it would need to convert to `matvar`.
+
+The advantage of #4 is simplicitly of use and simplicitly of compiler implementation. If `varmat`
+types are used everywhere, and there is no attempt to balance against the advantages of
+`varmat` against `matvar`, then the main difficulty will just be making sure to convert `varmat`
+variables to `matvar` variables and back again when used in functions that do not support `varmat`.
+In the limit that all functions can be written to support `varmat`, then #2, #3, and #4 become
+equivalent.
```
rok-cesnovar, line 237: 4 was kind of meant more in the sense switch to `varmat` when we get everything in Math and scratch `matvar`.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+Discussions:
+ - [Efficient `static_matrix` type?](https://discourse.mc-stan.org/t/efficient-static-matrix-type/2136)
+ - [Static/immutable matrices w. comprehensions](https://discourse.mc-stan.org/t/static-immutable-matrices-w-comprehensions/12641)
+ - [Stan SIMD and Performance](https://discourse.mc-stan.org/t/stan-simd-performance/10488/11)
+ - [A New Continuation Based Autodiff By Refactoring](https://discourse.mc-stan.org/t/a-new-continuation-based-autodiff-by-refactoring/5037/2)
+
+[JAX](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-In-Place-Updates) is an autodiff library developed by Google whose array type has similar constraints to the constant matrix type proposed here.
```
rok-cesnovar, line 248: There are no constraints with vatmar now.

rok-cesnovar:
Also, maybe we should not list option 2 here as 1, 3 and 4 are end-game options: 
- new type
- smart use of varmat (avoid when heavy single-cell indexing is used)
- varmat only

2 is just a midpoint because we can't realistically expect to implement everything for 2.26. If everything in Math has varmat 2 becomes either 3 or 4.

bbbales2:
I'll add that 2 is kindof a stepping stone, but if it is something we're considering for release we should have it in the design doc.

designs/0005-static-matrices.md
bbbales2: Good point, I forgot the 2

designs/0005-static-matrices.md
bbbales2: I added a bit about the pointers

designs/0005-static-matrices.md
bbbales2: This is simpler than #3 though right? This is like #3, except we don't try to be smart about making stuff a `varmat` or a `matvar`. We just, as much as possible, make things `varmat` types.

If this doesn't make sense to you as a proposal we should change it though lol, cuz this was my interpretation of a list you made :P.

rok-cesnovar: No, I meant it as "wait till everything it varmat" and then just switch to it. It's not a good option, I just wanted to present all cases.

bbbales2: Okay, I'll remove this one then. If it's not realistic and we don't want to do it, we shouldn't worry about listing it.

We haven't truly listed all options, so let's not worry about that (for instance, we could also stick with `_varmat`).

rok-cesnovar: Ok yeah, good point.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+a pointer to the underlying autodiff object implementation.
+
+The basic Eigen type used by Stan is `Eigen::Matrix<var, R, C>`. For brevity,
+this will be referred to as a `matvar` type (a matrix containing var variables).
+There are no major differences between a matrix of vars or a vector of vars or
+a row vector of vars.
+
+This proposal will enable an `N x M` Stan matrix to be defined in terms of one
+autodiff object with two separate `N x M` containers, one for its values and
+one for its adjoints.
```
SteveBronder, line 42: I think this would clean up a bit with some pseudocode examples. I think you can take (1) and (2) that you have written below and include them in the example for matrix multiplication

-------------------------

Currently, Stan represents matrices as an `Eigen::Matrix<var, Eigen::Dynamic, Eigen::Dynamic>`. There are no major differences between a matrix of vars or a vector of vars or a row vector of vars. The internal representation of an Eigen matrix is a pointer to an array.

```cpp
// pseudo class representing the storage of a dense matrix
template <typename T>
class DenseStorage {
  T* data_;  
};
```

Stan math's reverse mode autodiff (`var`) is a smart pointer type that itself points to an underlying object (`vari`) that actually holds the values and adjoints that Stan math uses when calculating gradients.

```cpp
// pseudo class holding the values and adjoints used in reverse mode AD
class vari {
  double val_;
  double adj_;
};

// pseudoclass that represents that stores the pointer to the underlying vari
class var {
  vari* vi_;
}
```

This means that for Eigen matrices of type `Eigen::Matrix<var, Eigen::Dynamic, Eigen::Dynamic>` it's internal representation is a pointer to an array of pointers.

```
// pseudo class that shows equivalently what the compiler sees
class DenseStorage {
  vari** data_;  
}
```

This becomes problematic with both pointer chasing and memory access. Typically when Stan math performs operations over matrices it will want all of the values of a `vari` for one part of an operation and the all the adjoints of another `vari` in another part of an operation. This can be wasteful as to fetch just the values or adjoints we need to go through the pointer of pointers and load both the values and adjoints into cache. And unless the value and adjoints are used together in the same operation then the CPU is also seeing non-contiguous data.

This proposal will enable an `N x M` Stan matrix to be defined in terms of one
autodiff object with two separate `N x M` containers, one for its values and
one for its adjoints. This is represented as a `var_value<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>`.

```cpp
// pseudo class for matrix vari
template <typename T>
class vari_value {
  Eigen::Map<T>* val_;
  Eigen::Map<T>* adj_;
}
// pseudo code representing var matrix
template <typename T>
class var_value {
  vari_value<T>* vi_;
};
```

This storage scheme allows for less pointer chasing while providing contiguous memory access to values and adjoints. For instance, when performing the reverse mode pass for matrix autodiff the calculation for propagating the adjoints of each input requires accessing the values of the other input

```cpp
A.adj() += result.adj() * B.val().transpose();
B.adj() += A.val().transpose() * result.adj();
```

With an Eigen matrix of var types this means that each multiplication in the above requires the program to fetch both the values and adjoints from the pointer to the array of pointers for both `A` and `B` when calculating either adjoint. Furthermore this makes those values non-continguous in the operation and so the compiler will not be able to use advanced instruction sets such as SIMD multiplication.

However, switching to a `var_value` with the values and adjoints separated into two arrays means that both adjoint calculations will only be fetching the values or adjoints they need in the operation. And those values and adjoints will be contiguous so that the compiler has a much better opportunity to utilize advanced instruction sets when performing the calculation.

designs/0005-static-matrices.md
SteveBronder: I think it's fine to leave. It is a performance cost that I think will show up in large sizes

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+5. Because values and adjoints of a `varmat` are each stored in one place, only
+two arena allocations are required to build an `N x M` matrix. `matvar` requires
+`N x M` allocations.
+
+See [here](https://github.com/stan-dev/design-docs/pull/21#issuecomment-625352581)
+for performance tests on a small subset of gradient evaluations using `varmat` and
+`matvar` types.
+
+`matvar` has two advantages that `varmat` does not duplicate. These are listed
+here to motivate the backwards compatibility of this proposal with `matvar`:
```
SteveBronder, line 82: I don't really like using the shorthand `varmat` and `matvar` here. I think we should just say "An Eigen matrix with a scalar var type" and "a var type that uses Eigen matrices as storage."

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+can be done by writing a new pointer over an old pointer. Assigning elements to a `varmat`
+can be done, but it requires either matrix copies (slow, and not implemented), or
+adding extra functions on the reverse pass callback stack (implemented and discussed later)
+
+# Math Implementation
+[math-implementation]: #math-implementation
+
+At this point versions of `var_value<T>` types have been implemented for `T` being an
+Eigen dense type, an Eigen sparse type, a limited number of Eigen expressions, a double,
+or a `matrix_cl`, a matrix with data stored on an OpenCL device (targetting GPUs).
```
SteveBronder, line 98:
```suggestion
or a `matrix_cl` which is a matrix with data stored on an OpenCL device (targeting GPUs).
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+
+As var previously, `var_value<T>` is a PIMPL (pointer to implementation) type. The
+implementation of `var_value<T>` is held in `vari_value<T>`, and the only data a
+`var_value<T>` holders is a pointer to a `vari_value<T>` object.
```
SteveBronder, line 102:
```suggestion
As before, `var_value<T>` is a PIMPL (pointer to implementation) type. The
implementation of `var_value<T>` is held in `vari_value<T>`, and the only data a
`var_value<T>` holds is a pointer to a `vari_value<T>` object.
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+## Assigning a `varmat`
+
+A limitation from the original design doc that is no longer necessary is the limitation
+that subsets of `varmat` variables cannot be assigned efficiently.
+
+The efficiency loss was so great in the original design (requiring a copy of the entire
+`varmat`) that this idea was discarded completely.
+
+There is a strategy available now to assign to subsets of `varmat` variables that is
+more efficient than this.
```
SteveBronder, line 165:
```suggestion
A limitation from the original design doc that is no longer necessary is the limitation that subsets of `varmat` variables cannot be assigned efficiently. The efficiency loss was so great in the original design (requiring a copy of the entire `varmat`) that this idea was discarded completely. There is a strategy available now to assign to subsets of `varmat` variables that is more efficient than this.
```

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,252 @@
...
+at the user level and some notation would need added to the user manual to indicate if a function
+is or is not available for optimization.
+
+The advantage of #3, similarly to #2, is that `varmat` and `matvar` types could be handled
+automatically by the compiler. If the user requires functions only available for `matvar`,
+there will be some slowdown as conversions between `varmat` and `matvar` are handled. However,
+if the conversions are only used in relatively cheap parts of the computation, then the expensive
+parts of the calculation can still try to take advantage of `varmat` types. The difficulty here
+is compiler implementation which would need to be more advanced to detect when it could use
+`varmat` and when it would need to convert to `matvar`.
```
SteveBronder, line 234: Is it possible to, as the compiler is building the program, trace through where an object is used? If so then an alt here is that when the compiler is parsing the AST, if it sees that a matrix is only going through functions supported by varmat and is not single cell accessed or read from, then the compiler can just swap out the `Eigen::Matrix<var>` with `var_value<Eigen::Matrix>`

pinging @rybern, is this a feature we have in the optimization stuff for the compiler?

SteveBronder (changes requested):

designs/0005-static-matrices.md
rok-cesnovar: Yes, that is possible, we do that for OpenCL. This is part of transform MIR.

rok-cesnovar: But before we go that direction we need to see how much single cell access affects performance and compare that with the cost of a varmat-> matvar.

SteveBronder: > Yes, that is possible, we do that for OpenCL. This is part of transform MIR.

Just a sanity check, do we also check the path of the result as well? Since `var<matrix>` functions return var matrices we need to traverse the whole path for the object

> But before we go that direction we need to see how much single cell access affects performance and compare that with the cost of a varmat-> matvar.

@t4c1 would you have any time to make benchmarks for this? I might be able to later in the week. My guess is that it's a big difference

rok-cesnovar: We need to do multiple passes then if we want to traverse.

designs/0005-static-matrices.md
bbbales2: Thanks, I agree the C++ is useful for showing what the types are. I reordered some stuff and dropped the DenseStorage things.

designs/0005-static-matrices.md
bbbales2: Ooof, that was hard to edit in, so I kept the shorthand. I think this notation has been handy during development, so I like it for the design document.

`varmat` also includes such things as a view into another `varmat`, or an autodiff type on a GPU, etc. It's sort of imprecise, but it's handy that way (and `varmat`/`matvar` shorthands now come after a section with C++ examples and talking specifically about the details of the two implementations)

bbbales2:
Edits in. Hopefully stuff looks okay? At this point the thing that bothers me are the old `varmat` benchmarks. We should have something newer.

@rok-cesnovar mentioned maybe we should remove the version of the compiler with explicit types from this. That was the original design, but I agree now it looks unnecessary and difficult. If it's not an option we're considering, let's remove it. If nobody objects I'll remove it tomorrow sometime.

Does tomorrow afternoon work to reopen the design doc pull with this? I think we should close the old one and open a new one just to get a refresh on the conversation. That last one was loooooong (https://github.com/stan-dev/design-docs/pull/21). Can ofc. link back to it.

designs/0005-static-matrices.md
SteveBronder: ^yeah that's what I mean, we need to have the ability to do multiple sweeps. ftr I have no idea if that's possible or reasonable in the compiler

designs/0005-static-matrices.md
SteveBronder: We only use varmat / matvar because we type it a 100 times a day :-P but for defining a tech spec idt we should use shorthand verbage like that

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,343 @@
...
+autodiff object with two separate `N x M` containers, one for its values and
+one for its adjoints. Filling in the templates, the storage of a
+`var_value<Eigen::MatrixXd>` is effectively:
+
+```cpp
+template <>
+class vari_value<Eigen::MatrixXd> {
+  Eigen::MatrixXd value_;
+  Eigen::MatrixXd adjoint_;
+}
```
SteveBronder, line 94: I'd put this back to the way I wrote it because on a first glance this code is going to leave dangling memory in `Eigen::MatrixXd`. I would go the opposite and just explain that we can use allocate the memory in our stack allocator but use `Eigen::Map<>` to act as if these are normal Eigen matrices

designs/0005-static-matrices.md
SteveBronder: Why did you drop the densestorage? I think it's one of the bigger points that the vari** impl causes a bunch of pointer chasing. I prefer just having that laid out simply because idt it's obvious at first

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,343 @@
...
+
+template <>
+class var_value<Eigen::MatrixXd> {
+  vari_value<Eigen::MatrixXd>* vi_; // pointer to implementation
+};
+```
+
+The full, contiguous matrices of values or adjoints can be accessed with only
+one pointer dereference, eliminating the two prime inefficiences of the
+`Eigen::Matrix<var, Eigen::Dynamic, Eigen::Dynamic>` implementation.
```
SteveBronder, line 104: It's still two pointer dereferences in the implementation. The main benefit is that we only have to  dereference twice will for the vari** we have to double a double dereference on each lookup

designs/0005-static-matrices.md
bbbales2: Was re-reading the last doc comments, a very early reference to something like varmat and matvar was Bob: https://github.com/stan-dev/design-docs/pull/21#discussion_r418700776 . So maybe it makes sense to people outside too? Anyway, this one I want to roll with until it's a problem.

designs/0005-static-matrices.md
bbbales2: I'm gonna add a note in t he comments. It won't actually leak memory cause it's not being allocated in the `arena` according to this code either :D.

The difficulty of C++ psuedo-code is it's basically guaranteed to be wrong (unless it's a copy of the implementation). Vague pseudo-code is nicer cause you have wiggle room.

But ofc. this discussion came up the first round too: https://github.com/stan-dev/design-docs/pull/21#discussion_r421806784

And I see Bob made comments that it was sort of inappropriate to just link C++ implementations like I've done here.

bbbales2:
A few interesting comments from the last design:

On assigning to varmat: https://github.com/stan-dev/design-docs/pull/21#discussion_r421833109
Something about a LHS -- seemed interesting but not sure what it was: https://github.com/stan-dev/design-docs/pull/21#issuecomment-623697543
How to construct varmats directly: https://github.com/stan-dev/design-docs/pull/21#issuecomment-625537542
Betancourt weighing in on elementwise modification: https://github.com/stan-dev/design-docs/pull/21#issuecomment-626918408

designs/0005-static-matrices.md
bbbales2: Fair, I'll add something back.

bbbales2:
From reading the previous design doc comments, what is missing here is a discussion of what member functions are required in the different `var_value<T>` types. The previous doc says linking to C++ implementation is not the right way to do this.

bbbales2:
Also, if I remove explicit types as an option here, this leaves two options at the compiler level:

```
1. Use `matvar` types if there is at least one function in the Stan program that does not support `varmat`

2. Use `varmat` types by default, but if variable assignment is detected, or a `varmat` is used
in an unsupported function convert to and from `matvar` automatically.
```

Is the first thing here something we're thinking about seriously? I don't like the first that much. The 2nd seems better and more like what is being discussed: https://github.com/stan-dev/design-docs/pull/29#discussion_r542767273

And there are the scalar access benchmarks. I'll try to do some of those today so we have them. Maybe with that we actually do have a pretty complete idea of what the language implementation will look like and can finish that part of the doc.

bbbales2:
Code below, double checks are appreciated.

scalar_timings

Here's the code: https://gist.github.com/bbbales2/347ea12273ed90b1df35d55cf247989e

A slowdown of 1 (the y-axis) would mean that the two things run at the same speed.

SteveBronder:
So reading this it looks like for something like

```stan
vector[N] A = //fill in;
vector[N] B;
for (i in 1:N) {
  B[i] = some_func(A[i]);
}
```

would be about 4x as slow?

bbbales2:
scalar_timings

I thought it would have been about the average of the two, but it's much much expensive!

Code [here](https://gist.github.com/bbbales2/16707dd614ac64fec876acb1d3a6a855)

bbbales2:
Just saving this text here:

```
For implementation of `vari_value<T>` where `T` is an Eigen dense type (the basic `varmat`),                                                                  
look [here](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/vari.hpp#L604)                                                                   
                                                                                                                                                              
For implementation of `vari_value<T>` where `T` is an Eigen expression, look                                                                                  
[here](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/vari.hpp#L546). The                                                                   
Eigen expressions implemented are ones that can be implemented as views into another matrix.                                                                  
So blocks of matrices, matrix transpose, column-wise access, etc. are supported.                                                                              
                                                                                                                                                              
For implementation of `vari_value<T>` where `T = double` (the reimplementation of var),                                                                       
look [here](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/vari.hpp#L81)                                                                    
                                                                                                                                                              
For implementation of `vari_value<T>` where `T` is an OpenCL type, look                                                                                       
[here](https://github.com/stan-dev/math/blob/develop/stan/math/opencl/rev/vari.hpp#L23)                                                                       
                                                                                                                                                              
For implmementation of `vari_value<T>` where `T` is an Eigen sparse type, look                                                                                
[here](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/vari.hpp#L746)
```

bbbales2:
Can we do a Hangouts tomorrow to write a 0.1 compiler spec? Just to get a reasonable baseline written out? Like 9 or 10AM 
 or 11AM EST?

I tried to write it myself and wasn't really getting very [far](https://github.com/stan-dev/design-docs/pull/29/files#diff-4266cf6088836511d41a540def204de9ea210a6dc117b7b6a2575becae6a659cR369).

I think I could share screen and ask questions and get something written down pretty quick, but I'm kinda clueless on compiler stuff by myself.

SteveBronder:
Yeah sure let's do like 11

SteveBronder:
Lest @rok-cesnovar and @t4c1 want to join and meet at an earlier time

bbbales2:
@SteveBronder Rok said he had car repair to do, so we'll flex the time a bit to what works for him. Will message on slack when it's go-time.

bbbales2:
@rok-cesnovar @SteveBronder @t4c1 alright, design doc up with a full compiler implementation we discussed this AM and the dependencies on the C++ implementation removed. Have another skim through (and then we can make this an official design-doc pull and get comments, etc. etc.).

SteveBronder:
I like it! Do we have a few other folks look over this then it goes to an accept / reject vote?

bbbales2:
Next step is making it the varmat design doc and getting comments from everyone. Rok was happy with this version, so since you're happy with it too I'll go ahead and make it the varmat doc and go about summoning comments.

PR #30, opened 2020-12-22
Matrix autodiff type

bbbales2:
This is a new version of the design doc for a matrix autodiff type in Stan.

[Rendered version](https://github.com/stan-dev/design-docs/blob/static-matrix-dec-update/designs/0005-static-matrices.md)

(This is a follow up to the original pull [here](https://github.com/stan-dev/design-docs/pull/21))

### DIFF
--- a/designs/0005-static-matrices.md
+++ b/designs/0005-static-matrices.md
@@ -18,7 +18,7 @@ Currently, a differentiable matrix in Stan is represented as an Eigen matrix hol
 
 At the Stan Math level a constant matrix is a `var_value<Eigen::Matrix<double, -1, -1>>` with an underlying pointer to a `vari_value<Eigen::Matrix<double, -1, -1>>`\*.  This means that accessors for the value and adjoint of `var_value` objects can access contiguous chunks of each part of the `vari_value`. Any function that accepts a `var_value<T>` will support static matrices.
 
-At the language and compiler level, a `matrix` can be substituted for a constant matrix if the matrix is only constructed once and never reassigned to.
+At the language and compiler level, a `matrix` can be substituted for a constant matrix if the matrix is only constructed once and it's subslices are never assigned to.
 
 ```stan
 const matrix[N, M] A = // Construct matrix A...
@@ -26,10 +26,9 @@ A[10, 10] = 5; // illegal!
 A[1:10, ] = 5; // illegal!
 A[, 1:10] = 5; // illegal!
 A[1:10, 1:10] = 5; // illegal!
-A = X; // illegal!
 ```
 
-Any function or operation in the stan language that can accepts a `matrix` as an argument can also accept a constant matrix.
+Any function or operation in the stan language that can accepts a `matrix` as an argument can also accept a constant matrix. Functions which can take in multiple matrices will only return a constant matrix if all of the other matrix inputs as well as the return type are static matrices.
 
 
 # Reference-level explanation
@@ -59,11 +58,7 @@ using ChainableStack = AutodiffStackSingleton<vari_base, chainable_alloc>;
 
 7. In the stanc3 compiler, support detection and substitution of `matrix/vector/row_vector` types for constant matrix types.
 
-8. Create a design document for allowing type attributes in the stan language.
-
-9. Add a `const` type attribute to the stan language.
-
-Steps (1) and (2) have been completed in the example branch with some work done on (3). Step (7-9) have been chosen specifically to allow more time to discuss the stan language implementation. The compiler can perform an optimization step while parsing the stan program to see if a `matrix/vector/row_vector`:
+Steps (1) and (2) have been completed in the example branch with some work done on (3). Step 7 has been chosen specifically to allow more time to discuss the stan language implementation. The compiler can perform an optimization step while parsing the stan program to see if a `matrix/vector/row_vector`:
 
 1. Does not perform assignment after the first declaration.
 2. Uses methods that have constant matrix implementations in stan math.
@@ -148,20 +143,5 @@ Discussions:
 [JAX](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#%F0%9F%94%AA-In-Place-Updates) is an autodiff library developed by google whose array type has similar constraints to the constant matrix type proposed here.
 
 [Enoki](https://github.com/mitsuba-renderer/enoki) is a very nice C++17 library for automatic differentiation which under the hood can transform their autodiff type from an arrays of structs to structs of arrays. It's pretty neat! Though implementing something like their methods would require some very large changes to the way we handle automatic differentiation.
-
-# Unresolved questions
-[unresolved-questions]: #unresolved-questions
-
-- What parts of the design do you expect to resolve through the RFC process before this gets merged?
-
-Whether the staged development process listed in the reference level explanation will suffice.
-
-If the restriction on not allowing assignment to the entire matrix such as `A = X` is too restrictive.
-
-- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
-
-Any intricacies of using the GPU via `var_value<matrix_cl<double>>` should be deferred to a separate discussions or can happen during the implementation.
-
-Methods involving changing the current `matrix` type in the Stan language
 
 *Interestingly, this also means that `var_value<float>` and `var_value<long double>` can be legal.
--- a/designs/0005-static-matrices.md
+++ b/designs/0005-static-matrices.md
@@ -1,6 +1,6 @@
 - Feature Name: static_matrices
 - Start Date: April 30th, 2020
-- RFC PR: [Example Branch](https://github.com/stan-dev/math/compare/feature/var-template)
+- RFC PR: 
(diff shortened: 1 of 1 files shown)


bbbales2:
A few extra things about this design doc.

1. Most of what is written here for Math is already in Math. That might seem to defeat the meaning of doing a design doc, but our excuses are:
    a. We did not introduce changes in the Stan language
    b. We did not remove the old stuff (the interface here is backwards compatible)
    c. We're still happy to discuss the stuff we already put in Stan Math

2. Though we haven't merged anything into Stanc3 yet, there is a prototype compiler, and we can get about a 2x speedup on a simple hierarchical model using the new autodiff types here ([here](https://github.com/bbbales2/computations_in_glms/blob/master/benchmark_model/mrp_varmat.stan))

3. The old design doc is [here](https://github.com/stan-dev/design-docs/pull/21). No need to read it -- just a link for convenience. Most of the conversation is still really relevant.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,444 @@
...
+internally defined variables and returned variables are `matvar` types as well.
+
+2. Assume every named and unnamed matrix variable in the `parameters`,
+`transformed parameters` and `model` blocks use `varmat` types
+
+3. If a matrix variable is used in a function that does not support `varmat`
+arguments, it is made a `matvar` variable.
+
+4. If single elements of a matrix variable are indexed, it is made a `matvar`
+variable.
```
nhuurre, line 373: Based on the example hierarchical model `varmat` vector supports multi-indexing. So is this _just_ scalar elements; selecting a single row or column does not require `matvar`?

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,444 @@
...
+
+5. If a `matvar` matrix variable is assigned to a second matrix variable,
+the second is made to be a `matvar` variable.
```
nhuurre, line 376: What about assigning a `varmat` to a `matvar`? Isn't that quite expensive too?

nhuurre (commented):
This looks great. I agree it makes sense to have the compiler figure out where to use `matvar` and `varmat` automatically without explicit types at the language level.

I didn't see any discussion of ["varmat comprehensions"](https://discourse.mc-stan.org/t/static-immutable-matrices-w-comprehensions/12641). Was that idea abandoned?

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,444 @@
...
+
+This leaves `b` and `mu`, matrix variables defined in the `parameters` and
+`model` block. Any matrix variable defined in a `transformed parameters` block
+would be decided the same way.
+
+The algorithm for figuring out the underlying matrix types for these variables
+is as follows:
+
+1. Define all user-defined functions only for `matvar` types, and all
+internally defined variables and returned variables are `matvar` types as well.
```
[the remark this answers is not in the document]
rok-cesnovar: Yes.

designs/0005-static-matrices.md
```diff
@@ -0,0 +1,444 @@
...
+
+2. Assume every named and unnamed matrix variable in the `parameters`,
+`transformed parameters` and `model` blocks use `varmat` types
```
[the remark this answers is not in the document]
rok-cesnovar: Yes, currently all Stan Math functions that do not take in an Eigen-type but do return it, return a matvar Eigen-type. I would say once we do support returning varmat in those, they would be separated via a template. Other options like suffixes would cause a ton of issues in the compiler.

designs/0005-static-matrices.md
rok-cesnovar: Scalar elements also work, they are just slow. Read or write access to scalars leads to slowdowns to a factor of 4. See
https://github.com/stan-dev/design-docs/pull/29#issuecomment-746870561

In the grand scheme of things of a typical model its debatable what effect that would have, but we want to be on the safe side.

rok-cesnovar: Good point on row access. I would imagine its faster or close, but a performance test is needed definitely.

designs/0005-static-matrices.md
rok-cesnovar: The way its envisioned now, that can not happen. The variable being assigned to will be labeled matvar and will be matvar for its entire "lifetime". Its a bit conservative.

nhuurre: I'm thinking of something like
```stan
parameters {
  matrix[N,K] x;
  matrix[K,L] y;
}
transformed parameters {
  matrix[N,L] z = x * y;
  print(z[1,1]);
}
```
`z` is `matvar` due to indexing. But what about `x` and `y`? Maybe they're `varmat` and the product is calculated as a `varmat` but then converted to a `matvar`. Or are _both_ `x` and `y` forced to be `matvar`? Or only one?

rok-cesnovar: Current plan is x and y are forced to be matvar. Mostly because it makes the rules simple. We tried to avoid stuff like: "if X has more than 5 single element access then matvar". We realize these cases are a bit unfortunate.

We circled around this one a bit, so if you have any suggestions, we are all ears.

nhuurre: Yeah, I don't have any ideas. I'm just trying to clarify the inference rules in this section.
It says earlier that if a function takes _any_ `varmat` argument then the return type is `varmat`. So conversely, if the return type needs to be `matvar` then all inputs must also be `matvar`.

rok-cesnovar: That is definitely a good point yes. That should be clarified.

designs/0005-static-matrices.md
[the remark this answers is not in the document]
bbbales2: > Not sure what these "unnamed matrix variables" are?

Temporaries, like:

```
matrix[5, 5] A = B * C + D;
```

Then there's probably a temporary variable for `B * C`, and it has no name.

> It is certainly possible for stanc3 to figure out the optimal return type for them.

For functions defined in Math, the list of function names + arguments -> return types will be given. Stanc3 won't have to do anything here.

For user-defined functions, in the current proposal, all return types are matvar.

Once we start deducing types for user-defined functions, then we could hopefully decide the return type using similar rules as we have everywhere else (use varmat unless there is a rule that says not to).

I hope we can avoid having the return type be part of the type of the function. It seems nice that we can take name + argument types and compute a return type. Functions with multiple return types sound annoying.

designs/0005-static-matrices.md
bbbales2: I'll do a benchmark of row/column access. The overhead here comes from adding a callback for every access. Accessing a row of elements is only one callback, so that should be quick still for varmat compared to matvar.

designs/0005-static-matrices.md
bbbales2: Yeah I think we need another rule for this.

bbbales2:
> I didn't see any discussion of "varmat comprehensions". Was that idea abandoned?

I think since it's possible to get mileage out of this without that function, it's been ignored. In GP examples and stuff building the matrices with normal autodiff is expensive though, so the uses are out there for sure.

Looking at the example in the link, we'd need to do some compiler things here to make this work:

```
matrix matrix_comprehend(real(int, int) f, int M, int N) {
  matrix[M, N] y;
  for (n in 1:N)
    for (m in 1:M)
      y[m, n] = f(m, n);
  return y;
}
```

If we want the output to be a varmat type, then we'd need to somehow make sure that the function `f` wasn't acting like a `matvar` function (emitting scalar autodiff variables). However, if `f` is a function from Math, it will do just that (emit `var_value<double>` variables and put callbacks on the autodiff stack that expect those `var_value<double>` things to be used).

With user-defined functions we could probably do something. User-defined functions would probably be the main use case.

This seems harder than I expected now that I think about it.

designs/0005-static-matrices.md
bbbales2: Row/column sums are slower for varmat for size 4 vectors but faster for 8 and up. Columns are easier than rows. Lower on this graph is better for varmat.

row_column_timings

bbbales2: [Here](https://gist.github.com/bbbales2/af6eb405402acc7ac8c7d27a5ea2f56e) is the benchmark code.

rok-cesnovar: Thanks. So I would say this is not a problem and row/column access does not rule out varmat.

designs/0005-static-matrices.md
bbbales2: Added a rule

designs/0005-static-matrices.md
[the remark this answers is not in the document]
bbbales2: I changed the verbage here a bit to make this clearer.

bbbales2:
@nhuurre Thanks for doing a review.

I made a couple edits and added responses. Should be ready to go again

nhuurre:
Looks good to me.

bbbales2:
Cool thanks @nhuurre. Did the repo not let you leave an approving review? We can up your Github permissions if you want em'. 

Otherwise @rok-cesnovar can click the button and collect the delicious Github points.

rok-cesnovar (approved):
Approving for @nhuurre.

PR #31, opened 2021-01-05
Profiling

rok-cesnovar:
The goal of this is design document is to enable measuring how computationally expensive different parts of a model are with
a profiling interface.

The ideas for this design were previously discussed [here](https://discourse.mc-stan.org/t/proposal-for-profiling-stan-models/15322) and [here](https://github.com/stan-dev/cmdstan/issues/960). In the latter there are also links to preliminary implementations of this for Math and stanc3.

Thanks to @bbbales2 for helping me with this design doc!

[Link to rendered version](https://github.com/stan-dev/design-docs/blob/d6f48f4b0424dc3d9ae5b3c3700da681aecad09b/designs/0013-profiling.md)

### DIFF
--- a/designs/0013-profiling.md
+++ b/designs/0013-profiling.md
@@ -303,6 +303,10 @@ well as inner blocks). It is fairly easy to add profile statements where they
 are needed manually, so this seemed unnecessary even if it is occasionally
 handy.
 
+- Sample based timers are common in other situations (stop the code 100 times
+a second and record the stack). This sort of timing would not work with
+the reverse pass though.
+
 - There were a variety of other Stan interfaces that were proposed. The current
 one was chosen for simplicity.
 
--- a/designs/0013-profiling.md
+++ b/designs/0013-profiling.md
@@ -83,7 +83,7 @@ different blocks record in the same place.
 `profile` can be used to measure the expense of different parts of a loop or
 branches of a conditional statement).
 
-5. Timing statements can be nested, but they must be different.
+5. Timing statements can be nested, but they must have different names.
 
 6. A timing statement cannot be recursively entered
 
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
Tagging a few people that commented in previous discussions: @bob-carpenter @nhuurre @wds15
Mind giving this a look if you have time? Thank you!

yizhang-yiz:
All look good to me. I'm curious what happens if `profile` is in/over an MPI `map_rect` call. Off the top of my head it should work out of box except maybe clutters output due to multi-process I/O.

rok-cesnovar:
> I'm curious what happens if profile is in/over an MPI map_rect call.

```stan
profile("map_rect");
 target += sum( map_rect( lp_reduce , beta , theta , xr , xi ) );
```
this works as any other profile().

If its inside the function called in map_rect with MPI, then it will return the time spent in the process with rank 0. There is nothing in the design that prevents us from adding collecting profiling data from other MPI processes in the future, I just did not envision it in this first iteration. If the cluster consists of  heterogeneous machine that info might be interesting I imagine?

yizhang-yiz:
Yeah the "above" situation is straightforward. I wonder why putting`map_rect` in `lp_reduce` only prints rank 0. In this case wouldn't each rank calls `lp_reduce` hence `profile`?

rok-cesnovar:
Ah yes, you are correct, each process will collect its own profiling info and also print the info at the end, obviously. Not sure what got me confused there. Each MPI process is essentialy its own cmdstan run.

yizhang-yiz:
Don't get me wrong as I think the doc is great and ready. Maybe just add this to that disadvantage list along with recursive functions etc?

rok-cesnovar:
Done. Thanks for pointing it out!

betanalpha:
I see that it's addressed in the design document but from a user experience perspective I think that having named explicit scopes,
```
profile("profile-name") {
  function_call();
}
```
will be much less confusing than having scopes implicitly end.  By forcing the user to write out explicit scopes there's no ambiguity to where the profile scope ends, especially when nested profiles are included.

Moreover the scopeless profile behavior conflicts with the control statement conventions:
```
if (a < b)
x = 5;  // Only next line is within scope of conditional
y = 3;  // Following lines are not within scope of conditional
```
but
```
profile("name");
x = 5;  // Next line is within scope of profile
y = 3;  // But so it the next line
```

The difference is small but I think that it will avoid many questions from inexperienced users on the forums, especially when they can pattern match to conditional behavior.  At the same time that forcing users to write more braces doesn't limit the expressiveness of the profiling or adds a minuscule typing and formatting burden to the majority of use cases.

bob-carpenter:
I agree with @betanalpha that importing RAII conventions into Stan may be very confusing for users.  On the other hand, I'll find it a lot easier.

<b>RAII is like variable declarations</b>

Conditionals are not a good analogy to RAII, but variable declarations are.  Those work with exactly the same scope as being proposed for profiling.  

```
for (n in 1:N)
  real a = 0;
  x = 5 * a + c;  // OK: a in scope
  y = x / a;      // OK:  a still in scope
}
z = a * 3;        // ERROR: a out of scope (no masking in Stan, so a couldn't be defined elsewhere)
```

Introducing scope forces another layer of nesting, which is the main reason that I think I'll find it easier to use the RAII pattern.  Of course, RAII would also require another layer of nesting if you only want to pick out a subsequence of the statements in a block.

<b>Start / Stop Syntax</b>

Yet another alternative (which I do not support) would be start/stop profiling.  It never requires new blocks.  It'd look like this:

```
S1;
start_profile("a");
S2;
S3;
end_profile("a");
```

That's like the malloc/free pattern.  And it would presumably require the start and end to be in the same block, so we couldn't have:

```
for (n in 1:N) {
  S1;
  start_profile("a");
  S2;
}
S3;
end_profile("b")
```

It'd probably be a pain in the compiler to check that everything's properly nested.

rok-cesnovar:
Thanks to both. 

Another point for Micheal’s suggestion is that the end-of-transformed parameters-scope issue goes away, as there would be an actual scope there in that case as opposed to now where the end of TP bracket doesnt actually mean end of scope. 

But that is an implementation detail anyways.

> Start / Stop Syntax

I started with this version and the original Discourse proposal has this. As you mention, the nesting checks get tricky, as do checks a user stopped a profile they started or started a profile they are trying to stop. Its also more prone to frustrating typo-related bugs.

yizhang-yiz:
Does it make sense to do RAII & proposed one-liner at the same time? I like the consistency of RAII but also want to slack off by inserting a single line.

bob-carpenter:
> the end-of-transformed parameters-scope issue goes away,

That's a very good point in favor of explicit scoping.  That's also a place where scope in Stan doesn't match scope in C++. 

But now I'm confused about what would work.  Would we allow profiling statements to mix in with declarations like this for RAII?

```
transformed parameters {
  real a = cos(b);
  profile("foo");
  real c = a^2;  //  want this to still be a block-level parameter
}
```

How about scoping?

```
transformed parameters {
  real a = cos(b);
  profile("foo") {
    real c = foo(...);
  }
  ...
}
```

because we'd want c to still be a block-level variable.  Or is the thinking to not allow profiling of compound declare/define statements?

>  do RAII & proposed one-liner

The RAII style is the one-liner.  I'd strongly prefer not to allow two ways of doing this.

betanalpha:
> Conditionals are not a good analogy to RAII, but variable declarations are. 

I agree that variable declarations are the right analogy to the proposed scoping behavior.  In fact to really take communicate that correspondence I would suggest a profile mechanism that mimics variable declaration and help guide users to the expected behavior,
```
profile_timer start(“name”);
```
In the case the profile scope would be the same as the `start` variable scope.

> Introducing scope forces another layer of nesting, which is the main reason that I think I'll find it easier to use the RAII pattern.

As ungainly as nested braces can be when the nesting is deep, I thinking that deep nesting would be relatively rare and the clarity of explicit scopes would be worth it.  

In particular I’m imagining all of the Discourse questions from users who are experimenting with profiling because they want to speed up their model and just copy/pasting profile statements without really understanding what they’re doing.  Explicit scoping doesn’t fix this problem but I think it will help to moderate it enough to be worthwhile overall.

> Another point for Micheal’s suggestion is that the end-of-transformed parameters-scope issue goes away, as there would be an actual scope there in that case as opposed to now where the end of TP bracket doesnt actually mean end of scope.

Yeah, I really like that.

The cascading scope of variables in the programming blocks is often confusing for new users, although ultimately worthwhile given the functionality.


In any case I think a tweak to the notation to mimic something with the appropriate scoping (fake variable declarations of the current scoping rules, explicit scopes if following explicit scope rules) would help avoid confusion and prevent users from having to learn yet another pattern.

rok-cesnovar:
The more I think about, the more I think explicit scope is the way to go. The main reason being explaining to the user how it works.

Example:
```stan
transformed data {
  profile("td");
  real a = 1.0;
  real b = 0.0;
}
parameters {
  real<lower = a> y;
}
model {
  y ~ normal(b, 1);
}
```
Users are accustomed that the scope of `a` and `b` in Stan is from the line were defined in to the end of model or optionally GQ, so `a` is in scope both in the parameters as well as model block in the example, the `"td"` profile would however only be in scope until the end of transformed data.

Now we would have to document and explain that profiles have a different scope, from the line they were defined in to the end of the current block. Its not very complicated, but I would say its also not trivial for non programmers. Also having two different styles of scoping is probably not a good design.

This example
```
profile_timer start(“name”);
```
suffers from the same problem and also introduces a kind of object-oriented style.

One downside of the explicit scope approach is that profiling compound declare/define statements is very limited. But even the current RAII approach is not the most friendly for profiling compound declare/define. Lets say we would want to profile the two statements in transformed data in the example above. We can profile b's declare/define statement because its at the end of the block, but can't profile the same for `a`.

```stan
transformed data {
  real a;
  {
     profile("a");
     a = 1.0;
  }
  profile("b");
  real b = 0.0;
}
```
The only approach that is friendly for declare/define statements is the start/stop one. Or if we would say that we only have one active profile at a time

```stan
transformed data {
  profile("a");
  real   a = 1.0;
  profile("b"); // this stops profile a and start profile b
  real b = 0.0;
}
```

rok-cesnovar:
Also thanks for the discussion so far, it has been very helpful! Going to give it a day or two to hear from everyone interested and then rework the doc with the explicit scope approach.

bob-carpenter:
I find @rok-cesnovar and @betanalpha's arguments for the scoped version persuasive, because I think the most important factors in choosing among equally performant interface designs are (a) how easy it is to doc, and (b) how easy the the code is to read.

For (a), the explicitly scoped version is clearly superior.

For (b), the explicitly scoped version will be easier to read for novices or even people unfamiliar with Stan, whereas the RAII style may be easier for experienced users because it doesn't interrupt the structure of the program.  Being able to scan the structure of programs is super important for readability.   But I don't think that's enough to tip the scale toward the RAII version.

bob-carpenter:
@betanalpha --- Our replies crossed, so I just wanted to confirm that I'm sold on explicit scoping due to its being easier to read and doc more than anything else.

rok-cesnovar:
The doc is now updated with the explicit scoping approach and is ready for additional reviews.

rok-cesnovar:
Do we have anyine that wants to officially review this? @bob-carpenter maybe?

bbbales2:
Requesting review. We want to get this out for 2.26. Freeze + release candidate for that is on the 18th.

designs/0013-profiling.md
```diff
@@ -0,0 +1,420 @@
...
+  ...
+}
+```
+
+## Note on parameters block
+
+Because only parameter declaration statements are allowed in the `parameters`
+block, `profile` cannot be used there. This means that the `lower`, `upper`,
+`multiply` and `offset` constraints/transforms can not be profiled.
+
```
yizhang-yiz, line 242: Does it make sense to say in general we don't profile declaration, no matter which block it is in?

designs/0013-profiling.md
```diff
@@ -0,0 +1,420 @@
...
+- C++ calls to profile will be generated as:
+
+    ```cpp
+    {
+      profile<T__> profile("literal", const_cast<profile_map&>(profiles__));
+      ... statements
+    }    
+    ```
+    where X will be replaced by a unique number. The `const_cast` is required
```
yizhang-yiz, line 299: Where is `X`?

yizhang-yiz (changes requested):

designs/0013-profiling.md
rok-cesnovar: Ah, we no longer require these to be unique as

```
{
  profile<var> profile__("name1", ...);
  {
   profile<var> profile__("name2", ...);
  }
}
```
is valid C++ and we cant have two profiles in the same level block. Will remove.

designs/0013-profiling.md
rok-cesnovar: One could technically define a local variable inside a profile.

```stan
profile("name) {
    vector a = b +c;
    d = a / 2;
}
```

Maybe "we can not profile top-level declarations".

yizhang-yiz: Syntax still allows the following, does it?
```stan
parameters {
profile("xxx") {
real x;
real y;
}
//...
}
```

rok-cesnovar: No, because parameters only allows top-leve variable declaration and nothing else.

```stan
parameters{
    profile("yes") {
        real a;
        real b;
    }
    real y;
}
```
errors with  `Expected top-level variable declaration or "}".`

While 

```stan
parameters{
    real y;
    profile("yes") {
        real a;
        real b;
    }
}
```
errors with `Only top-level variable declarations allowed in data and parameters blocks.`

yizhang-yiz (dismissed):

yizhang-yiz (approved):

rok-cesnovar:
Thanks @yizhang-yiz !

bbbales2 (approved):
Approving since maybe I have write permissions here and @yizhang-yiz doesn't

bbbales2:
Lemme see if it's obvious how our Github permissions are different so we can fix yours @yizhang-yiz .

PR #36, opened 2021-02-10
adds proposal for experimental folder

SteveBronder:
This proposes allowing Stan projects to add an `experimental` and/or `unsupported` folder into their projects.

[Rendered docs here](https://github.com/SteveBronder/design-docs/blob/proposal/experimental/designs/0005-experimental.md)

### DIFF
--- a/designs/0005-experimental.md
+++ b/designs/0005-experimental.md
@@ -1,0 +1,50 @@
+- Feature Name: Experimental or Unsupported Folders
+- Start Date: 2021-02-10
+
+# Summary
+[summary]: #summary
+
+This proposes adding `experimental` and `unsupported` folders to all of the projects in `stan-dev`.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


mitzimorris:
> The main alternative is to keep our current system, which is fine though a bit too conservative for some projects to be accepted.

I am strongly in favor in keeping the current system, for the following reasons:

- Stan is OSS - folks are free to fork the repo and distribute their experiemental systems themselve.

- this creates new problems of gatekeeping the "experimental" projects.  if there's some incentive to be a part of Stan, because Stan is fairly widely known and used, then folks will want to get in on the game.  we don't want to get dragged into political and marketing battles.

- a corollory of the above is that such systems are likely to serve the needs of very small communities and therefore will take Stan resources away from focussing in the needs common to all Stan users.  and there are a lot of these.

I'm also worried that this will allow dependencies in an experimental system to influence the design and implementations of the core Stan libraries in ways that impede progress elsewhere.

riddell-stan:
PyStan 2 did have an experimental submodule. In retrospect, I think it was a bad idea, for the reasons that @mitzimorris  offers. I'd even emphasize the first point more. Not only is it possible to fork a project in principle, it's getting easier and easier to do so in practice---thanks to platforms such as GitHub, GitLab, gitea, etc.

PR #37, opened 2021-02-17
Adjoint ode design

wds15:
This PR discusses how the adjoint ODE approach can be integrated with Stan math. The focus of the document is to

- describe the merits of the approach
- how the CVODES user manual notation lines up with Stan math notation
- shows how the CVODES functionality is integrated with the Stan math autodiff library
- proposes a super-set of tuning parameters for the numerical procedures which need to be assessed by the community in order to weed them out somewhat (or not at all)

I am not sure on how to present a rendered version of this. I used this R command to render the text:

```R
rmarkdown::render("0027-adjoint-ode.md", output_format="bookdown::html_document2")
```

Here is a rendered pdf of the document of easier reading: 
[0027-adjoint-ode.pdf](https://github.com/stan-dev/design-docs/files/6064565/0027-adjoint-ode.pdf)

updated version 7. March 2021

[0027-adjoint-ode.pdf](https://github.com/stan-dev/design-docs/files/6097413/0027-adjoint-ode.pdf)

update 23rd March 2021

[0027-adjoint-ode.pdf](https://github.com/stan-dev/design-docs/files/6192512/0027-adjoint-ode.pdf)

update 7th April 2021

[0027-adjoint-ode.pdf](https://github.com/stan-dev/design-docs/files/6274031/0027-adjoint-ode.pdf)

### DIFF
--- a/designs/0027-adjoint-ode.md
+++ b/designs/0027-adjoint-ode.md
@@ -349,10 +349,15 @@ simplifcation is made for usability reasons and the emprical
 observation that neither any CVODES example codes nor Julia's
 DifferentialEquations.jl package does facilitate a vector absolute
 tolerance for the backward quadrature integration. A user may work
-around this limitation by scaling the states and the parameters which
-will affect the order of the gradients accordingly and therefore
-change the tolerance checks of CVODES whenever a gradient approaches
-zero.
+around this limitation to some extent by scaling the states and the
+parameters which will affect the order of the gradients accordingly
+and therefore change the tolerance checks of CVODES whenever a
+gradient approaches zero. As the lack of the feature of a vector
+absolute tolerance for the parameter quadratures may potentially pose
+a limitation for future uses of the adjoint ODE solver, the
+implementation in stan-math should be written in an manner allowing
+for an inclusion of this feature at a later time (for example through
+an overload of the proposed signature).
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
--- a/designs/0027-adjoint-ode.md
+++ b/designs/0027-adjoint-ode.md
@@ -208,14 +208,16 @@ Jacobian-vector products $\frac{df}{dy}\,v$, which we can compute
 directly using *forward mode*. In addition all 3 integration problems
 have their own relative and absolute tolerance targets.
 
-The propsed function should has the following signature:
+
+
+The proposed function should has the following signature:
 
 ```stan
-vector[] ode_adjoint(F f,
+vector[] ode_adjoint_tol_ctl(F f,
     vector y0,
     real t0, real[] times,
     real rel_tol_f, vector abs_tol_f,
-    real rel_tol_b, real abs_tol_b,
+    real rel_tol_b, vector abs_tol_b,
     real rel_tol_q, real abs_tol_q,
     int max_num_steps,
     int num_checkpoints,
@@ -234,7 +236,7 @@ The arguments are:
 6. ```abs_tol_f``` - Absolute tolerance vector for each state for
    forward solve (data)  
 7. ```rel_tol_b``` - Relative tolerance for backward solve (data)  
-8. ```abs_tol_b``` - Absolute tolerance for backward solve (data)  
+8. ```abs_tol_b``` - Absolute tolerance vector for each state backward solve (data)  
 9. ```rel_tol_q``` - Relative tolerance for backward quadrature (data)  
 10. ```abs_tol_q``` - Absolute tolerance for backward quadrature (data)  
 11. ```max_num_steps``` - Maximum number of timesteps to take in integrating
@@ -249,6 +251,37 @@ The arguments are:
 15. ```solver_b``` solver used for backward ODE problem: 1=Adams,
     2=bdf, 3=bdf_iterated  
 16. ```arg1, arg2, ...``` - Arguments passed unmodified to the ODE right
+hand side. The types ```T1, T2, ...``` can be any type, but they must match
+the types of the matching arguments of ```f```.
+
+In addition a function version is made available which mimics the
+existing ode interface for tolerances. This version will allow users
+to quickly try out the new adjoint interface. **The extra control
+parameters will be set based on best guesses as of now and are subject
+to change**:
+
+
+```stan
+vector[] ode_adjoint_tol(F f,
+    vector y0,
+    real t0, real[] times,
+    real rel_tol, real abs_tol,
+    int max_num_steps,
+    T1 arg1, T2 arg2, ...)
+```
+
+The arguments are:  
+1. ```f``` - User-defined right hand side of the ODE (`dy/dt = f`)  
+2. ```y0``` - Initial state of the ode solve (`y0 = y(t0)`)  
+3. ```t0``` - Initial time of the ode solve  
+4. ```times``` - Sorted arary of times to which the ode will be solved (each
+  element must be greater than t0)  
+5. ```rel_tol``` - Relative tolerance for all solves (data)  
+6. ```abs_tol``` - Absolute tolerance for all solves (data)  
+7. ```max_num_steps``` - Maximum number of timesteps to take in integrating
+  the ODE solution between output time points for forward and backward
+  solve (data)  
+8. ```arg1, arg2, ...``` - Arguments passed unmodified to the ODE right
 hand side. The types ```T1, T2, ...``` can be any type, but they must match
 the types of the matching arguments of ```f```.
 
--- a/designs/0027-adjoint-ode.md
+++ b/designs/0027-adjoint-ode.md
@@ -230,7 +230,7 @@ The proposed function should have the following signature:
 vector[] ode_adjoint_tol_ctl(F f,
     vector y0,
     real t0, real[] times,
-    real rel_tol_f, vector abs_tol_f,
+    real relative_tolerance_forward, vector abs_tol_f,
     real rel_tol_b, vector abs_tol_b,
     real rel_tol_q, real abs_tol_q,
     int max_num_steps,
@@ -246,7 +246,7 @@ The arguments are:
 3. ```t0``` - Initial time of the ode solve  
 4. ```times``` - Sorted arary of times to which the ode will be solved (each
   element must be greater than t0)  
-5. ```rel_tol_f``` - Relative tolerance for forward solve (data)  
+5. ```relative_tolerance_forward``` - Relative tolerance for forward solve (data)  
 6. ```abs_tol_f``` - Absolute tolerance vector for each state for
    forward solve (data)  
 7. ```rel_tol_b``` - Relative tolerance for backward solve (data)  
@@ -331,6 +331,28 @@ missing out for now is to exploit the sparsity structure of the
 ODE. This would allow for more efficient solvers and even larger
 systems, but this is not possible at the moment to figure out
 structurally the inter-dependencies of inputs and outputs.
+
+Of note is that the adjoint ODE method is not efficient by
+construction to calculate the gradient wrt to multiple outputs as
+needed for a Jacobian. Since for each function output the gradient for
+all parameters must be calculated by resetting the adjoints back to
+zero and then rerunning the backward sweep of AD (chain method) once
+again. Hence, this triggers for each function output gradient a rerun
+of the backward integration problem, which is not needed for the
+forward ODE method.
+
+The proposed function signature restricts the absolute tolerances for
+the quadrature backward problem to be a scalar only as opposed to
+allowing for a vector absolute tolerance which assigns to each
+parameter qudarature problem it's own absolute tolerance. This
+simplifcation is made for usability reasons and the emprical
+observation that neither any CVODES example codes nor Julia's
+DifferentialEquations.jl package does facilitate a vector absolute
+tolerance for the backward quadrature integration. A user may work
+around this limitation by scaling the states and the parameters which
+will affect the order of the gradients accordingly and therefore
+change the tolerance checks of CVODES whenever a gradient approaches
+zero.
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
--- a/designs/0027-adjoint-ode.md
+++ b/designs/0027-adjoint-ode.md
@@ -107,31 +107,31 @@ phase.
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-The ODE adjoint method for ODEs is relatively involved mathematically. The
-main challenge comes from a mix of different notation and conventions
-used in different fields. Here we relate commonly used notation within
-Stan math to the [user guide of
+The ODE adjoint method for ODEs is relatively involved
+mathematically. The main challenge comes from a mix of different
+notation and conventions used in different fields. Here we relate
+commonly used notation within Stan math to the [user guide of
 CVODES](https://computing.llnl.gov/sites/default/files/public/cvs_guide-5.6.1.pdf),
-the underlying library we employ to handle the numerical
-solution. The goal of reverse mode autodiff is to calculate the
-derivative of some function $l(\theta)$ wrt. to it's parameters
-$\theta$ at some particular realization of $\theta$; here $\theta$
-denotes a set of parameters. The function $l$
-now depends on the solution of an ODE given as initial value problem
+the underlying library we employ to handle the numerical solution. The
+goal of reverse mode autodiff is to calculate the derivative of some
+function $l(\theta)$ wrt. to it's parameters $\theta$ at some
+particular realization of $\theta$; here $\theta$ denotes a set of
+parameters. For example, the likelihood is defined as the normal log
+probability density for a single observation at time-point $T$. The
+mean of this normal log probability density is modelled as the
+solution of an ODE. The ODE itself depends on some parameters $p$
+which are a subset of $\theta$. The ODE is given as initial value
+problem
 
 \begin{align}
 \dot{y} &= f(y,t,p) (\#eq:odeDef) \\
  y(t_0) &= y^{(0)}. (\#eq:odeInitial)
 \end{align}
 
-The ODE has $N$ states, $y=(y_1, \ldots, y_N)$, and is
-parameterized in terms of parameters $p$ which are a subset of
-$\theta$. Let's now further assume for simplicity that the function
-$l$ only depends on the solution of the ODE at the time-point
-$T$. During the reverse sweep of reverse mode autodiff we are then
-given the autodiff adjoints of $a_{l,y_n}$ wrt to the output state
-$y(T,p)$. These
-are the partials of $l$ wrt to each state
+The ODE has $N$ states, $y=(y_1, \ldots, y_N)$. During the reverse
+sweep of reverse mode autodiff we are then given the autodiff adjoints
+of $a_{l,y_n}$ wrt to the output state $y(T,p)$. These are the
+partials of $l$ wrt to each state
 
 $$ a_{l,y_n} = \left.\frac{\partial l}{\partial y_n}\right|_{t=T} $$
 
@@ -268,12 +268,17 @@ The arguments are:
 hand side. The types ```T1, T2, ...``` can be any type, but they must match
 the types of the matching arguments of ```f```.
 
-In addition a function version is made available which mimics the
-existing ode interface for tolerances. This version will allow users
-to quickly try out the new adjoint interface. **The extra control
-parameters will be set based on best guesses as of now and are subject
-to change**:
-
+In addition a simpler function version which mimics the existing ode
+interface for tolerances should be made available once more experience
+has been gained with the new ODE adjoint method. While at the current
+stage it is premature to make such a simpler interface available, the
+simpler interface is intended to let users quickly try out the new
+adjoint interface. **As it is currently not clear how the extra
+control parameters will be set based the function signature will not
+be made available in the Stan language in the first version of the
+adjoint ODE solver**. Nontheless, once sufficient experience has been
+gained with the adjoint ODE method the following signature should be
+made available:
 
 ```stan
 vector[] ode_adjoint_tol(F f,
@@ -303,12 +308,19 @@ The best guesses based on preliminary experiments for the remaining
 tuning parameters are then set as
 
 - same relative tolerance in all problems
-- `abs_tol_f = abs_tol / 100`
-- `abs_tol_b = abs_tol / 10`
+- `abs_tol_f = abs_tol / 10`
+- `abs_tol_b = abs_tol / 3`
 - `abs_tol_q = abs_tol`
 - $250$ steps between checkpoints
 - bdf solver for forward and backward solve
 - hermite polynomial method for forward function approximation
+
+The above best guesses are only based on early experiments and will
+need to be refined. This design doc intentionally does not define when
+"sufficient" experience has been collected until the simplified
+function can be made available. However, the above defaults (or an
+updated set) should be included in the documentation of the adjoint
+ODE solver.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/designs/0027-adjoint-ode.md
+++ b/designs/0027-adjoint-ode.md
@@ -14,12 +14,12 @@ implement in Stan the so-called forward-mode ODE method in order to
 obtain in addition to the solution of the ODE the sensitivities of the
 ODE. The cost of this method scales as 
 
-$$ N * M . $$
+$$ N \cdot M . $$
 
 The computational cost of the adjoint method is much more favorable
 in comparison which is
 
-$$ 2 * N + M .$$
+$$ 2 \cdot N + M .$$
 
 The advantage of the adjoint methods shows in particular whenever the
 number of parameters are relatively large in comparison to the number
@@ -38,10 +38,10 @@ adjoint ODE solving method Stan will be able to scale to much larger
 problems involving more states and more parameters. Examples are large
 pharmacokinetic / pharmacodynamic systems or physiological based
 pharmacokinetic models or bigger susceptible infectious & removed
-models. As the adjoint ODE method will grow the computational cost
-only linearly with states and parameters, the modelers can more freely
-increase the complexity of the ODE model without having to worry too
-much about infeasible inference times.
+models. The adjoint ODE method will grow the computational cost only
+linearly with states and parameters and therefore the modelers can
+more freely increase the complexity of the ODE model without having to
+worry too much about infeasible inference times.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
@@ -54,12 +54,13 @@ package which we already include in Stan.
 
 From a more internal stan math perspective, the key difference to the
 existing forward mode solvers will be that the gradients are not any
-more precomputed. Instead, during the forward sweep of reverse mode
-autodiff, the ODE is solved first for the N states (without any
-sensitivities). When reverse mode AD then performs the backward sweep,
-the adjoints of the input operands are directly computed given the
-input adjoints. This involves a backward solve in time with N states
-and solving M quadrature problems in addition.
+more precomputed during the forward sweep of reverse mode autodiff
+(AD). Instead, during the forward sweep of reverse mode autodiff, the
+ODE is solved only for the N states (without any sensitivities). When
+reverse mode AD then performs the backward sweep, the adjoints of the
+input operands are directly computed given the input adjoints. This
+involves a backward solve in time with N states and solving $M$
+quadrature problems in addition.
 
 The numerical complexity is higher for the adjoint method in
 comparison to the forward method. While most of the complexity is
@@ -84,33 +85,124 @@ During an experimental phase of this feature we can hopefully learn
 which of these are relevant and which can be dropped for a final
 release of the `ode_adjoint` function.
 
-**Leaving the template text here for now**
-
-Explain the proposal as if it was already included in the project and you were teaching it to another Stan programmer in the manual. That generally means:
-
-- Introducing new named concepts.
-- Explaining the feature largely in terms of examples.
-- Explaining how Stan programmers should *think* about the feature, and how it should impact the way they use the relevant package. It should explain the impact as concretely as possible.
-- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
-- If applicable, describe the differences between teaching this to existing Stan programmers and new Stan programmers.
-
-For implementation-oriented RFCs (e.g. for compiler internals), this section should focus on how compiler contributors should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.
-
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-Here I would like to document exactly the math of the approach using
-the notation of the CVODES manual.
+The adjoint method for ODEs is relatively involved mathematically. The
+main challenge comes from a mix of different notation and conventions
+used in different fields. Here we relate commonly used notation within
+Stan math to the [user guide of
+CVODES](https://computing.llnl.gov/sites/default/files/public/cvs_guide-5.6.1.pdf),
+the underlying library we employ to handle the numerical
+solution. The goal of reverse mode autodiff is to calculate the
+derivative of some function $l(\theta)$ wrt. to it's parameters
+$\theta$ at some particular realization of $\theta$. The function $l$
+now depends on the solution of an ODE given as initial value problem
 
-TODO !!!
+\begin{align}
+\dot{y} &= f(y,t,p) (\#eq:odeDef) \\
+ y(t_0) &= y^{(0)}. (\#eq:odeInitial)
+\end{align}
 
-This is the technical portion of the RFC. Explain the design in sufficient detail that:
+The ODE has $N$ states, $y=(y_1, \ldots, y_N)$, and is
+parametrized in terms of parameters $p$ which are a subset of
+$\theta$. Let's now further assume for simplicity that the function
+$l$ only depends on the solution of the ODE at the time-point
+$T$. During the reverse sweep of reverse mode autodiff we are then
+given the adjoints of $a_{l,y_n}$ wrt to the output state $y(T,p)$ which
+are the partials of $l$ wrt to each state
 
-- Its interaction with other features is clear.
-- It is reasonably clear how the feature would be implemented.
-- Corner cases are dissected by example.
+$$ a_{l,y_n} = \left.\frac{\partial l}{\partial y_n}\right|_{t=T} $$
 
-The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+and we must calculate the contribution to the adjoint of the parameters
+
+$$ a_{l,p_m} = \sum_{n=1}^{N} \left. a_{l,y_n} \,  \frac{\partial y_n}{\partial p_m}\right|_{t=T}, $$
+
+which involves for each parameter $p_m$ the partial of every state
+$y_n$ wrt to the parameter $p_m$. Note that the computational benefit of the
+adjoint method stems from calculating the above sums *directly* in
+contrast to the forward method which calculates every partial
+$\left. \frac{\partial y_n}{\partial p_m}\right|_{t=T}$ separatley.
+
+In the notation of the CVODES user manual, the function $g(t,y,p)$
+**is equal to** $l(\theta)$ essentially. Through the use of Lagrange
+multipliers, the adjoint problem is transferred to a backward problem
+in equation 2.20 of the CVODES manual ([see here for a step-by-step
+derivation](https://arxiv.org/abs/2002.00326)),
+
+\begin{align}
+\dot{\lambda} &= - \left(\frac{\partial f}{\partial y}\right)^*
+\lambda - \left(\frac{\partial g}{\partial y}\right)^* (\#eq:lambdaOde) \\
+ \lambda(T) &= 0. (\#eq:lambdaInitial)
+\end{align}
+
+This ODE is referred to as the *backward* problem, since we fix the
+solution $\lambda$ at the final end-point $T$ instead of the initial
+condition at $t_0$ (it's a terminal value problem rather than an
+initial value problem). The CVODES manual then proceeds with deriving
+that
+
+\begin{equation}
+\left.\frac{d g}{d p}\right|_{t=T} = \mu^*(t_0) \, s(t_0) +
+\left.\frac{\partial g}{\partial p}\right|_{t=T} + \int_{t_0}^{T}
+\mu^* \, \frac{\partial f}{\partial p} \, dt.
+(\#eq:derivg)
+\end{equation}
+
+Here $s(t_0)$ is the state sensitivity wrt to the parameters at the
+initial time-point. The term $\left.\frac{\partial g}{\partial
+p}\right|_{t=T}$ is the partial derivative wrt to the parameters of
+$g(t,y,p)$. This term is determined by the terms in $g(t,y,p)$ which directly
+depend on the parameters and not implicitly due to the parameters
+affecting the ODE state. Thus, this term is being computed by the
+autodiff system of stan-math itself as part of the adjoints of the
+parameters, $a_{l,p_m}$. Finally, $\mu = \frac{d \lambda}{d T}$ such
+that $\mu$ is obtained by the equivalent backward problem (taking the
+total derivative of $\dot{\lambda}$ in \@ref(eq:lambdaOde) wrt to $T$)
+
+\begin{align}
+\dot{\mu} &= - \left(\frac{\partial f}{\partial y}\right)^* \, \mu
+(\#eq:muODE) \\
+\mu(T) &= \left(\frac{\partial g}{\partial y}\right)^*_{t=T}. (\#eq:muInitial)
+\end{align}
+
+If we now recall that $g(t,y,p)$ is equal to the function $l(\theta)$
+being subject to autodiff we see that
+
+$$\left(\frac{\partial g}{\partial y_n}\right)^*_{t=T} = a_{l,y_n}^*, $$
+
+which is the input we get during reverse mode autodiff (the conjugate
+transpose does not apply for Stan math using reals only).
+
+It is important to note that the backward ODE problem requires as
+input the adjoint $a_{l,y_n}^*$ such that the backward problem must be
+solved during the reverse pass of reverse mode autodiff. Thus, CVODES
+is used during the forward pass to *solve (forward)* the ODE in
+\@ref(eq:odeDef). Once the backward pass is initiated, the adjoints
+$a_{l,y_n}^*$ are used as the terminal value in order to solve the
+*backward problem* of \@ref(eq:muODE). Whenever any parameter
+sensitivities are requested, then we must solve along the backward
+problem an additional quadrature problem, which is the integrand in
+equation \@ref(eq:derivg).
+
+TODO: outline generalization to multiple time-points.
+
+In total we need to solve 3 integration problems which consist of a
+forward ODE problem, a backward ODE problem and a backward quadrature
+problem. The forward ODE and the backward ODE problem can be solved
+with either a non-stiff Adams or a stiff BDF solver of CVODES (the
+choice is independent for each problem). For the Newton steps of the
+BDF routine a linear solver routine is required. The routine can
+either be a dense solver or an approximate iterative solver. As we
+target large ODE systems it can be useful to allow users to choose the
+solver being used. For example, whenever the number of ODE states is
+very large, then an iterative solver may be preferable. The reason is
+that the dense solver will require the calculation of the full
+Jacobian $\frac{df}{dy}$ of the ODE RHS in \@ref(eq:odeDef) wrt to the
+states $y$. In contrast, the iterative solver only needs to evaluate
+Jacobian-vector products $\frac{df}{dy}\,v$, which we can compute
+directly using *forward mode*. In addition all 3 integration problems
+have their own relative and absolute tolerance targets.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -122,14 +214,23 @@ ODE. This would allow for more efficient solvers and even larger
 systems, but this is not possible at the moment to figure out
 structurally the inter-dependencies of inputs and outputs.
 
-# TODO: Update the rest
-
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-- Why is this design the best in the space of possible designs?
-- What other designs have been considered and what is the rationale for not choosing them?
-- What is the impact of not doing this?
+There is no other analytical technique available which scales in a
+comparable way. Hence, larger ODE problems ( many parameters and/or
+states) are currently out of scope for Stan and the adjoint technique
+does enable Stan to cope with these larger systems.
+
+The numerical complexities are rather involved as 3 nested
+integrations are performed. This makes things somewhat fragile and
+less robust. What makes the backward integration in
+particular involved is that the solution of the forward problem must
+be stored as a continuous function in memory and hence an
+interpolation of the forward solve is required. This is provided by
+CVODES via a checkpointing procedure. In summary, we do heavily rely
+on CVODES infrastructure as these building blocks are rather complex
+and heavy to craft on our own.
 
 # Prior art
 [prior-art]: #prior-art
--- a/designs/0027-adjoint-ode.md
+++ b/designs/0027-adjoint-ode.md
@@ -185,7 +185,11 @@ sensitivities are requested, then we must solve along the backward
 problem an additional quadrature problem, which is the integrand in
 equation \@ref(eq:derivg).
 
-TODO: outline generalization to multiple time-points.
+So far we have only considered the case of a single time-point
+$T$. For multiple time-points one merely starts the backward
+integration from the last time-point and keeps accumulating the
+respective terms which result from integrating backwards in steps
+until $t_0$ is reached.
 
 In total we need to solve 3 integration problems which consist of a
 forward ODE problem, a backward ODE problem and a backward quadrature
@@ -204,6 +208,51 @@ Jacobian-vector products $\frac{df}{dy}\,v$, which we can compute
 directly using *forward mode*. In addition all 3 integration problems
 have their own relative and absolute tolerance targets.
 
+The propsed function should has the following signature:
+
+```stan
+vector[] ode_adjoint(F f,
+    vector y0,
+    real t0, real[] times,
+    real rel_tol_f, vector abs_tol_f,
+    real rel_tol_b, real abs_tol_b,
+    real rel_tol_q, real abs_tol_q,
+    int max_num_steps,
+    int num_checkpoints,
+    int interpolation_polynomial,
+    int solver_f, int solver_b
+    T1 arg1, T2 arg2, ...)
+```
+
+The arguments are:  
+1. ```f``` - User-defined right hand side of the ODE (`dy/dt = f`)  
+2. ```y0``` - Initial state of the ode solve (`y0 = y(t0)`)  
+3. ```t0``` - Initial time of the ode solve  
+4. ```times``` - Sorted arary of times to which the ode will be solved (each
+  element must be greater than t0)  
+5. ```rel_tol_f``` - Relative tolerance for forward solve (data)  
+6. ```abs_tol_f``` - Absolute tolerance vector for each state for
+   forward solve (data)  
+7. ```rel_tol_b``` - Relative tolerance for backward solve (data)  
+8. ```abs_tol_b``` - Absolute tolerance for backward solve (data)  
+9. ```rel_tol_q``` - Relative tolerance for backward quadrature (data)  
+10. ```abs_tol_q``` - Absolute tolerance for backward quadrature (data)  
+11. ```max_num_steps``` - Maximum number of timesteps to take in integrating
+  the ODE solution between output time points for forward and backward
+  solve (data)  
+12. ```num_checkpoints``` number of steps between checkpointing forward
+    solution (data)  
+13. ```interpolation_polynomial``` can be 1 for hermite or 2 for
+    polynomial interpolation method of CVODES  
+14. ```solver_f``` solver used for forward ODE problem: 1=Adams,
+    2=bdf, 3=bdf_iterated  
+15. ```solver_b``` solver used for backward ODE problem: 1=Adams,
+    2=bdf, 3=bdf_iterated  
+16. ```arg1, arg2, ...``` - Arguments passed unmodified to the ODE right
+hand side. The types ```T1, T2, ...``` can be any type, but they must match
+the types of the matching arguments of ```f```.
+
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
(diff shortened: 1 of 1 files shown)


wds15:
For anyone who wants to get his hands dirty with the running prototype of this, you need to:

- clone cmdstan with the stan and stan-math submodule
- cd into `cmdstan/stan/lib/stan_math/`
- checkout the branch `feature/adjoint-odes`
- move back to cmdstan directory
- grab the latest stanc experimental version created by Rok as listed on the PR https://github.com/stan-dev/math/pull/1905
- Currently that means to head to https://github.com/stan-dev/stanc3/actions/runs/621697392
- from there you grab `mac-stanc` (for mac) which you put into cmdstan/bin. I then copy this thing manually to bin/stanc and make sure that macos allows you to run this thing by right-clicking on the binary in the Finder while holding down the Control key. Then press "open" and from now on you can run the binary from the command line. This stuff is not needed on Linux... no idea about windows.
- at this point it should be a matter of building cmdstan the usual way
- Stan models can now include a `ode_adjoint_tol` or `ode_adjoint_tol_ctl` call.

An example model I used for benchmarks is in `cmdstan/stan/lib/stan_math/ode-adjoint/linked-mass-flow.stan` .

bbbales2:
@wds15 use the cmdstan tarball builder. Docs for it are hidden [here](https://github.com/stan-dev/example-models/pull/202)

It will allow you to build a tarball that is easy to share and use in cmdstanr (see the closures tarball [here](https://github.com/stan-dev/stanc3/pull/742#issuecomment-787099251))

funko-unko:
Alright, so this is how it worked for me via @wds15 's instructions:
```
# Assumes that a compatible `stanc` file is located at ./stanc
# Uses 8 cores to build cmdstan
git clone --recurse-submodules https://github.com/stan-dev/cmdstan
cd cmdstan/stan/lib/stan_math
git checkout -b feature/adjoint-odes origin/feature/adjoint-odes
cd ../../../
mkdir bin
cd ..
cp stanc cmdstan/bin/stanc
cd cmdstan
chmod +x bin/stanc
make build -j8
```

And then this showed up, which also worked for me: 
https://github.com/stan-dev/math/pull/1905#issuecomment-791361097

I've found this template for the design doc:
https://github.com/stan-dev/design-docs/blob/92e173efc04aaa3d3dae277dcc7b8681f1008543/0000-template.md

I guess we want to keep to that template? Is a "stan programmer" as mentioned in the template a user of stan ("programming stan models") or a developer of stan? I guess a simple example comparing forward/adjoint would be helpful both for users and developers?

Anyhow, some comments:

* As described in the doc, the complexity of the forward mode grows *quadratically*, not _exponentially_.
* `num_checkpoints` should be renamed into something like `num_steps_between_checkpoints` (as we talked about).
* Increasing `num_steps_between_checkpoints` too much degrades performance, I suspect because the memory needed gets allocated anew at every leapfrog step? Not something that we should worry/optimize about for now I guess. Also, making this too large will of course result in failure to allocate the memory. I guess this should not really be a problem?
* Different values of `num_steps_between_checkpoints` appear to lead to different results and then to (not insignificantly) different samples. Maybe something to keep in mind. 
* The interface looks fine to me. 

What needs to be done?

wds15:
What we need to do in my view is

- Find out if the simplified interface has sane defaults setup which work "OK" for many problems.
-  Gather experience with the adjoint method in problems. So find out when it is better to use adjoint is what would be good to be able to answer to some degree with some accuracy.
- Collect as much information as we can about making recommendations for the different knobs of the Ctl method. This will be hard to do in general, but maybe we can still distill some rules of thumb.
- For example: I found that it is better - I think - to have greater precision (absolute tolerance being smaller) for the forward solve as for the backward solve and the quadrature stuff should have the least precision (highest absolute tolerance value).

Did I miss any option to expose for the Ctl thing?

Thanks for the other comments, will update the doc accordingly.

funko-unko:
Hmmm, I believe at least for the interpolation method the polynomial interpolation appears to be preferred. The manual (https://computing.llnl.gov/sites/default/files/public/cvs_guide-5.7.0.pdf) states

> The Hermite cubic interpolation option is present because it was implemented chronologically first and it is also used
by other adjoint solvers (e.g. daspkadjoint). The variable-degree polynomial is more memory-efficient (it requires
only half of the memory storage of the cubic Hermite interpolation) and is more accurate.

As for the number of steps between checkpoints, there probably is no way to tell a priori which number is best. `250` sounds fine I guess, most ODE problems in Stan will still be rather small, won't they? With 1k `double` valued unknowns, that would mean 2 MBs of RAM? The AD tape is probably larger (and also kept in memory?)?

As for the tolerances, :man_shrugging:. We'll need a suite of testcases for that.

I do believe more and simpler (scalable) examples would be useful, not only to evaluate the use cases of the adjoint method. They could also be presented to users as examples when which methods perform well?

wds15:
I will switch to the polynomial method for the simpler method then. Maybe we even drop the option for the polynomial entirely?

For the steps between checkpoints: I tested with the example values like 10, 50, 250, 500 and found that you see speedups when going from 10 to 250, but then it got flat. In the CVODES examples they use 100s (100 or 500) as I recall. Memory wise I would not expect an issue here at all. ODE problems are small in a Bayesian context as I have seen it (or you will have to wait forever for your results).

Tolerances: I opt for making the chosen tolerance the least tolerance you get from this. So forward & backward have higher precision and the tolerance given is the one from the computed parameter gradients.

For now we need to come to the conclusion that this is all good enough to land in Stan. Maybe we mark the simple interface as "experimental" in the doc to warn people that we may change the underlying numeric details? Do you have any speicfic examples in mind before proceeding?

Tagging @charlesm93 who wanted to get his hands dirty on this.

betanalpha:
Section 1:

We should be consistent with what terms are kept in the scaling.  Either N (M + 1)  and 2 * N + M or N * M and N + M.

I think that it’s worth noting the overheads here, even just C_{sens} = N * (M + 1) + overhead_{sens} and C_{adj} = 2 * N + M + overhead_{adj}, and mentioning that the adjoint overhead is larger and can be limiting for small systems.  This will also better prepare the reader for the discussion in Section 6.


Section 3: 

"the adjoints of the input operands are directly computed given the input adjoints” reads pretty awkwardly.  Firstly to avoid confusion I recommend being explicit and redundant and everywhere referring to "autodiff adjoints" when referring to the tangents/cotangents passed along the expression graph and “system adjoint” or “ODE adjoint” when referring to backwards solve method for propagating the autodiff adjoints through the ODE solve.  Secondly using input to refer to both the operands and the autodiff adjoints is a big confusing given that each refers to a different direction through the expression graph.

Section 4:

Derivations for general l(theta) can be hard to connect back to the autodiff circumstance; I recommend always mentioning that taking l(theta) = delta * y(t) will recover the adjoint-jacobian product needed for reverse mode autodiff.

I recommend noting that lambda is a vector, one for each state, and the backwards solve (5) requires solving a coupled system of N equations (hence the cost of N).  Each parameter sensitivity is one-dimensional which is why each of the corresponding integrals can be solved with a straightforward, one-dimensional quadrature (it might also help to refer to “one-dimensional quadrature” everywhere instead of just “quadrature”).

I think it would help to list the control parameters currently assumed for the current integrate_ode_bdf signature.

I really don’t think that we understand these solvers well enough to implement a `ode_adjoint_tol` function with the initial feature.  Even ignoring the danger of having some implicit defaults this signature assumes that that uniform tolerances across all of the solvers is useful for which I haven’t seen any strong support.  Users will be able to quickly try out the new function with recommended defaults for all of the control parameters that can be cut and paste into the Stan program.  This has the added benefit of communicating exactly what was run (to correlate with any performance results they share) instead of relying on the exact repo commit when the default parameters might be changing.  In the end I don’t see any strong reason for including this signature.

Section 6:

Blanket statements like “ODEs with many states and parameters are currently out of scope for Stan” can be more confusing than useful (everyone had different computing resources and hence thresholds that define practically useful tools, “scope” is ambiguous, performance depends on not just the solve time but also the behavior of the likelihood functions as the number of states and parameters increases, etc).  What’s wrong with just saying that the performance of the current methods scale poorly with the number of states and parameters, but the adjoint method offers better scaling?  For a fixed computational budget the adjoint method may allow for larger systems to be fit, and for a fixed system it may speed up existing fits.


Section 7:

I think that it will be helpful to differentiate between prior uses of the adjoint method for specific functions and the adjoin method for general autodiff.  The adjoint method for general functions is used all over the sciences but the requirement of a fixed function makes it too rigid for general purpose tools like Stan (indeed this is why we didn’t consider the adjoint method much easier even though it was always in CVODES).  Using the adjoint method for adjoint-Jacobian products, however, can be integrated into any reverse mode autodiff implementation.  The application to autodiff is the really exciting new bit (the functionality built into tools like JAX for neural ODEs would be a recent prior art).

wds15:
@betanalpha thanks for looking over it.

- Section 1: I adapted accordingly.
- Section 3 & 4: Will work on that. 
- Section 6: I reworded it.
- Section 7: I adapted as suggested.

Wrt. to the suggested `ode_adjoint_tol` signature which sets some defaults: The thought is that users should be able to quickly try out the adjoint method. So when users already spend some thinking on their ODE problem as they use the `tol` flavor, then it should really be easy for users to just try adjoint. I think we should be able to setup a default which gives reasonable performance (maybe not best). This is also a big argument to encourage users to try the prototype. It's just about swapping out the function name. I would suggest to keep the simpler interface at least for getting users to test this prototype during this month... @funko-unko liked the short function call for basically those reasons given.

wds15:
@betanalpha ok, now I hopefully got your other comments straight as well. Thanks for looking into it. I am now talking of "autodiff adjoints" and "ODE adjoint states" as you suggest, which makes sense to hopefully get down the confusion level...

I suggest to now launch until Easter an experimental phase with Stan users. The goal is to collect some feedback on

- the utility of the method
- the utility of the tuning parameters (we can probably drop the interpolation_polynomial, for example... I think)
- the utility of the `ode_adjoint_tol` call

We can collect feedback on the forum and in summarized form in the wiki.  After Easter we then proceed with finishing this up and merging it to stan math during April such that it will be ready for the next May release.

Sounds like a plan?

Tagging @charlesm93, @bbbales2 , @rok-cesnovar & @funko-unko ... @yizhang-yiz

funko-unko:
Can we maybe just slap a warning on the `ode_adjoint_tol` function? Something along the lines of 

> The other implicit tuning parameters may change in the future, the current setting is equivalent to `ode_adjoint_tol_ctl(..., values of implicit parameters)`.

such that the user may just swap out the function call? Maybe with a notice to include this message if they post in the forums? This would discourage the use of the `ode_adjoint_tol` function while still making it easy to "just use" the adjoint method, and we would always know all of the tuning parameters?

PS: Clearly my opinion should have very little weight.

rok-cesnovar:
Longterm we might want to use tuples or even better structs (though I am not sure there is any work being done onr structs) for default arguments.

So the signature becomes:
```stan
vector[] ode_adjoint_tol_ctl(F f,
    vector y0,
    real t0, real[] times,
    (real rel_tol_f, vector abs_tol_f, real rel_tol_b, vector abs_tol_b,
    real rel_tol_q, real abs_tol_q, int max_num_steps, int num_checkpoints,
    int interpolation_polynomial, int solver_f, int solver_b)
    T1 arg1, T2 arg2, ...)
```
and you could call it as:

```stan
ode_adjoint_tol_ctl(udf, y0, t0, times, default_ode_adjoint_args(), ...)
```
where default_ode_adjoint_args() would be a built-in function that would return a tuple with the default arguments.

One could also do this:
```stan
transformed data {
  (real, vector, real, vector, real, real, int, int, int, int, int) args = default_ode_adjoint_args();
  args.7 = 500; // would set max_num_steps
}
```
Structs with named elements obviously works much better for this.

betanalpha:
Great, thanks. 

That all sounds good, although again I think `ode_adjoint_tol` is premature at the moment.  Introducing it offers the danger that users will not understand its limitations (it doesn’t matter how many warnings you put in, if you allow defaults people will use them and expect them to work) and we can always add it later without any real cost.

betanalpha:
Could we not get the same effect using variadic arguments for all of the control parameters?

rok-cesnovar:
Current the compiler can not deduce where the control arguments stop and UDF args start. 
For example:
```stan
vector[] ode_adjoint_tol_ctl(F f,
    vector y0, real t0, real[] times,
    real rel_tol_f, vector abs_tol_f, real rel_tol_b, vector abs_tol_b,
    real rel_tol_q, real abs_tol_q, int max_num_steps, int num_checkpoints,
    int interpolation_polynomial, int solver_f, int solver_b)
```
```stan
vector[] ode_adjoint_tol_ctl(F f,
    vector y0, real t0, real[] times,
    real rel_tol_f, vector abs_tol_f, real rel_tol_b, vector abs_tol_b,
    real rel_tol_q, real abs_tol_q, int max_num_steps, int num_checkpoints,
    int interpolation_polynomial, int arg1, int arg2)
```
are the same. We could add logic where we check what args `f` has and deduce based on that and add missing control args for the C++ call (C++ compiler can not do the same). But that would mean stanc3 would hold the ode default values and I am not sure I like that.

So doable but not sure its worth. I agree with "if you allow defaults people will use them and expect them to work".

betanalpha:
Yeah, I guess we’d need named arguments. 

Regardless the defaults would have to be stored somewhere, either in each program, in the ODE function implementation, or in some default parameter object.  Personally I’d prefer the first initially and then the second eventually, if only to reduce the code associated with the ODE implementations (i.e. default configurations are defined and docs in the ODE  functions directly and no searching is required).

wds15:
I just updated some more details.

@betanalpha did I get your comments in as you had in mind?

It would be good to get the design doc agreed so that it's ready to be merge by the end of this month.

The only points left open at this stage should be settled during the experimental phase which ends at Easter. The questions open are

- Does the `ode_adjoint_tol_ctl` function have all required tuning parameters?
    * Can some parameters be dropped?
    * Are some additional ones needed?
- Should we keep the simpler `ode_adjoint_tol` function signature?
    * Are the defaults sensible?

As it has been a sticking point to include the simpler `ode_adjoint_tol`: Is it also a possibility to include the function, but mark it in the documentation clearly that the defaults can be changed with releases? That would mimic essentially the statement about the default priors in rstanarm, which are not guaranteed to stay the same between releases.

Pinging @bbbales2 @yizhang-yiz @rok-cesnovar @funko-unko @charlesm93

betanalpha:
Mostly small comments.  I'm still not comfortable with including a simpler function in the initial release.  _Users will not respect any "experimental feature" warnings_ and they will complain about changing behavior from version to version even if the behavior is explicitly documented as not stable.   Even when they don't complain they often end up using the same function without realizing the difference, which is already a huge problem with `rstanarm` priors.

Section 1

“severely costly to fit” -> “expensive to evaluate”.  The autodiff method is independent of how hard ODE-based models are to fit.

ressources -> resources

This matters less in the design doc and more in final user facing doc, but given the heterogeneity in terminology across fields we should try to mention as many common terms as possible.  For example not just “so-called forward-mode ODE method” but rather “forward-mode, direct, sensitivity”.

Section 2

“susceptible infections & removed” -> “susceptible, infected, and recovered (SIR)”

Section 3

“merely” -> “many”, although I think “more” by itself is sufficient here.

“in addition to solve M one-dimensional quadrature problems" -> "in addition to solve M one-dimensional quadrature problems that depend on the adjoint states".  To harm being clear to hear to avoid any misunderstanding about the solves parallelizeable or the like.

"We need to first collect some experience before a simpler version can be made available (if that is feasible at all)."  I very much agree, but then why is a simpler version being considered for the initial implementation?

Are the quadrature tolerances different for each of the M integrations?

Section 4

Use theta or p consistently to denote the parameters throughout.

g(t, y, p) will not equal l(p) for the autodiff application -- it will equal dy/dt = f(t, y, p) which allows us to propagate the autodiff adjoint information from any subsequent expression graph.

wds15:
Thanks @betanalpha !

> Mostly small comments. I'm still not comfortable with including a simpler function in the initial release. Users will not respect any "experimental feature" warnings and they will complain about changing behavior from version to version even if the behavior is explicitly documented as not stable. Even when they don't complain they often end up using the same function without realizing the difference, which is already a huge problem with rstanarm priors.

I did change my mind in the process of all this. So I do think that we should supply a simplified version to make it easy for users to try this thing out (which is why there could be left-overs saying otherwise). Let's be pragmatic here and I suggest to proceed as follows: We don't provide the simpler version in the Stan language itself. Instead, we pack the simplified function call into the Stan user documentation in the form of a Stan function. ... unless another reviewer here can convince @betanalpha otherwise, of course (sounds unlikely ;) ). Is that reasonable? 

> Are the quadrature tolerances different for each of the M integrations?

Right now the quadrature tolerances are all the same... but I just checked, the CVODES interface allows to specify the absolute tolerance for each parameter separately (I thought that is not possible). Should we change that and make that also a vector for the absolute tolerances? That's an awful lot of stuff to specify then and it's a bit inconvenient as the length of the vector changes whenever you add or drop a parameter...even worse is that the correspondence of absolute error tolerance to actual parameter is not so transparent in that we have a variadic interface, but the absolute tolerances would correspond to a "flattened" version of the variadic parameter pack... for the sake of simplifying at least something already now, I would leave things as they are right now and keep the same absolute tolerance for all parameters. Agree?

> Use theta or p consistently to denote the parameters throughout.

My intention is to keep "theta" as denoting a super-set of parameters and p is a subset of parameters needed for the ODE. I would like to keep notation as is, since this sets apart notation in stan-math (theta) and notation in the CVODES manual (p).

> g(t, y, p) will not equal l(p) for the autodiff application -- it will equal dy/dt = f(t, y, p) which allows us to propagate the autodiff adjoint information from any subsequent expression graph.

Maybe not ideal as it is right now... I will have one more look, but I would like to make it as simple as possible. Possibly we just say that l(theta) is - as an example - a normal lpdf of an ODE state at time T.

rok-cesnovar:
> Instead, we pack the simplified function call into the Stan user documentation in the form of a Stan function

I agree, this way the default values are out in the open in the UDF code and you also get the option to switch to adjoint ODEs "easily".

wds15:
There is just one catch with putting the function into the manual: It would be a user-defined Stan function and we do not support variadic arguments to user defined functions (or dow we?).

Anyway, we can still agree at a later time-point on the simpler function being part of the language itself... unless anyone else reviewer this has different views.

@betanalpha is that a plan?

betanalpha:
> Are the quadrature tolerances different for each of the M integrations?
> 
> Right now the quadrature tolerances are all the same... but I just checked, the CVODES interface allows to specify the absolute tolerance for each parameter separately (I thought that is not possible). Should we change that and make that also a vector for the absolute tolerances? That's an awful lot of stuff to specify then and it's a bit inconvenient as the length of the vector changes whenever you add or drop a parameter...even worse is that the correspondence of absolute error tolerance to actual parameter is not so transparent in that we have a variadic interface, but the absolute tolerances would correspond to a "flattened" version of the variadic parameter pack... for the sake of simplifying at least something already now, I would leave things as they are right now and keep the same absolute tolerance for all parameters. Agree?
> 
But from the point of view of the solver that’s a completely arbitrary simplification.  If we can’t argue _why_ a configuration should be simplified then we shouldn’t simplify it.

Yes the variadic pattern complicates this but parameters can also have arbitrary scales which means that useful tolerances in the quadratures can vary wildly.

> Use theta or p consistently to denote the parameters throughout.
> 
> My intention is to keep "theta" as denoting a super-set of parameters and p is a subset of parameters needed for the ODE. I would like to keep notation as is, since this sets apart notation in stan-math (theta) and notation in the CVODES manual (p)
> 
Then you have to define theta and p accordingly.  Right now p comes out of nowhere if you haven’t read the CVODES docs.

Heterogeneity of notation is a general problem and the best we can do is discuss all of the common notations and translate between them.

> g(t, y, p) will not equal l(p) for the autodiff application -- it will equal dy/dt = f(t, y, p) which allows us to propagate the autodiff adjoint information from any subsequent expression graph.
> 
> Maybe not ideal as it is right now... I will have one more look, but I would like to make it as simple as possible. Possibly we just say that l(theta) is - as an example - a normal lpdf of an ODE state at time T.
> 
Yes,  simple as possible _but no simpler_.  One of the reasons why this adjoint stuff is so confusing is that so much of the writing tries to simplify too far making it impossible to see necessary abstraction.  In this case trying to motivate the adjoint solver from the usual “I care only about l(y, theta) and not all of the y independently” perspective makes it really hard to see how the method can be used for general autodiff applications where l(y, theta) is an implicit function of the rest of the program that the ODE node won’t have any idea about.

> There is just one catch with putting the function into the manual: It would be a user-defined Stan function and we do not support variadic arguments to user defined functions (or dow we?).

I’m not even sure we need a user-defined function.  If the goal is to get people to test out some arbitrary default behavior and report back on their results then why not just write out an example with nominal values for the solver configuration and ask people to copy/paste into their programs?  That way they can use whatever variadic pattern is appropriate for the rest of the signature and there’s it puts the responsibly of using the experimental configuration on the user.  Honestly if a user isn’t conformable with that kind of copy/pasting then they’re not appropriate for testing an experimental feature (which is okay — part of the goal here is to be very clear about the target audience).

wds15:
> But from the point of view of the solver that’s a completely arbitrary simplification.  If we can’t argue _why_ a configuration should be simplified then we shouldn’t simplify it.

> Yes the variadic pattern complicates this but parameters can also have arbitrary scales which means that useful tolerances in the quadratures can vary wildly.

Reasons not to expose absolute tolerances by parameters:

- The usability of the function is severely complicated.
- User should anyway scale their parameters to unity (or not too far off). If users cannot deal with unity scaling, then an advanced method is probably also a bit too hot / anyone who can deal with an ODE system requiring an adjoint method should be able to scale to unity. We should doc that.
- None of the CVODES examples use an absolute tolerance for each parameter. All of the CVODES examples (and there are quite a few) use a single absolute tolerance parameter for the quadrature.
- After all it should not matter too much speed wise. The hard part of all of this is the forward/backward integration. The quadrature is cheap numerically in comparison. Thus, one can simply be a bit more cautious on the absolute tolerance and set it to the lowest required value there is.

I really don't think we should expose a vector absolute tolerance for the quadratures given all these reasons. The reasons above are hopefully sound to others as well.

> I’m not even sure we need a user-defined function.  If the goal is to get people to test out some arbitrary default behavior and report back on their results then why not just write out an example with nominal values for the solver configuration and ask people to copy/paste into their programs?  

I would like to put something into the doc which is like abs_tol_f = 1E-8, abs_tol_b=abs_tol_f*3 and abs_tol_q=3*abs_tol_b... as an example. This combination hopefully gives good results in many circumstances and will allow people who are able to use the `tol` interface of the other solvers to immediately get started with the adjoint solver which has the specialty to require so many tolerances.

I will try to clarify the other bits about notation and stuff in my next commit.

wds15:
Bump. PDF is updated as well.

bbbales2:
I like scalar tolerances on the quadratures. I like the simplicity and we've gotten this far with scalar tolerances.

funko-unko:
Huh, I was wondering why there would only be a single absolute tolerance for the parameters...

Maybe allowing the possibility to specify per-parameter abs. tols. can be helpful. We do not know, do we? But as @bbbales2  said, so far everything worked out with a single tolerance, and the interface itself and its implementation looks very painful..

wds15:
@funko-unko I gave a number of reasons as to why we do not need per-parameter absolute tolerances. What do you think about these?

funko-unko:
@wds15 Hm, I think I agree with them/you in that we shouldn't include parameterwise absolute tolerances for now.

BTW, do we know when these absolute tolerance become important? What values are compared with these tolerances? I wouldn't think the parameters themselves, but their adjoints? 

I would assume then that loose absolute parameter tolerances become a problem when our parameters are a bad fit for the data, i.e. are in the tail of the posterior, where the gradient will not be very steep?

wds15:
Just a note to myself: The adjoint method has the disadvantage that it's notoriously inefficient for Jacobian calculations as then we have to run the backward sweep for every output. That is obviously not an issue for stan applications.

Other than that I do think that the design is fine for approving and merging. I will do one last pass the next days and then it would be great to merge it once someone here approves, of course.

betanalpha:
> Reasons not to expose absolute tolerances by parameters:


> The usability of the function is severely complicated.

Same argument can be made for any of the state-varying tolerances, but one of the main insights we’ve had is that a single tolerance just isn’t enough in complicated problems.  We want the functions to be as simple, and hence as usable, as possible but no simpler.  Forcing simplicity despite the computational consequences is the original sin that we made with ODEs in the first place!
> User should anyway scale their parameters to unity (or not too far off). If users cannot deal with unity scaling, then an advanced method is probably also a bit too hot / anyone who can deal with an ODE system requiring an adjoint method should be able to scale to unity. We should doc that.
Same augment can be made for any of the state-varying tolerances.  Why are the quadrature tolerances different?
> After all it should not matter too much speed wise. The hard part of all of this is the forward/backward integration. The quadrature is cheap numerically in comparison.
It’s cheap but there’s no reason why the accuracy wouldn’t be as sensitive to the quadrature tolerances as the forward and backward solve tolerances.  
> Thus, one can simply be a bit more cautious on the absolute tolerance and set it to the lowest required value there is.

Again, same argument can be made for any of the tolerances.  If anything treating the forward/backwards tolerances differently from the quadrature tolerances makes the function _less_ usable because the best practice for setting the different tolerances are not consistent.

A single scalar tolerance _could_ be sufficient for most problems of interest, but at the moment that is as much speculation as the decision to make all of the other tolerances scalar or vector (unless anyone has actual experience with adjoint solvers in which case I welcome their contributions).  My point is that while we’re stepping back and recognizing the need to research the empirical consequences of these tolerances more carefully we should have the initial function signature be as generic as possible to facilitate that research, and not force rigidity that ends up becoming problematic until we have actual evidence that it’s not problematic.

I agree that the variadic signature complicates matters significantly, although if anything this speaks to the limitations of a variadic signature when one wants to be able to control as much of the ODE solver as possible.  Personally I would prefer the initial implementation to be a non-variadic signature that allows for trivial vector quadrature tolerances that allows for the question to be researched and resolved empirically.  Yes it’s annoying to use a non-variadic signature, but it’s straightforward to translate between the two and more variadic signatures can always be added later.  

If a variadic signature is deemed necessary then we should at least have documentation about not only how to compensate for the need for varying quadrature tolerances by lowering the single scalar tolerance but also empirical behaviors one should look out for to identify that need.  If not enough is known about those behaviors then I would suggest again that we don’t actually know enough for a single tolerance signature to be sufficient at first.

bbbales2:
> but one of the main insights we’ve had is that a single tolerance just isn’t enough in complicated problems

Did we have that insight?

I read through the discourse thread (https://discourse.mc-stan.org/t/adjoint-ode-prototype-rfc-please-test/21212) and here and the math PR and didn't find it.

> A single scalar tolerance _could_ be sufficient for most problems of interest, but at the moment that is as much speculation as the decision to make all of the other tolerances scalar or vector (unless anyone has actual experience with adjoint solvers in which case I welcome their contributions)

Why is this speculation? Sounds like a few people have models running with this which seems like actual experience.

wds15:
I really think we need the variadic interface in order to make it easy for users to jump on the adjoint thing. I mean this thing is meant for *large* models with lots of parameters. Chances are high that this model exists in a form written for the variadic interface by now. In that case a user will not invest in porting it to an old interface which imposes painful restrictions. So variadic is really what we want here to make it simple to specify the model. This way we will gain more experience with the method faster (and a variadic interface is basically the norm by now for ODE things).

Apart from that - yes sure, we can add some doc on how to get around the restriction that the quadratures have a single absolute tolerance... that will require scaling states and parameters such that the partials come out at a changed order as needed.

However, we have models running with this interface already such that there is experience with this function in its current form.

And again - none of the CVODES examples (of which there are many) uses a vector absolute tolerance for the quadratures. That does suggest to me that this is not a truly urgent feature and people can solve their problems without it (well).

betanalpha:
> I really think we need the variadic interface in order to make it easy for users to jump on the adjoint thing. I mean this thing is meant for large models with lots of parameters. Chances are high that this model exists in a form written for the variadic interface by now. In that case a user will not invest in porting it to an old interface which imposes painful restrictions. So variadic is really what we want here to make it simple to specify the model. This way we will gain more experience with the method faster (and a variadic interface is basically the norm by now for ODE things).
> 
The variadic signature fundamentally limits how well the ODE solver can be configured, and this limitation was never considered in the original variadic design discussion.  Yes it somewhat facilitates building large systems (although the packaging of variables into arrays really is not that difficult and the burden of porting isn’t much more than a few minutes of programming) but what use are those larger systems if they can’t be solved reliably?  Preferring one choice over the latter simply because it has become the status quo is, in my opinion, not a compelling argument.

The adjoint method is really powerful but the complexity of the adjoint system introduces many new opportunities for unexpected behavior.  Limiting the configuration options not only limits how much control users have, but it also sets the expectation that those inaccessible configurations are not relevant.  Any awkward behavior arising from the inability to control those inaccessible configurations then becomes a burden for users who have to be able to recognize and manage that behavior, let alone report it.  I am not very comfortable with the prioritizing of putting this new functionality into as many user programs as possible based on the hope that the solver configurations that are hard to set will conveniently be unimportant.
> Apart from that - yes sure, we can add some doc on how to get around the restriction that the quadratures have a single absolute tolerance... that will require scaling states and parameters such that the partials come out at a changed order as needed.
> 
The doc also needs information on what happens when the scaling is wrong.  If it were straightforward to work out the scales for those partials a priori then _all_ of the tolerances would be rendered unnecessary with proper scaling of the ODE system.

I’m also not sure that the suggested partials intuition is sufficient.  The quadrature scales should be determined by how much the adjoint-Jacobian product for each parameter _varies_ along the evolution of the system, not the final sensitivities or adjoint-Jacobian product.  This variation is then determined by the specific dynamics — my guess is that they will be more important in non-equilibrium systems where the nature of the dynamics is consistently changing but that’s just speculation.
> And again - none of the CVODES examples (of which there are many) uses a vector absolute tolerance for the quadratures. That does suggest to me that this is not a truly urgent feature and people can solve their problems without it (well).
> 
But the CVODES examples are all based on single solves with fixed parameters.  We can’t ignore the larger scope of ODE discussions in Stan where that single solve context is not particularly representative of the Stan context where the parameters are varying quickly and the ODE solver needs to be robust enough for all of those solves.  The relevance of any existing ODE best practices to Stan has to be taken with some skepticism until we better understand the quickly varying parameter context.


It’s clear that I’m in the minority here.  The proposed functionality exposes many solver configuration options but not all, and those inaccessible are missing because of technical limitations with the variadic signature rather than any strong theoretical/empirical evidence.  Of course we can’t ignore technical limitations in the design but what I’m not seeing here are discussions of how the consequences of forcing a scalar quadrature scale will be studied and validated moving forwards.  Is there anything planned other than simply hoping that no one complains?

funko-unko:
> Yes [the variadic signature] somewhat facilitates building large systems (although the packaging of variables into arrays really is not that difficult and the burden of porting isn’t much more than a few minutes of programming)

To be honest, I haven't really been using/exploiting the variadic signature, so I would actually be fine with the old signature + parameterwise tolerances. @betanalpha is right that we really do not know what level of control we have to expose, so maybe we should really expose as much as possible? Maybe an option would be to provide (something like) the old interface with parameterwise tolerances? Then again, we/you do not want to add/maintain to many different signatures/implementations...

Maybe we should also advise users to check that the tolerances are appropriate for parameters from the prior, as done here https://users.aalto.fi/~timonej3/case_study_num.html ? Meaning that users should *always* check convergence of the trajectories/sensitivities?

wds15:
The variadic interface is not what is preventing us to expose the per-parameter absolute tolerances. The ODE functions are variadic and they are meant to stay like that. Thus, the adjoint ODE function also needs to be variadic. Also, the variadic interface simplifies to a non-variadic one if you choose to ignore the variadic bits.

I firmly think that we do not need the functionality of a per-parameter absolute tolerances. In addition to CVODES not having it their examples, the library DifferentialEquations.jl also does not expose such level of granularity - and DifferentialEquations.jl has been designed to be integrated into parameter estimation frameworks where parameters will vary wildly as in Stan.

I also personally do not think that we need to study the consequences of a scalar parameter tolerances. If someone considers this important, then please suggest a strategy for it. To me it is absolutely fine if some problems remain out of scope from this implementation. At least to start with this is a very rasonable choice and I also do not think that a large class of problems won't be solvable; in fact, the reverse is true is what I would think in that we can solve the vast majority of problems. 

What we have right now looks to be a major feature of Stan and in it's current form will be a big plus for the Stan user base. In case we discover some real problems where we can't get things to work satisfactory, then we can still come back and extend the existing implementation (should we not be able to work around issues by scaling of states and parameters).

From my view we should now concentrate on what we have in order to meet the deadline for the upcoming release.

With that I would request that we now merge this design doc in its current form. I am happy to document that the per-parameter tolerances are not being exposed on purpose and that these are left out for the moment being as a future feature should that be deemed necessary.

wds15:
Bump. I just updated the design doc with the limitations as discussed.

It would be great if we can merge this design doc by next week.

As we have probably not reached full alignment I wonder if we can proceed or need a voting thing to be setup? I don't hope we need that, but if yes, then we probably need to give that process some prep (and we would need to ask how we actually do this).

@rok-cesnovar I also spelled out all the arguments for the tolerances, but that should not change the C++, maybe the error messages from the parser? Not sure, but now naming is entirely aligned with the other ODE solvers.

bbbales2:
> I firmly think that we do not need the functionality of a per-parameter absolute tolerances.

Yeah I'm in this camp too.

I tried digging in the cvodes doc for what the vector absolute tolerances even meant:

> The quadrature variables z can be optionally included in the error test, in which case corresponding relative and absolute tolerances must be provided.

Now I'm just more confused on this stuff. Apparently these tolerances are straight up optional in some cases?

@yizhang-yiz do you have time to look at this design doc?

bbbales2:
> To me it is absolutely fine if some problems remain out of scope from this implementation.

I also agree with this, and this has a clear use case for systems where the number of parameters in the system gets large (so we should be enabling some of those problems).

As far as ODE signatures go, if this turns out to be a big mistake, this isn't so hard to patch up. We didn't need to introduce any new library code for this, and it would be easy to remove the adjoint solver, mark the function deprecated, and wire in another ODE solver as a compatibility implementation.

wds15:
> As far as ODE signatures go, if this turns out to be a big mistake, this isn't so hard to patch up. We didn't need to introduce any new library code for this, and it would be easy to remove the adjoint solver, mark the function deprecated, and wire in another ODE solver as a compatibility implementation.

No need to deprecate the current function. We could just introduce an overload where the absolute tolerances for the parameters are a vector instead of a scalar. The implementation can be written in a way that this extension is easy to plumb through (but it will be a bit of work). Actually I would not deprecate the current interface, since the usability of the more complex per-parameter function is much worse in case you do not need this functionality.

wds15:
Now... I think is time to finalise this. As such I am suggesting to go through a voting procedure starting **today** and lasting for two weeks. This means that the voting period ends on Monday April the 26th 2021. The final result will then be available on April 27th. Please vote for 

1. reject the design PR in its current form
2. accept the design PR as is

I would appreciate any rationale for rejecting should you vote for that option. This would help to improve the document for a second run.

I am tagging anyone coming to my mind who have known ODE background or have been involved in some way in this PR so far.

@bbbales2 @betanalpha @funko-unko @charlesm93 @rok-cesnovar @yizhang-yiz @syclik @SteveBronder @bob-carpenter 

Also tagging @martinmodrak who suggested voting.

I will cross-post this call for vote on discourse, but the process itself should only happen as comments on the PR.

Many thanks everyone!

The PR is accepted if a majority (>50%) of votes go for option 2. Please cast your vote by noting the numeric option 1 or 2 which you prefer. People having been part of this PR are eligible to vote or those reading into the doc and have solid knowledge/experience with ODEs in Stan.

yizhang-yiz:
@bbbales2 sorry that I've been busy last wk. Will go through the thread asap.

charlesm93:
I read through the design doc and I think overall it's very well written. The only comments I have is that I expect the adjoint method to work better than forward diff also in the small number of parameters, large number of states regime -- to be tested, I'll put something together -- and that one reference we could include is the work on Neural ODEs by Chen et al (https://arxiv.org/abs/1806.07366).

I agree with @betanalpha and @funko-unko that we should have per-parameter absolute tolerances for the quadrature step. This is consistent with other absolute tolerances we have, I don't think it's a burden to users who won't use it, and it can help us study adjoint methods for ODEs in a Bayesian context. It seems a reasonable application would be a case where we use varying absolute tolerances for the forward and backward solves. CVODEs looks at different contexts and the Julia package is fairly novel -- though I'd be interested to hear their reasoning.

In my view, the most compelling reason for not having varying absolute tolerances right now is that it's more work for the developers and it'll delay the feature's release -- Can someone confirm this? That's fair enough. We can release the version as is but the design doc should acknowledge varying absolute tolerance is a desirable extension if not a top priority. And also that design as is doesn't hinder this extension. This strikes me as an acceptable compromise.

I'd love some documentation on the benefits of scaling parameters in ODEs, some examples, and how well this works in practice!

funko-unko:
I believe part of the problem with settling the question whether or not these parameterwise tolerances are needed is that there is no quick/easy way to estimate the impact of these tolerances, neither for us in the very general setting, nor for users with a specific model.

I believe the only way we have right now to estimate this is to just run full HMC and see how the sampler performs, which will take "forever". I think it would help a lot if it were (easily or at all) possible to explore how the solver performs for samples from the prior, i.e. what tolerances are needed for a given (estimated) final error in the density and in the gradient.

The first you can do in a slightly hacky way, but for the latter you have to jump through a thousand hoops. I strongly believe that this should be made very very (very) easy for the user. 

One way to do this would be to be able to [pass a csv file to `fixed_param`](https://discourse.mc-stan.org/t/fixed-param-but-for-whole-fit/21384/13?u=funko_unko). And while we're at it, users should also be very interested in the amount of time it takes to integrate the ODE for a given set of parameters, which as we all know may vary widely across the parameter space.

wds15:
At this point it would be great to really get to the point of rejecting (option 1) or accepting (option 2) with your post. Rationale as to why a specific option has been chosen is a great bonus, of course.

The design doc already has a short discussion under "Drawbacks" about the lack of the per-parameter tolerances. Here are a few more points to this:

1. We have not yet seen any case where the per-parameter absolute tolerances would have been needed. The empirical observation is that other communities (CVODES, Julia) do not make use of such a feature and hence the need for this feature should be driven by an actual problem which we cannot solve with the simplified approach. Moreover, I think that we should have a discussion about wether we need this feature or not only once we have proven that there is a need. Per se the default is that the feature is not needed for real-world problems for which Stan is being used for.
2. Our current ODE solvers also do not allow for different tolerances for anything. They are all the same and no one has ever complained about it. There were discussions on extending that to by-state, but that did not happen for various reasons. In any case, all Stan programs using ODEs so far have been working nicely with just a single absolute tolerance for states and gradients. Thus, if this assumption is wrong by now, then this actually affects all our solvers and we should treat this separately from adjoint would be my view.
3. The usability is severely complicated given one has a variadic interface. One would have to respect ordering of parameters and their storage order. That can be avoided if the user restricts to only using 1d array inputs, but still. 
4. Indeed, my own resources are limited and in addition I do personally not think that the per-parameter thing is good for users as it complicates usability a lot. 
5. The per-parameter tolerance thing can be emulated - to some extend and with some work - using scaling and shifting of parameters and states. This won't work perfect, but that is totally possible to do,
6. Finally, the code will be written in a way so that the per-parameter tolerance thing can be added at a later point - should it be deemed useful for problems we encounter (and deem these problems to be important enough to warrant an implementation).

So far we do not even have any strong use-case for per-parameter tolerances for the parameters. Instead, we do have seen that the existing system gives users huge speedups.

yizhang-yiz:
What's the established voting procedure on design doc and on github?

wds15:
I am not calling out for an established formal ruling here. Instead I brought up the topic at the Stan Gathering last week Thursday April 8th and it was suggested to go through a voting here in the way I started it (at least this is how I understood the meeting outcome). To me that is reasonable given that

- we are in principle in alignment on the proposal in that we want adjoint ODE
- we seem to have different views on a specific feature
- I would assume that we can resolve this now with a voting procedure in a constructive and collaborative manner
- I also assume that we can trust each other here as I don't see any reason as to why someone would intend to manipulate, cheat or do any other odd things... we do agree on the goal to get adjoint ODE in such that I would expect people to be constructive here.

I hope that makes sense to everyone involved.

SteveBronder:
In the past for these we had a one week voting period where folks would use the github review process to review or request changes. If the majority is approve then we merge the design doc. 

If we wanted anyone to be able to vote we could also just have a comment people could thumbs up or down. I'm fine with either but prefer the review process so that all the tallies stay at the bottom of the page and dissents show up as `request review` comments which are easier to note looking over a proposal's history

yizhang-yiz:
Thanks @SteveBronder . Looks like this procedure is missing in the [voting proposal discussion](https://discourse.mc-stan.org/t/revised-sgb-proposal-for-voting-on-technical-issues/17063/1).

SteveBronder:
I don't want to get the doc discussion off track so maybe this would be better to discuss in the general meeting Thursday. The actual review process in the [design doc readme](https://github.com/stan-dev/design-docs#the-process) is a bit vague, which tbh I think is good actually as it lets the scope of the voting pool be decided by the reviewer(s). Whatever way people want to vote is cool and good by me

betanalpha:
We’re in an infinite loophole  and no progress will be made without a vote, but I want to point out a few things to have my peace.
> Our current ODE solvers also do not allow for different tolerances for anything. They are all the same and no one has ever complained about it. There were discussions on extending that to by-state, but that did not happen for various reasons. In any case, all Stan programs using ODEs so far have been working nicely with just a single absolute tolerance for states and gradients. Thus, if this assumption is wrong by now, then this actually affects all our solvers and we should treat this separately from adjoint would be my view.
By this reasoning the adjoint signature should not have any vector tolerances.  The fact that the non-variadic arguments expose vector tolerances, as was strongly motivated by stability issues that are starting to arise in more complicated ODE models that are outside of the typical CVODES scope, completely invalidates this line of reasoning.
> The usability is severely complicated given one has a variadic interface. One would have to respect ordering of parameters and their storage order. That can be avoided if the user restricts to only using 1d array inputs, but still.
The adjoint solver requires a sophisticated configuration that can have strong affects on the solver stability that we do not yet fully understand (not only how to optimally configure but also how to diagnose poor configurations).  Yes we will always have to balance functionality with robustness, and we will never be able to release functionality the we understand completely.  The original ODE solvers in Stan were developed when users were working with much simpler ODE systems, but contemporary applications, especially the large systems for which the adjoint method will be advertised, as much more complex.

Prioritizing simple function signatures to maximize easy of use externalizes the consequences of any poor solver configuration to users who are least prepared to investigate any problems that arise.  A non-variadic signature would be a frustration but not an obstruction for the more experienced users working with complex ODE systems who would benefit from this speedup the most.
> Indeed, my own resources are limited and in addition I do personally not think that the per-parameter thing is good for users as it complicates usability a lot.
Exposing _any_ of the solver configuration complicates usability. 

Limiting what is exposed by forcing suboptimal defaults only shifts the complication from how easy it is to incorporate the function into a Stan program to how easy it is to diagnose and investigate stability problems in the model fit.  Usability arguments like this would make a lot more sense if the decision was between exposing _no_ solver configuration options vs exposing _all_ solver configuration options, but the current decision is between exposing just the configuration options that are immediately compatible with the variadic signature verses all configuration options.  
> The per-parameter tolerance thing can be emulated - to some extend and with some work - using scaling and shifting of parameters and states. This won't work perfect, but that is totally possible to do,
I would love to see any non-speculative guidance on how this can be accomplished.  The quadrature tolerances will depend on how parameter Jacobians evolve between the observed times _relative_ the adjoint states.  These differential behaviors cannot be controlled by constant, linear transformations of the system that leave the Jacobians invariant.
> Finally, the code will be written in a way so that the per-parameter tolerance thing can be added at a later point - should it be deemed useful for problems we encounter (and deem these problems to be important enough to warrant an implementation).
> So far we do not even have any strong use-case for per-parameter tolerances for the parameters. Instead, we do have seen that the existing system gives users huge speedups.
> 
Placing the burden of proof on one outcome but not the other by framing that outcome as the default and the other as alternative is entirely artificial here.

Right now every comparison of ODE solver performance has been complicated by the tolerance configurations — difference solvers can leapfrog each other, and become more or less stable, with non-default tuning.  There is no question that the adjoint method will be extremely useful in the right circumstances, but it also exposes even more configuration options that can drastically effect its performance making the method more fragile.

State tolerance sensitivity has proven to be important for previous solvers, and the current signature exposes all of the state tolerances accordingly for both the forward and reverse solves.  While the exposure of the full configuration is somewhat independent of introducing an adjoint solver, this will be the first solver that we’ve introduced since the community has really appreciated how sensitive performance can be to these configurations.  This is all great so far.

The adjoint solver, however, also requires one-dimensional quadrature that parallels the reverse adjoint solve.  No existing ODE method requires a similar quadrature and so there’s no local evidence from which we can draw about how important the quadrature tolerances might be.  The closest analog is the 1D integrator exposed in Stan; although the quadrature algorithm isn’t exactly the same the tolerances in the 1D integrator can be very important for getting reasonable results out.  Beyond the lack of empirical evidence no one has made any theoretical arguments regarding under what circumstances the adjoint quadrature would be robust or not so we have little understanding on how fragile the quadrature might be and, more importantly, how to identify any problems arising from a poor configuration.  The situation is very much parallel to the state tolerances and to me that requires burden of proof for any argument to _not_ include quadrature tolerances.

The only arguments for including solver tolerances but not quadrature tolerances — independent of design considerations — are based on extrapolating a few documentation examples in packages focusing on different use cases than Stan.  What frustrates me here is that the current arguments are all based on maintaining the variadic signature, which was introduced without any consideration for parameter-dependent solver configurations.  Instead of considering that fundamental design incompatibility might be an argument for limiting the use of variadic signatures when parameter-dependent solver configurations are relevant the variadic solver is being taken as mandatory and incompatible functionality being rejected by default.

I very much hope that the quadrature tolerances do not end up that important, or that there are straightforward ways to identify poor configurations of those tolerances so that users can at least understand when the adjoint functionality might not be stable enough for their use, so that simpler signatures are sufficient.  I do not think we have anywhere near enough evidence to support a simpler signature, however, and exposing one prematurely shifts all risks to unprepared users.  At the very least I hope that everyone voting yes here will accept the responsibility of monitoring the forums for problems and providing consistent user support to minimize the costs of a poor user experience.

I continue to advocate for a more careful rollout, first exposing only a signature that requires the full adjoint solver configuration to be tuned so that users understand the requirements of using the functionality and can experiment and provide feedback.  If that feedback motivates a simpler signature then we can expose new functions to the language in the future.  Forcing speculative signatures, on the other hand, risks deprecated functions that complicate documentation and user experience.

mitzimorris:
I have read through this thread and was present at the online meeting where the issues were clearly laid out and discussed by Sebastian and Charles, among others.
Sebastian's proposal is both pragmatic and promising.  I vote yes.

wds15:
As per discussion today with @charlesm93 I have added a small sentence to clarify that a vector absolute tolerance for the parameters can be added later to the implementation. That is, the C++ code should be written in a way to allow for an extension should this be deemed necessary at a later point. Otherwise no change.

Great to see one vote here (even better that it is positive from my personal perspective).

bbbales2:
I vote yes. Reasons over [here](https://discourse.mc-stan.org/t/call-for-vote-on-adjoint-ode-design-doc-pr/21829/2)

funko-unko:
For what it's worth, I also vote yes. But I do see @betanalpha's point.

rok-cesnovar:
I also give a positive vote. Even though I am familiar with the underlying implementation and have worked on the Stanc3 side of things, I do have to acknowledge that I am by no means an expert in modeling with ODEs so put a lesser weight on my vote :)

Comments: 
- I agree with Micheal's comments about this being a somewhat arbitrary subset of exposed arguments, though I think the reasoning for this particular one seems reasonable to me (easier to implement, no per-parameter tolerances are currently being used in ODE solvers in Stan/CVODES/Julia, etc) so I think this is a good step 1 that does not hinder us for any future evolutions of the ODE adjoint interface.

- This implementation does not make a per-parameter tolerances implementation any more difficult to implement in the future if we find that to be needed. The only cost is that we would have 2 Stan function signatures but the backend code would overlap so there is not much maintenance cost there. I do not think we will have to deprecate the existing one as it seems to be working for at least the problems currently used.

- By waiting with the per-parameter tolerances implementation we could add much friendlier and less error-prone signature in the future once tuples are introduced (maybe for this release but probably 50/50 at this point) as instead of parameters we would supply a (parameter, tolerance) tuple. So the signature would be

```stan
vector[] ode_adjoint_tol_ctl(F f,
    vector y0,
    real t0, real[] times,
    (real rel_tol_f, vector abs_tol_f, real rel_tol_b, vector abs_tol_b,
    real rel_tol_q, real abs_tol_q, int max_num_steps, int num_checkpoints,
    int interpolation_polynomial, int solver_f, int solver_b)
    (T1 arg1, T1 arg_tol), (T2 arg2, T2 arg2_tol) ...)
```

So bottom line for me is: this seems useful (I trust the judgement of Sebastian, Charles and the rest), should not put a 
significant maintenance burden on us even if the per-parameter tolerance implementation would become available.

SteveBronder:
+1 I vote yes!

charlesm93:
I also vote yes.

wds15:
BUMP.

6 positive votes and no (definite) negative vote by now.

Thanks for everyone taking the time to dive into this so far.

Voting closes on Monday 26th April 2021.

syclik:
My vote: yes.

wds15:
BUMP.

Last day for voting is today!

7 positive votes and no (definite) negative vote by now.

betanalpha:
I vehemently vote no.

The intent of this implementation was to expose all of the tolerances (and not give them defaults) so that advanced users would be forced to experiment and provide feedback.  Prioritizing the variadic signature negates this intent, both in limiting what tolerances can be configured and allowing unexamined use of the method.  The design fundamentally changed halfway through the review process solely to rationalize merging a feature.

There is zero principled arguments for why the quadrature tolerances, and only the quadrature tolerances, can be ignored and no one has provided any strategy for how problems due to misconfigured quadrature tolerances can be, let alone will be, identified and studied.

Note also that in the history of Stan we have never actually deprecated a function from the language.  Premature features will persist and users will continue to incorporate them into their Stan programs.

yizhang-yiz:
> Note also that in the history of Stan we have never actually deprecated a function from the language

IMO Allowing deprecation is a healthy to do in the long term. Maybe we should have a discussion on that at some point.

rok-cesnovar:
> Note also that in the history of Stan we have never actually deprecated a function from the language.

The functions are deprecated in the sense that the compiler will warn the users that a function or some syntax (˙#˙ is deprecated for comments, as is `<-`, etc.). We have not removed them as that would require a bump of the major version and when that will happen requires a bit more planning and discussion then just a typical release cycle. There is no immediate plan on when we will do that.

I have personally been slowly collecting things that will need to be removed once we bump (will open an issue somewhere at some point). Some already have an issue in one of the repos, others do not. 

So far collected:

Language:
- _log, _cdf_log, _ccdf_log suffixed functions
- multiply_log, binomial_coefficient_log, cov_exp_quad
- integrate_ode_* functions
- get_lp, if_else, increment_log_prob
- abs in favor of fabs (there is flip-flopping so this one is a maybe)
- the old array syntax in favor of `array[]` (another maybe, I dont think this was decided, though it would simplify code)

Others:
- bin/print in cmdstan
- remove Math code associated with all deprecated functions

wds15:
My understanding form the start was to expose a function signature which exposes a large set of tuning options - and we do have that now. In fact, I personally was for including an additional bdf variant, but in discussions over it I did concur with others that this functionality is not needed. Everything is a trade off.

There is no variadic vs per-parameter thing here, not at all. Also, I really do not see a need to deprecate at some point the proposed signature in favor for the more complicated one with per-parameter tolerances. Deprecating the simpler signature would only be needed if almost all cases we encounter require the more complicated signature - but that is really unlikely (time will tell).

Let's see it positive: We are really making progress here! Solving large ODE systems in Stan will be faster for a very large class of problems (at least for those we benchmarked).

betanalpha:
> The functions are deprecated in the sense that the compiler will warn the users that a function or some syntax (˙#˙ is deprecated for comments, as is <-, etc.). We have not removed them as that would require a bump of the major version and when that will happen requires a bit more planning and discussion then just a typical release cycle. There is no immediate plan on when we will do that.

Yes, a major version bump has been “planned” for many years now with political and logistical issues preventing any progress.  Deprecation is absolutely a useful tool in software development but until we can demonstrate that we can actually implement deprecation as a development team its use to motivate aggressive feature releases is, in my opinion, weak.
> There is no variadic vs per-parameter thing here, not at all.
> 

Quoting Sebastian:

"Right now the quadrature tolerances are all the same... but I just checked, the CVODES interface allows to specify the absolute tolerance for each parameter separately (I thought that is not possible). Should we change that and make that also a vector for the absolute tolerances? That's an awful lot of stuff to specify then and it's a bit inconvenient as the length of the vector changes whenever you add or drop a parameter...even worse is that the correspondence of absolute error tolerance to actual parameter is not so transparent in that we have a variadic interface, but the absolute tolerances would correspond to a "flattened" version of the variadic parameter pack... for the sake of simplifying at least something already now, I would leave things as they are right now and keep the same absolute tolerance for all parameters. Agree?"

A signal quadrature tolerance was used in the initial signature based on your misunderstanding of CVODES.  Honest mistake.  But reframing that initial signature as a default that’s “good enough” is not.

> Also, I really do not see a need to deprecate at some point the proposed signature in favor for the more complicated one with per-parameter tolerances. Deprecating the simpler signature would only be needed if almost all cases we encounter require the more complicated signature - but that is really unlikely (time will tell).

The exact same argument holds for _any_ of the solver configuration arguments.  Why are we exposing any of them?  Why are we just now exposing vector tolerances?  Why did everyone agree that a no-tolerance signature was a bad idea?

We’re starting to recognize that they have substantial influence on how the solvers behave and we have _no idea_ which are important.  At best the choice of which tolerances are exposed and not exposed is arbitrary, at worst it’s based on forcing the variadic signature.
> Let's see it positive: We are really making progress here! Solving large ODE systems in Stan will be faster for a very large class of problems (at least for those we benchmarked).
> 
Introducing a new feature that is not fully understood is progress only when there are not poor consequences for users.  Being able to fit larger problems is useful only if the solvers maintain their robustness in those larger problems, especially as prior specification becomes more and more subtle in higher-dimensional spaces.  

I have no illusions of convincing anyone here; everyone is very optimistic that things will be fine based on little to no evidence.  I again just hope that everyone voting “yes” because they speculate that there won’t be problems will take responsibility and do the work to review user feedback for stability problems due to the adjoint quadrature configuration and help users work through those issues.

wds15:
Voting period is over as of today 27th April 2021

Result:

- positive votes: 7
- negative votes: 1

Thus, the decision is that we accept this PR as is ... and merge it.

Thanks for everyone taking the time to go through the material!

The PR for this in stan-math is getting along very well.

martinmodrak:
I don't have an opinion on the technical aspects of the design doc, but the discussion prompted me to think about culture I would personally like to see in the Stan dev community (while others may obviously differ).

I think it is important to bear in mind that everybody here shares a big portion of values - we all want Stan to be useful, reliable, usable, fast, ... We all value the effort others put into improving Stan, we all want our perspectives to be treated seriously. Disagreements like this are IMHO rarely due to some deep fundamental differences. Instead they arise partly from genuine uncertainty about what a design choice will actually entail in the real world and partly from somewhat different weights each of us puts on individual values/goals and how we evaluate the uncertainty. And I think this should be a reason for humility - some of the judgements any of us makes will be wrong. Some of our values/goals may turn out to lead to outcomes we didn't envision and don't like. Moreover whether a decision will in fact turn out to be good in practice  doesn't in my experience correlate neatly with ability to convince others or to come up with good arguments. To me a consequence is that in most projects one should expect to rarely have all their requirements/expectations/... met - otherwise it would likely mean that the project only reflects the values of some of its team members.

This humility also IMHO entails that we should try to avoid looking at results of votes (or any decision) as personal wins or losses, or try to keep scores on who was right how many times, or engage in some form of "I told you so" when it turns out somebody was wrong. We share most of our goals and we all benefit if Stan works great and lose out when there are issues. And I felt it is important to remind ourselves of this from time to time.  End of my two cents.

PR #41, opened 2021-05-03
Complex scalar, vector, and matrix types

bob-carpenter:
This is a proposal to add complex scalar, vector, and matrix types to the Stan language.

### DIFF
--- a/designs/0030-complex-numbers.md
+++ b/designs/0030-complex-numbers.md
@@ -1,0 +1,307 @@
+---
+output:
+  html_document: default
+  pdf_document: default
+---
+
+# Complex Numbers Functional Specification
+
+This is a proposal for adding complex number, vector, and matrix types to the Stan language. This involves proposals for
+
+- language types, constructors, and accessors
+- I/O format for the generated C++ model class
+- input format for JSON and R dump
+- suggested data formats for the CmdStan, R, and Python interfaces
+
+This proposal is backed by a full implementation of the [C++ complex number standard](https://en.cppreference.com/w/cpp/header/complex) in the Stan math library. 
+
+At their heart, complex numbers are nothing more than pairs of real numbers with a bunch of special functions defined over them.
+
+
+## Goals
+
+- The goal of this feature is to allow users to code models involving complex numbers.
+
+- The goal of this document is to be precise enough that
+    - users can understand the feature well enough to evaluate it, and
+    - developers can work on a technical specification for the implementation.
+
+## Design considerations
+
+- The main design consideration is that Stan programs involving complex numbers be easy to write, and just as importantly, easy to read.
+
+- We should follow the basic design of Stan types, including promotion structure and associated covariance (in the CS not stats sense).
+
+- I/O structures need to
+    - balance self documentation and efficiency, and
+    - be easy to adapt to the interfaces and analysis tools.
+
+
+## Assumptions
+
+- The language will target C++ code in the Stan math library.
+
+
+## Stan data types
+
+The following four basic data types will handle all of the complex number functionality we need:
+
+- `complex`:  scalar type
+- `complex_vector`: vector of complex values
+- `complex_row_vector`: row vector of complex values
+- `complex_matrix`: matrix of complex values
+
+As usual, arrays can be constructed from any of these types.
+
+### Constructors
+
+A complex number is constructed from a pair of real numbers, one representing the real component and one the imaginary component.  The proposal is to have a function do that work.
+
+```
+real r;
+real i;
+complex z = to_complex(r, i);
+```
+
+This is very verbose, but we don't want to use just `complex(r, i)` because we don't want functions confused with types.  We could use a simple pair structure instead of `to_complex(r, i)`, such as `{ r, i }`, `<r, i>`, or `[r, i]`, but these notations are already used for arrays, constraints, and row vectors respectively.  Or we could use an ad-hoc abbreviation like `cmplx(r, i)`, but that's hard to remember and we elsewhere try to avoid abbreviations where possible.
+
+### Accessor lvalues
+
+The real and imaginary components of a complex number may be accessed as follows.
+
+```
+complex z;
+...
+real r = z.real;
+real i = z.imag;
+```
+
+This follows the C++, Java, and Python conventions for accessing components of structures (including tuples).  It uses the C++ standard library's naming convention for the components.  We could instead use a function-like accessor notation, such as `z.real()`, but that feels less clear that it can be used an lvalue.  
+
+These accessors work as lvalues, so that the following will be allowed
+
+```
+complex z;
+...
+z.real = 5.2;
+```
+
+with the result being that `z` now has 5.2 as its real component---that is, it's a destructive operation.
+
+### Equality
+
+Equality is defined component-wise, so that `z1 == z2` for complex typed variables `z1` and `z2` is true if `z1.real == z2.real` and `z1.imag == z2.imag`.  
+
+```
+complex z1;
+complex z2;
+int a = (z1 == z2);
+```
+
+Note that this imports all of the usual difficulties of comparing floating-point numbers. Testing for equality of floating point and hence complex numbers should be avoided if possible.
+
+### Promotion and function arguments
+
+Currently in Stan, function arguments work like assignment.  You can use a variable for a function argument if and only if it could be assigned to a variable of the type of that argument.  
+
+Currently Stan allows promotion of `int` values to `real` values but not vice-versa, so that 
+
+```
+int a;
+real b = a;
+```
+
+is legal, but
+
+```
+real b;
+int c = b; // ILLEGAL
+```
+
+is not.
+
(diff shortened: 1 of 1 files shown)


SteveBronder:
Linking to the rendered version below

https://github.com/stan-dev/design-docs/blob/complex-design/designs/0030-complex-numbers.md

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+Lower and upper bounds could be construed as bounds on absolute value (or modulus) of a complex number, with
+```
+complex<lower = 1> z;
+```
+requiring `modulus(z) > 1`, where $\mod x + yi = \sqrt{x^2 + y^2}.$  Upper bounds and mixed upper and lower bounds work the same way.  The bounds are real valued.
+
+Lower and upper bound constraints can be applied to any of the complex types.
+
+This feature is included for completeness, but marked "optional" because it is not clear it will be useful as realistic complex constraints are typically multivariate.
```
WardBrian, line 71: This was not implemented

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+A complex number is constructed from a pair of real numbers, one representing the real component and one the imaginary component.  The proposal is to have a function do that work.
+
+```
+real r;
+real i;
+complex z = complex(r, i);
+```
+
+This overloads the type name in the same way as a C++ constructor.  `to_complex` seems too wordy and `cmplx` too cryptic.
```
WardBrian, line 84: In the end, `to_complex` was used

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Accessor lvalues
+
+The real and imaginary components of a complex number may be accessed as follows.
```
WardBrian, line 88: While the following section would still be cool, the current implementation differs, using

`get_real(z)` and `get_imag(z)`. I don't believe these can be used as lvalues

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+Stan programs are converted to C++ class definitions by the stanc3 transpiler.   There are two components in this compiled class that.  This section covers data input and parameter output.  User-facing I/O is handled by the interfaces.
+
+### Data input format
+
+The C++ model class uses an object of class `var_context` to input variables declared in the data block.  This object needs to be extended to read complex number objects.  To read a complex number, we must read a sequence of two real numbers to pass to the C++ `complex<double>` constructor.
+
+### Parameter output format
+
+Parameters are output as an array.  Complex numbers will be sequenced with their real components before their imaginary components.  The output names for a complex variable `z` will be `z.r` and `z.i` for the real and imaginary components.
```
WardBrian, line 265: These ended up as `z.real` and `z.imag`, not `r` and `i`

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+but that's going to be far too verbose.
+
+### R dump format input
+
+We can adopt the standard R dump format for complex number output, which separates the real and imaginary component with the 
+
+```
+z1 <- 3-1.5i
+z2 <- 2.3+4.7i
+```
```
WardBrian, line 298: This has not been implemented, but complex numbers can be read in with something like 

`z1 <- c(3, 1.5)`

similar to JSON

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+## Rationale and alternatives
+
+Most of these proposals are just applying the simplest possible thing that is mathematically and computationally coherent.  For example, it's natural to promote `real` to `complex` in mathematics, so we allow it in the language.  Similarly, it's natural to
```
WardBrian, line 302: The final sentence here is incomplete

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Container type alternative
+
+It might be tempting to try to decompose types as C++ does, for example using `matrix<real>` and `matrix<complex>` or even `complex matrix`, etc.  One obstacle to this is that we use `<...>` for constraints, as in `matrix<lower = 0>`.  And we want to start stacking constraints, as in `real<multiplier = 7, offset = 3.9><lower = 0>`.  But the real reason not to do this is that there's not much shared between the matrices other than constructors, getters, and setters.
+
+For example, complex numbers do *not* work with constraints.  It does not make sense to write `complex<lower = 0> z` or `complex<offset = 7>`.  
+
+### Full covariance
+
+The biggest issue is how much covariance do we want to support given that it's not yet fully supported for `int` and `real`.  For example, it'd be nice to allow assignment of `int[]` to `real[]` and of `real[]` to `complex[]`.
```
WardBrian, line 312: Since the writing of this design this has become possible

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Constructor alternatives
+
+We could use a simple pair structure instead of `complex(r, i)`, such as `{ r, i }`, `<r, i>`, or `[r, i]`, e.g.,
+```
+complex z = { r, i };
+```
+but these notations are already used for arrays, constraints, and row vectors respectively.
```
WardBrian, line 320: This should discuss the literal syntax added later, `3+4i`

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,355 @@
...
+
+For accessors, we could use method-like notation, such as `z.real()`, but Stan does not use this kind of object-oriented notation elsewhere and it is less clear that the result can be used an lvalue.
+
+### Offset and multiplier addition
+
+In addition to lower and upper bound constraints, it'd be possible to use complex offsets and multipliers, 
+```
+complex<offset = o, multiplier = m> z;
+```
+Here, the offset and multiplier could be complex scalars or real scalars or integers.
```
WardBrian, line 333: Also not implemented, and in an above paragraph it is stated this would not make sense

WardBrian (commented):
These comments are mainly to bring this document in line with what was implemented in https://github.com/stan-dev/stanc3/pull/940 before proceeding with the addition of complex matrices and vectors

designs/0030-complex-numbers.md
bob-carpenter: I think it's OK to leave it as an optional feature.  I'm not sure it's useful, so also feel free to cut.

designs/0030-complex-numbers.md
bob-carpenter: We should change this, then.

designs/0030-complex-numbers.md
bob-carpenter: Maybe when we get tuples?  So this should be updated with the . notation either removed or moved to a section marked "optional".

designs/0030-complex-numbers.md
bob-carpenter: At least it's only one line :-).  The doc should change.

designs/0030-complex-numbers.md
bob-carpenter: OK, then this also needs to be updated.

designs/0030-complex-numbers.md
bob-carpenter: delete.

designs/0030-complex-numbers.md
bob-carpenter: nice, so we can roll that into the proposal.

designs/0030-complex-numbers.md
bob-carpenter: Yes, please.

designs/0030-complex-numbers.md
bob-carpenter: Right.  We should just remove it.

bob-carpenter (commented):
@WardBrian :  this was just comments at this point.  Do you want to make the changes and then I can approve?  Or would you rather I make the changes?

designs/0030-complex-numbers.md
bob-carpenter: Removed.

designs/0030-complex-numbers.md
bob-carpenter: fixed.

designs/0030-complex-numbers.md
```diff
@@ -0,0 +1,328 @@
...
+real r = get_real(z);
+real i = get_imag(z);
+```
+
+These accessors do not work as lvalues. 
+
+
+### Equality
+
+Equality for scalars is defined component-wise, so that if `z1` and `z2` are are complex typed variables, then `z1 == z2` is true if `to_rea(z1) == to_real(z2)` and `to_real(z1) == to_real(z2)`.
```
WardBrian, line 98:
```suggestion
Equality for scalars is defined component-wise, so that if `z1` and `z2` are are complex typed variables, then `z1 == z2` is true if `get_real(z1) == get_real(z2)` and `get_imag(z1) == get_imag(z2)`.
```

WardBrian (approved):
One typo/inconsistency but otherwise this looks good to me. Thanks

bob-carpenter:
Thanks for catching that.  I made the update and pushed.

WardBrian (approved):

PR #43, opened 2021-09-27
Deprecation schedule for minor (breaking) language changes

WardBrian:
Rendered version: https://github.com/WardBrian/design-docs/blob/deprecation-schedule/designs/0031-language-deprecation-schedule.md

Prior discussion on forums (largely replicated in the doc): https://discourse.mc-stan.org/t/a-deprecation-schedule-for-stan/24228


designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+community review period during which we can recieve feedback and course-correct
+if we have made a poor choice.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Deprecated minor features may be removed from Stan during minor version changes.
+Here, 'minor change' is restricted to be changes which can be done automatically
+through the use of the stanc3 canonicalizer, or a user-driven find and replace
+operation. By their very nature, minor changes are primarily syntactic. This is
```
bob-carpenter, line 54: Is there an example of when they wouldn't be syntactic?

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+in contrast to 'major changes', which are by their nature semantic, and require
+human-in-the-loop reasoning to accommodate. Major changes, such the interface
+changes to the differential equation solvers, may be announced via deprecation
+warnings in minor versions, but will not be implemented/removed prior to a
+version 3.0.0.
+
+A minor feature must be deprecated for a minimum of 3 minor versions
+(approximately 1 year on the current release schedule) before removal. The exact
+timeline for a feature's removal will be decided during the development which
```
bob-carpenter, line 63: Does "during the development" mean that the deprecation schedule needs to accompany the pull request in which the deprecation is introduced?

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+through the use of the stanc3 canonicalizer, or a user-driven find and replace
+operation. By their very nature, minor changes are primarily syntactic. This is
+in contrast to 'major changes', which are by their nature semantic, and require
+human-in-the-loop reasoning to accommodate. Major changes, such the interface
+changes to the differential equation solvers, may be announced via deprecation
+warnings in minor versions, but will not be implemented/removed prior to a
+version 3.0.0.
+
+A minor feature must be deprecated for a minimum of 3 minor versions
+(approximately 1 year on the current release schedule) before removal. The exact
```
bob-carpenter, line 62: I think we should just make this unequivocal and then argue exceptions as usual on a case-by-case basis.  Otherwise, I fear every single minor feature could potentially lead to a discussion of how long its deprecation period should be.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
+- Feature Name: deprecation_schedule
+- Start Date: 2021-09-21
+- RFC PR: (leave this empty)
+- Stan Issue: (leave this empty)
+
+# Summary
```
bob-carpenter, line 6: The document needs a more descriptive title than "summary".

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+[summary]: #summary
+
+Introduce a process by which minor features can be removed from the Stan
+language following a 1 year (3 minor release) period of deprecation. 'Minor
```
bob-carpenter, line 10: Is there a reason this only discusses the language and not the rest of Stan?  Could we make the policy universal across repos?  The only concern I have is that the math lib has never respected backward compatibility release over release, at least at the detailed level.  (I think what we need to do is clearly indicate a non-changing public API for the math lib that won't change.)

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+timeline for a feature's removal will be decided during the development which
+would mark it as deprecated.
+
+A minor feature is deprecated and a candidate for future removal when:
+1. stanc emits a compile-time warning with the following 3 parts:
+    1. A description of the deprecated feature, such as "Use of <- for assignment
+       is deprecated".
+    2. An _action_ the user should take, such as "Use = instead". If stanc3 can
+       automate this for them, we should call that out, leading to a full
+       message like "Use = instead or run stanc3 --print-canonical to update
```
bob-carpenter, line 72: I think the stanc3 command needs a path to the file name.  And it should mention what's being replaced.  It's also not clear what the boundary of the command is, so I'd put it last, like this:

```
Use = instead of <- for assignment;  this can be automatically fixed by
running: stanc3 --print-canonical <stan-program-file>
```

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+       print an updated version of your code."
+    3. A timeline which includes a date or version after which this feature is
+       slated for removal, for example "This deprecation will expire in Stan
+       version 2.31.0 (expected Sept. 2022)".
+2. It has been added to the [documentation on
+   deprecation](https://mc-stan.org/docs/reference-manual/deprecated-features-appendix.html)
+   with the same 3 pieces of information, and additional information as needed
+   (such as any of: reasoning for removal, advantages of new behavior, ongoing
```
bob-carpenter, line 80: such as any of: -> e.g.,
reasoning -> reason
etc). -> etc.).

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+   related changes, etc).
+3. Examples in the documentation have been updated to remove any examples of the
+   deprecated syntax, except for those referring to its deprecation. This change
+   may lag behind bullets 1 and 2, but must occur before the removal.
```
bob-carpenter, line 84: The main documentation should remove a feature and include it in the deprecation section as soon as the feature is deprecated.   Otherwise, users will still think it's fine.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+
+Our goal is to only deprecate a feature when doing so provides a tangible
+benefit to users, such as improving error messaging or allowing new features to
+be added to the language while keeping a low maintenance burden. Changes should
```
bob-carpenter, line 88: This is confused.  I think it needs to provide a tangible benefit to either users _or_ developers.  The second clause is about maintenance, which isn't a user concern.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+never take users (even those with many or large models) long to adjust to, and
+as few such changes should be made in each version as is reasonable.
```
bob-carpenter, line 90: I don't see the point of spreading deprecations over multiple releases if they're ready.  We should only make deprecations that are reasonable, but given that, any number of them should be possible.  So I'd just drop line 90.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Developers
+From a developer standpoint, this policy is rather straightforward. When a
+change is made, it should identify any deprecation candidates: new keywords that
+must be implemented (thus 'deprecating' the use of that word as a identifier),
+new syntax which supersedes prior syntax, (such as `=` or `array[]`), or a
+similar **syntax-level** change. If supporting both the new and old way of doing
```
bob-carpenter, line 101: [optional]
I'd recommend italics (single *) rather than bold for emphasis.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+things requires a 'hack' or difficult to maintain code, this is a prime
+candidate for this deprecation-removal process. If a change meaningfully changes
+the semantics of the language, it is not.
+
+>Current examples include the use of 'partial keywords' like offset,
```
bob-carpenter, line 106: Why is this block quoted?

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Developers
+From a developer standpoint, this policy is rather straightforward. When a
+change is made, it should identify any deprecation candidates: new keywords that
+must be implemented (thus 'deprecating' the use of that word as a identifier),
+new syntax which supersedes prior syntax, (such as `=` or `array[]`), or a
+similar **syntax-level** change. If supporting both the new and old way of doing
+things requires a 'hack' or difficult to maintain code, this is a prime
+candidate for this deprecation-removal process. If a change meaningfully changes
+the semantics of the language, it is not.
```
bob-carpenter, line 104: I think we can have arbitrarily compatibility-breaking deprecations.   So maybe rewrite this doc to explicitly introduce major deprecations (anything that's not minor).  We don't want to say it's impossible, just that it's not going to auto-deprecate in three releases.  We'll have to wait for major changes.  

Do we want to wait at least a year after deprecation before changing major version in the same way?

designs/0031-deprecation-schedule.md
bob-carpenter, line 106: The word "offset" should be quoted and "partial keywords" should not be.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+>which are not reserved words but act like them in some contexts, or multiple
+>syntaxes which lead to parsing difficulties, like the new array syntax. We would
+>like to not only improve the error messaging and implementation of these current
+>features, but also avoid introducing new issues of this variety in the long term.
+
+>Current non-examples include the ODE solver interface changes, which requires
+>changes to the user-defined function defining the ODE.
+
+Once it has been decided to deprecate a feature, the lexer, parser, or
+typechecker (whichever is most apropriate) should be updated to emit a
```
bob-carpenter, line 116: typechecker -> type checker

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+compile-time warning notifying the user. This must contain the above 3 criteria:
```
bob-carpenter, line 117: I'd strongly recommend using "this" less and repeating nouns more.  So rather than "This" I'd say "The warning" here.  

But the real issue is that these aren't criteria.  Small numbers used as adjectives should be spelled out, so it's "three" not "3" here.  So overall, I'd go with

```
The warning message must specify three pieces of information, (a) what
is deprecated, (b) how to adjust code to replace the deprecated feature, and (c) the
version in which the deprecated feature will be removed.
```

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+1. What is deprecated
+2. How to adjust your code
+3. When you need to adjust it by.
+
+The final bullet point should be decided during the review process with 3
+versions as a minimum. Many changes, such as reserving a new keyword, will not
+require longer periods, but syntax changes which the user may need time to
+adjust to could require longer time frames.
```
bob-carpenter, line 125: I think a year is more than enough for any kind of change.  So let's just cast 3 versions in stone.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+
+This information must also be added to the documentation in the accompanying PR
+to the change.
```
bob-carpenter, line 128: I think this paragraph is redundant.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+
+Finally, developers may wish to weigh in on how the implementation might look
```
bob-carpenter, line 130: We're not talking about what the devs want, but what they're allowed to do, so:

"wish to" -> ""

But don't we just want to say that they must discuss this as part of deprecating a feature?  Or are you talking about the OCaml code instead of the Stan code?  (It always helps to be as specific as possible in specs.)

Also, I think we can give them the latitude to do this under any circumstances, so:

", if they can foresee..." -> "."

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+after the removal of the deprecated feature as part of the text of the pull
+request, if they can foresee code improvements which can occur at that time.
+
+## Code reviewers
+From a code reviewer standpoint, it is important to ask the following questions:
```
bob-carpenter, line 135: : -> .

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+1. Is this deprecation tied to a useful feature or noticeable improvement to the
+   underyling code (not just 'for the sake of it')?
```
bob-carpenter, line 137: (not just ...) -> ""

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+2. Does the new warning meet the criteria above and explain itself clearly?
```
bob-carpenter, line 138: The criterion listed above is that three descriptive items are included in a warning message.  One might argue there are three criteria, one for the inclusion of each part of the message, but I wouldn't.  Honestly, I'd just move away from "criterion" altogether.  You might want to call these "review criteria" if you really want to get the word in :-).

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+3. Is there a simple (preferably automated) way for users to accommodate this
+   change, e.g. through the `print-canonical` flag to stanc3 or a
```
bob-carpenter, line 140: e.g. -> e.g.,  [e.g. and i.e. are always followed by commas]\

I'd also put the whole example in parens, so it'd be "(e.g., [text])?"

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+   find-and-replace feature of their text editor?
+4. Is the removal timeline appropriate for this change?
```
bob-carpenter, line 142: I think we should just keep things simple and always have a 3-release timeline for deprecations.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Releases
+Finally, during the release cycle an additional bullet point should be added to
```
bob-carpenter, line 145: You don't need the adjective "additional" when the verb is "added".

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+the release checklist to ensure that the deprecations slated for removal in the
+coming version have been properly handled. To accommodate this, the documentation
```
bob-carpenter, line 147: Is this something that just needs to go in a template somewhere?  If so, say where if you can.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+page on deprecations should be sorted/categorized by removal version.
+
+Additionally, the release notes should contain a section on **Deprecations** and
```
bob-carpenter, line 150: So the last "Finally" wasn't really a finally.  Overall, I'd just stay away from qualifying sentential adverbs like this.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+**Expired Deprecations** (alt. **Removals**) for each version. This should
```
bob-carpenter, line 151: Let's just go with "Removed features" or "Breaking changes".  I think "expired deprecations" has one too many negative items and also isn't in the user's language.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+duplicate the information on the what and how of the change from the error
+message and documentation.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Drawbacks to this are rather obvious - backwards compatibility is certainly a
```
bob-carpenter, line 159: "Drawbacks to this are rather obvious --" -> ""

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+virtue, and any breaking change will lead to some headaches for a portion of the
+user base, no matter how small. In the limit, we certainly want to avoid any
```
bob-carpenter, line 161: I don't see what limit we're talking about here.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+large enough change that users prefer not to update (a core reason these changes
+should be tied to improvements in the language, not just for their own sake) and
```
bob-carpenter, line 163: ", not just for their own sake" -> ""

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+any changes which require dedicated rewriting or rethinking of models. This
```
bob-carpenter, line 164: I think this is just rehashing the major vs. minor deprecation issue and can be dropped.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+desire motivated the definition of 'minor change' above and the restriction to
+primarily automatable syntax changes.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This design allows for us to care for the long term health of the Stan language
+and codebase while clearly communicating our intent to users. Small changes can
+be made to make the development of new features easier without "sneaking up on"
```
bob-carpenter, line 173: Choose either `'` or `"` for quotes, but don't mix.  But the real issue is that you shouldn't be quoting common idioms.  Either write with idioms or don't, but don't quote them.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+should be tied to improvements in the language, not just for their own sake) and
+any changes which require dedicated rewriting or rethinking of models. This
+desire motivated the definition of 'minor change' above and the restriction to
+primarily automatable syntax changes.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+This design allows for us to care for the long term health of the Stan language
+and codebase while clearly communicating our intent to users. Small changes can
```
bob-carpenter, line 172: Make this active.  Deprecating features empowers developers to improve the language, without introducing breaking changes that sneak up on users.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+be made to make the development of new features easier without "sneaking up on"
+anyone or breaking things willy-nilly.
```
bob-carpenter, line 174: "or breaking things willy-nilly" -> ""

We don't need to remind devs not to break things willy nilly!

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+1. The first is to stick strictly to the semantic versioning standards, and make
+   a strong push toward a Stan version 3.0 which makes backwards incompatible
+   changes.
+
+   Unless we attempt to systematically release major versions with some
+   regularity, we still end up with the same problem of these tiny issues
+   accumulating in the meantime. Additionally, saving all such minor changes for
+   a major version bump dramatically increases the friction for updating. While
+   some may find it preferable to have to update their models only once, the
+   small requirement of (at most) 1-2 changes per version is likely preferable
```
bob-carpenter, line 186: I wouldn't legislate the number of deprecations per release.  For one things, that would require a way of counting them!

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+  policy](https://docs.oracle.com/javase/9/core/enhanced-deprecation1.htm#JSCOR-GUID-23B13A9E-2727-42DC-B03A-E374B3C4CE96)
+  in Java 9 (which is technically a minor version, 1.9, for the JDK). The main
+  motivation for clarification was the response of users being unsure if a
+  deprecated feature would ever actually be removed. There was a
+  [study](https://doi.org/10.1007/S10664-019-09713-W) of this change published
+  in 2019. It contains the following relevant passage and some interesting
+  discussion of user reactions to a deprecation warning.
+  > By not removing deprecated features from an API after a transition period
+  > has passed, API producers and the Java JDK developers themselves have
+  > cheapened the meaning of deprecation.
```
bob-carpenter, line 232: [comment]
Uh oh.  We've also cheapened the term!

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+- https://semver.org/ - the description of proper 'semantic versioning'. In the
+  strictest interpretation, Stan already deviates from this scheme with some
+  regularity, but it is a common understanding of the major.minor.patch
+  versioning.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Above I set the term of deprecations at 3 minor versions/1 year. This may be
```
bob-carpenter, line 241: Let's just fix to 3 minor versions or 1 major version like Python.  We don't want every deprecation to involve a debate.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+  too long for some features and too short for others, and is worth discussing.
+- We have many existing deprecations which predate this proposed policy. How
```
bob-carpenter, line 243: Please move this up into the spec itself.  It's a reasonable proposal.

designs/0031-deprecation-schedule.md
```diff
@@ -0,0 +1,254 @@
...
+  they should be handled or removed is not explicitly covered; the easiest path
+  is to consider the first version after this policy is enacted to be their
+  'deprecation version', and remove them 3 minor versions later.
+- This is independent of questions about larger features like the ODE solver
+  changes and the potential of a Stan version 3.0, but these should be
+  considered when thinking of the long term health of the project as well. How
+  to handle the 'major' or semantic changes to the language is not resolved by
+  this proposal.
```
bob-carpenter, line 251: I think it is.  Anything that's not a minor deprecation as defined here will require a major version bump.  Is there anything else to say?

bob-carpenter (changes requested):
The main contentful comment here is that we should commit to 3 minor releases or 1 major release for minor deprecations.  We should also say anything that's not a minor deprecation as defined here will require a major version bump.

I hope you appreciate the more nitpicky comments rather than finding them annoying.

designs/0031-deprecation-schedule.md
WardBrian: Even Python includes the line

> It's fine to wait more than two releases.

when describing their policy. I agree that a priori it seems strange to think of when something would be fine to remove, but wanting more than 3 versions.

designs/0031-deprecation-schedule.md
WardBrian: I was just going off the template in this repo. It seems like other docs all use the same format? E.g. https://github.com/stan-dev/design-docs/blob/master/designs/0003-opencl_kernel_generator.md#summary

designs/0031-deprecation-schedule.md
WardBrian: No, I think 'primarily' was an artifact of me writing imprecisely

designs/0031-deprecation-schedule.md
WardBrian: My belief is yes, because that pull request will have to write the warning message, and the warning message should always contain a deadline.

designs/0031-deprecation-schedule.md
bob-carpenter: Another way of looking at this is the whole spec is aimed at preventing the "let's just wait" approach to deprecation.

Let me ask this another way.  When would we want to wait more than 3 releases?

designs/0031-deprecation-schedule.md
WardBrian: I focused it on the language mainly because I think that is where the primary examples of where we currently have issues with this arise. I think that pystan, cmdstanr, etc will all follow the conventions of their own ecosystems, and recently the Math repo I think formally stated they're not making backwards compatibility guarantees. 

I'm happy to extend it to other parts of the ecosystem, but I'd need help imagining what that looks like for things outside the language.

designs/0031-deprecation-schedule.md
WardBrian: I agree, I mainly mean that updating each example in all other parts of the dics may not happen right away. We could require this if we wanted, and it may even serve as a reasonable barrier to spurious deprecations. I didn't want to make it blocking like the first two bullets, mainly because it is a much larger task

designs/0031-deprecation-schedule.md
bob-carpenter: I'd prefer a title, but will accept "Summary" given it's what the ones that have been merged do (other than CmdStanPy).

designs/0031-deprecation-schedule.md
bob-carpenter: If you want to keep it to the language, I think the title should change from `deprecation_schedule` to make that clear.  I don't know where the convention came from of naming these like file names without spaces.

designs/0031-deprecation-schedule.md
bob-carpenter: I'd like to keep the docs up to date, so we should require the doc fix when the deprecations go in.  I'm always happy to help revising doc.

designs/0031-deprecation-schedule.md
WardBrian: We will be counting them when we produce release notes, but I think you're correct its hard to codify this fully.

designs/0031-deprecation-schedule.md
WardBrian: I think it's all about expectations. My instinct is major versions would not require the same timeline, but that is partially because I expect a new major version would be announced pretty far in advance anyway.

designs/0031-deprecation-schedule.md
WardBrian: I meant the OCaml, meaning something like "Once x is reserved as a keyword formally, we can change the parser to behave like y and simplify its behavior".

designs/0031-deprecation-schedule.md
WardBrian: I don't know where this template lives exactly, @rok-cesnovar seems to be the steward of it

designs/0031-deprecation-schedule.md
WardBrian: It's a really interesting article, in part because it seems like nearly _everybody_ cheapens the term

designs/0031-deprecation-schedule.md
WardBrian: That's fair. I've updated it other places, but it is in the "Unresolved questions" portion in part because I wonder if 3 is the best number for a hard and fast rule or if we would prefer 2 minor versions (so you have the deprecating version, a buffer version, and the removal version).

designs/0031-deprecation-schedule.md
WardBrian: You're right, I more mean that the question of "when will we do stan 3 and what will be in it" is orthogonal to this.

designs/0031-deprecation-schedule.md
rok-cesnovar: I currently just take the list from the previous release cycle (for example https://github.com/stan-dev/cmdstan/issues/1037) and add stuff there. It would probably be best to save it somewhere more stable, but I dont have any ideas as to where the best place for that would be.

WardBrian: We could make a stan-dev/administration repo with a markdown file for that? I'm sure we will find other uses for such a repository as well

rok-cesnovar: Yeah. Though that reminded me that we have https://github.com/stan-dev/ci-scripts

The name CI might be a bit misleading but maybe?

WardBrian: The release folder there's not a terrible place to keep it IMO if we want to just reuse the existing repo. Just a checklist.md we can keep updated and copy/paste as needed

rok-cesnovar: There it is: https://github.com/stan-dev/ci-scripts/blob/master/release-scripts/checklist.md

WardBrian: Thanks Rok!

WardBrian:
I'd like to ask for any more feedback on this. We will be discussing it during Thursday's language meeting and then merging unless there are objections

bob-carpenter (approved):
Looks good.  Thanks!

PR #44, opened 2022-01-18
Refactor inference algorithm outputs, replace single CSV file with multiple, cleanly structured outputs

mitzimorris:
The Stan CSV output file format is problematic.   To address this problem we need to move to multiple output files, where each file contains a single kind of information and whose format matches the structure of the contents.
This will allow for more and better downstream processing of the inference engine outputs.

Linking to the rendered version below

https://github.com/stan-dev/design-docs/blob/cmdstan-output-formats/designs/0032-stan-output-formats.md


betanalpha:
The link to the rendered version isn’t working so I can’t comment on the details, but I’m worried about the definition of “downstream processing”.  CmdStan was designed for command line use and the current output design was validated against standard command line tools.  The more programmatic processing needed for wrapper interferes was never part of the design nor promised by any part of the API.
One immediate challenge with multiple output files is that they clutter directories, making it not only more annoying to navigate around from the command line but also harder to keep the outputs of a run consistently organized.  

Without a dedicated file format there’s no great way to keep information organized in one place but compartmentalized.

mitzimorris:
this is ready for review.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The CSV format itself is not fully standardized.
+The original intent of the format is to represent a table's worth of data in plain text.
+Each row of the file contains one row of table data, and all rows have the same number of fields.
+Over time, conventions have been adopted, notably use of row headers for column labels and comment rows.
+
+The Stan CSV file, however, contains far more information than a single table's worth of data.
+To cram all the information from a CmdStan run we 
+abuse the CSV format by using blocks of comment lines to record additional information.
+The current Stan API requires the information in a comment line be assembled and flattened into a text string.
+Even when we can use a fast CSV parser to extract the CSV data, we still need to build ad-hoc parsers for the comment blocks
```
WardBrian, line 26: I would add a note here about the relative inefficiency of CSV compared to a binary format, and the lack of types (e.g. int/real distinctions).

On the other hand, it's worth noting that this plain-text format has the advantage of being very 'accessible' in the sense that no library support is really needed, just a lot of post-processing due to our additional information we add in.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The [Stan CSV file format](https://mc-stan.org/docs/cmdstan-guide/stan_csv.html) is limited and limiting for several reasons.
+
+The CSV format itself is not fully standardized.
+The original intent of the format is to represent a table's worth of data in plain text.
+Each row of the file contains one row of table data, and all rows have the same number of fields.
+Over time, conventions have been adopted, notably use of row headers for column labels and comment rows.
```
bob-carpenter, line 20: I'm pretty sure most of this came in from day 1.  The original intent was to have it be a matrix of draws in CSV format if you ignored comments.  I have never liked the reading of the comments by CmdStan.

designs/0032-cmdstan-output-formats.md
bob-carpenter: The original intent was just to have the matrix of draws data to read in pure CSV format (with comments, but not comments with content to read).

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+The Stan CSV file, however, contains far more information than a single table's worth of data.
+To cram all the information from a CmdStan run we 
+abuse the CSV format by using blocks of comment lines to record additional information.
+The current Stan API requires the information in a comment line be assembled and flattened into a text string.
+Even when we can use a fast CSV parser to extract the CSV data, we still need to build ad-hoc parsers for the comment blocks
+to re-assemble the flattened information.
+
+The sample methods produce the largest amount of output data and these are the most difficult to process quickly and efficiently.
```
bob-carpenter, line 29: Doesn't ADVI produce exactly the same form of output (modulo the mass matrix and step size)?  Also, @andrewgelman has been lobbying for L-BFGS to return draws from a Laplace approximation, which would put it on the same footing.  No warmup in those cases, and usually the draws are independent so no need for long chains.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The first part of a CmdStan CSV file always a series of comment rows which contain configuration and version information,
+and this comment header continues to accumulate new bits of information about program version and compilation options.
+The NUTS-HMC methods output a series of comment lines which contain the stepsize and metric.
+The sample methods also record timing information at the end.
+
+In order to monitor the progress of CmdStan's inference algorithms,
+the outputs from CmdStan must be available to downstream readers on a streaming basis.
```
bob-carpenter, line 36: I think this motivation needs to go at the top.  Is this really a requirement?  If so, I think you may need to define a different interface rather than trying to share one that streams to file.  

What exactly do you expect to read downstream?

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+To facilitate monitoring different aspects of the algorithm, a more granular set of outputs,
```
bob-carpenter, line 37: The more granular set of outputs needs to be at least sketched here, or make a reference to later in the doc.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+instead of a single CSV file would make this task easier.
+Another consideration is the use of multi-chain methods; currently multi-chain NUTS-HMC and in the future,
+multi-chain Pathfinder.  When running any kind of ensemble method, we need to record the process/chain_id
+and make this available 
+
+Therefore we need to re-implement the set of output methods used by CmdStan and the core Stan services
```
bob-carpenter, line 43: I'm confused again.  "Re-implement" implies the interface isn't changing.  Do you mean change what is being output, too?

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+to capture more information in a more structured way.
+At the same time, we need to continue to support the current Stan CSV format in order to support existing downstream processing packages.
```
bob-carpenter, line 45: I think you should move this up front as a top-level goal of backward compatibility.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For a given run of an inference algorithm given a model and data,
```
bob-carpenter, line 51: given a model and data -> with a specific model and data set

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+2. Inference engine outputs: various diagnostics.
+For the NUTS-HMC sampler, the current outputs are the stepsize and metric, and the per-iteration sampler variables (state).
+The optimization and variational algorithms also report on their respective iterations.
+
+3. Interface-level outputs:
+   + model name, compile options, Stan compiler and (Cmd)Stan version and compile options.
+   + inference algorithm, user-specified configuration options, and default config.
+   + input and output data descriptors.
+   + process-level information:  chain, iteration.
+   + timing information: timestamp information reported by processing events.
```
bob-carpenter, line 68: Do we really want to save timing info?

Timestamping (when it was run) is more relevant.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+There are two general kinds of information structure: tabular and hierarchical.
+[CSV files](https://en.wikipedia.org/wiki/Comma-separated_values) are tabular,
+as are R and Python dataframes and database tables.
+JSON is a [structured data formats](https://en.wikipedia.org/wiki/Data_exchange#Data_exchange_languages)
+which can represent hierarchical structures and arrays.
+
```
bob-carpenter, line 75: JSON can also represent tabular data.  It's just not as compact as CSV or binary formats.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+### Output mechanism:  callback writers
+
+Outputs are handled by objects of base class `stan::callbacks::writer`
+which is constructed from an output stream.
+Through function call operator overloading, the callback writer class
+methods provide the appropriate formatting for different kinds of data.
+When the user sends a request to the interface to do some kind of inference given a model and data,
+the interfaces first instantiate the Stan model object and the writer callbacks,
+then pass these in to `stan::services` functions which invoke the appropriate algorithm.
```
bob-carpenter, line 88: Isn't there one service function per algorithm?  This makes it sound like there's some kind of dispatch.  Or are we now up at the cmdstan or cmdstanpy level?

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The use of writer callbacks keeps the inference algorithms output-format agnostic.
+The current `stan::services` layer utilities are completely Stan CSV format-centric and will need to be refactored.
```
bob-carpenter, line 90: Can you say in what way these are CSV format-centric?  I thought we just output draws as vectors to the callbacks.  That doesn't imply any kind of output at all, much less CSV.  This is evidenced by RStan and PyStan using them with no CSV files in sight.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+
+### Stan program outputs
+
+The Stan model class `write_array` method returns a vector of doubles for all
+parameter, transformed parameter, and generated quantities (non-local) variables,
+in order of declaration in the Stan program.
+
+The model class methods `constrained_param_names`, `unconstrained_param_names`, `get_constrained_sizedtypes`, `get_unconstrained_sizedtypes`
+provide name, type, and shape information.
```
bob-carpenter, line 100: There's a distinction between the model class methods that are static (don't require a model instance) and those that depend on sizes.  The sizes are only available once a model is instantiated with data, whereas the other functions can be static class functions and don't have to be methods (i.e., member functions).

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The reason that the model reports on both the constrained and unconstrained parameters is the Stan parameter types
+simplex, correlation, covariance, and Cholesky factors have different shapes on the constrained, unconstrained scale.
+For example, a simplex parameter is a vector whose elements sum to one on the constrained scale which means that it has $N-1$ free parameters
```
bob-carpenter, line 103: You might want to say $N$ vector (though I don't think math renders here) to give some context as to the N - 1 used later.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+when computing on the unconstrained scale.
+The Stan model parameter inputs and outputs are on the *constrained* scale, therefore the number of parameter values reported by the `write_array` method
+is reported by the `get_constrained_sizedtypes` method.
+The sampler computes on the *unconstrained* scale, therefore the size of the metric (inverse mass matrix) is reported by the `get_unconstrained_sizedtypes` method.
+(Note: A further complication for optimization is that the optimization algorithm computes on the unconstrained scale, which means that the Hessian of the covariance matrix is on the unconstrained scale.)
```
bob-carpenter, line 108: I think you mean the inverse Hessian approximation to the covariance matrix.  But that's not even a part of CmdStan.  Are you proposing to add that as part of this feature?

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+
+### Inference algorithm outputs
+
+The outputs from the inference algorithm necessarily differ according to the method, and to a lesser degree, according to the developers.
```
bob-carpenter, line 113: I'd drop the editorial about devs.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The NUTS-HMC sampler does full Bayesian inference by assembling a set of draws from the posterior distribution using an HMC sampler
+with tuning parameters step-size and metric.
+The `stan::services::util::mcmc_writer` class is constructed with a callback writer and provides methods
+`write_sample_names`, `write_sample_params`, `write_adapt_finish`, `write_timing`.
+
+- The `write_sample_names` and `write_sample_params` produce the CSV header row and data rows, respectively.
```
bob-carpenter, line 119: They actually just produce a vector of strings.  Using them for CSV headers is just the decision that CmdStan made.  I think it's important to keep the levels separated here.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The initial columns consist of sampler state and diagnostic information from the `stan::mcmc::sample` object,
+(called `sample_params`), the remaining columns are the Stan model variable names and values returned by
```
bob-carpenter, line 121: The name `sample_params` is terrible!  Are you sure it's not even `sampler_params`?

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+the `constrained_param_names` and `write_array` methods.
```
bob-carpenter, line 122: Is `constrained_param_names` a method or static class function?  If I recall, there are two forms, one that prints out just the names, and that's the same for constrained and unconstrained, and also one that prints out the whole thing as strings with indexes.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+- The `write_adapt_finish` and `write_timing` produce comment blocks in the CSV file.
+Adaptation information consists of the stepsize and metric.  This comment block follows
+the CSV header row, either immediately, or following the warmup data rows.
+The timing information comments are written out at the end.
+
+
+### Interface-level outputs
+
+The CmdStan interface dumps all configuration information into the Stan CSV file as a series of initial comment lines.
```
bob-carpenter, line 132: Does that include initializations or is that not considered part of config?

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+Tabular data:
+
+- The Stan model variables,
+i.e., the `constrained_param_names` method generates the output column names
+and the `write_array` method generates the output data.
+
+- The inference method variables (`sampler_params` for HMC, tbd for optimization, variational inference).
+
+- The HMC metric (?).  The HMC metric is either a matrix diagonal vector or a dense matrix.
```
bob-carpenter, line 153: The metric is a positive-definite matrix, which can be either diagonal or dense.  If it's diagonal, then we only need to store the vector corresponding to the diagonal entries.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+It is a dense, symmetric positive definite matrix.
+The sampler algorithms allow the user to specify the initial mass matrix (in JSON or rdump formats).
+
+- The HMC stepsize is a scalar.  It can trivially be output in CSV or JSON format.
```
bob-carpenter, line 157: It's a positive real number, to be specific.

How do we output one number in CSV or JSON format?  Is that just the number in ASCII?

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+
+Hierarchical data:
+
+- Stan model variables schema: block, name, type, shape, size, output column name(s)
+
+- Information about the inference engine run:  all information currently jammed into the Stan CSV file header
```
bob-carpenter, line 164: You can drop things like "jammed" here.  We get the point.  Nobody likes the current output format, so you don't have to sell getting rid of it.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+and the closing timing comments.
+
+The user interfaces will manage creating and naming the suite of output files.
```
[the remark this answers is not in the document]
bob-carpenter: Yes, please.

bob-carpenter (changes requested):
The main thrust of these comments is that the design doc needs to be clarified as to scope and filled out with details from a client perspective.

I'll leave it to @WardBrian to approve.

designs/0032-cmdstan-output-formats.md
WardBrian: I think that the current set of callbacks and their usage is CSV-centric in at least a couple ways (which I agree should be made explicit in this design doc)

1. The information is all flattened before being passed to these callbacks, so any format is expected to be csv-like in that it only accepts flat rows of data. This is in contrast to something like Arrow/Protobuf/NetCDF/whatever binary format, which generally support things like nested arrays as a native part of their data model. 
2. Anything that is not a draw, e.g. the adaptation information, is passed as pre-formatted string, since the current assumption is that this will just have a `#` prepended to make it a comment in a CSV

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The CmdStan argument parser uses indentation to indicate nested argument structure.
+Following the argument parser are several more comment lines of key-value pairs of information
+about the software version and compiler options used to compile the Stan model executable.
+
+
+## Functional specification
+
+We propose to break down the information in the current Stan CSV file into several files
+according to information source, whose format matches the information structure of the contents.
+The default file format for tabular data will be CSV.
```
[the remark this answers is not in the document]
WardBrian: I think that this proposal would be to change the internals of the callback and output structure, but (at least initially) have the end result look mostly the same, e.g., a CSV file. I'm guessing for backwards compat. that they CSV will need to keep the header comment, but maybe this information is _also_ available in a json for redudancy/ease when available?

The refactor would allow multiple output options, such as Arrow/Parquet, but I believe that the user would need to opt-in. That's just based on my reading of it, which is why I think clarification in the text would be good

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+the outputs of interest can be categorized in terms of information source and content
+and information structure.
+
+There are three sources of information:
+
+1. Stan program outputs: estimates for all model parameters, transformed parameters, and generated quantities.
+
+2. Inference engine outputs: various diagnostics.
+For the NUTS-HMC sampler, the current outputs are the stepsize and metric, and the per-iteration sampler variables (state).
+The optimization and variational algorithms also report on their respective iterations.
```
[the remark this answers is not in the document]
wrongu: Yes – the natural space to consider outputs for variational inference, or mixtures-of-variational-infererence – is the parameters of the approximation. Drawing samples and importance-weighting could be a second post-processing step.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+The Stan CSV file, however, contains far more information than a single table's worth of data.
+To cram all the information from a CmdStan run we 
+abuse the CSV format by using blocks of comment lines to record additional information.
+The current Stan API requires the information in a comment line be assembled and flattened into a text string.
+Even when we can use a fast CSV parser to extract the CSV data, we still need to build ad-hoc parsers for the comment blocks
+to re-assemble the flattened information.
+
+The sample methods produce the largest amount of output data and these are the most difficult to process quickly and efficiently.
+The first part of a CmdStan CSV file always a series of comment rows which contain configuration and version information,
```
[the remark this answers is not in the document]
mitzimorris: perhaps I should talk about the problem in more general terms?
I agree that this isn't really about CmdStan - to address this problem we're going to need to add more stuff to core Stan.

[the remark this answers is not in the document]
mitzimorris: revised per comments

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
+- Feature Name: cmdstan-output-formats
+- Start Date: 2022-01-11
+- RFC PR:
+- Stan Issue:
+
+# Summary
+[summary]: #summary
+
+Provide alternatives to the [Stan CSV file format](https://mc-stan.org/docs/cmdstan-guide/stan_csv.html)
```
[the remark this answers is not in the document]
mitzimorris: Alternatives - updated title accordingly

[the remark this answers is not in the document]
mitzimorris: agree about shellstan.  changed title and discussion to make it clear this isn't about CmdStan - the goal is to provide better support to other interfaces (like CmdStanPy and CmdStanR and whatever comes after them).

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The sample methods produce the largest amount of output data and these are the most difficult to process quickly and efficiently.
+The first part of a CmdStan CSV file always a series of comment rows which contain configuration and version information,
+and this comment header continues to accumulate new bits of information about program version and compilation options.
+The NUTS-HMC methods output a series of comment lines which contain the stepsize and metric.
+The sample methods also record timing information at the end.
+
+In order to monitor the progress of CmdStan's inference algorithms,
+the outputs from CmdStan must be available to downstream readers on a streaming basis.
+To facilitate monitoring different aspects of the algorithm, a more granular set of outputs,
+instead of a single CSV file would make this task easier.
```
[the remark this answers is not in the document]
mitzimorris: agreed.  revised per comments.

designs/0032-cmdstan-output-formats.md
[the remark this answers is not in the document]
mitzimorris: noted.

designs/0032-cmdstan-output-formats.md
mitzimorris: moved to top.
will drill down on downstream processing.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+The callback writer is a good design but the `mcmc_writer` class needs to be refactored.
+Multiple output files require one callback writer per file.
+These writers are managed by a composite object which manages a set of callback writer objects
+and defines the appropriate set of output methods needed to dispatch information
+to the correct writer.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This refactoring cleans up aspects of the Stan services layer at the cost of making the interfaces do more work.
```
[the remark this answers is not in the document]
mitzimorris: you're right - changed this accordingly.

designs/0032-cmdstan-output-formats.md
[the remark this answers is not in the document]
mitzimorris: updated motivation - Brian is correct - for backwards compatibility, we will maintain the current Stan CSV files.   revised accordingly.

designs/0032-cmdstan-output-formats.md
mitzimorris: sketched.

designs/0032-cmdstan-output-formats.md
mitzimorris: moved.

designs/0032-cmdstan-output-formats.md
mitzimorris: revised to make it clear - leave what's there, there.  add new stuff.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For a given run of an inference algorithm given a model and data,
+the outputs of interest can be categorized in terms of information source and content
+and information structure.
+
+There are three sources of information:
+
+1. Stan program outputs: estimates for all model parameters, transformed parameters, and generated quantities.
```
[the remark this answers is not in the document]
mitzimorris: thanks - revised accordingly

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+2. Inference engine outputs: various diagnostics.
+For the NUTS-HMC sampler, the current outputs are the stepsize and metric, and the per-iteration sampler variables (state).
+The optimization and variational algorithms also report on their respective iterations.
+
+3. Interface-level outputs:
```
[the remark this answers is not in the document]
mitzimorris: the code goes into `stan::services` layer.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+   + model name, compile options, Stan compiler and (Cmd)Stan version and compile options.
+   + inference algorithm, user-specified configuration options, and default config.
+   + input and output data descriptors.
+   + process-level information:  chain, iteration.
```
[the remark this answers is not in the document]
mitzimorris: I like the long form idea alot - yes.

designs/0032-cmdstan-output-formats.md
mitzimorris: no, we don't want timing info - as you say - timestamping much more useful.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+1. Stan program outputs: estimates for all model parameters, transformed parameters, and generated quantities.
+
+2. Inference engine outputs: various diagnostics.
+For the NUTS-HMC sampler, the current outputs are the stepsize and metric, and the per-iteration sampler variables (state).
+The optimization and variational algorithms also report on their respective iterations.
+
+3. Interface-level outputs:
+   + model name, compile options, Stan compiler and (Cmd)Stan version and compile options.
+   + inference algorithm, user-specified configuration options, and default config.
+   + input and output data descriptors.
```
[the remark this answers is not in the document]
mitzimorris: filenames?  sorry for getting too fancy.

designs/0032-cmdstan-output-formats.md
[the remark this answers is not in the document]
mitzimorris: added

designs/0032-cmdstan-output-formats.md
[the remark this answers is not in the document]
mitzimorris: made it clear - human-readable tabular format is CSV, binary format is Apache Arrow.

designs/0032-cmdstan-output-formats.md
mitzimorris: reworded.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+There are two general kinds of information structure: tabular and hierarchical.
+[CSV files](https://en.wikipedia.org/wiki/Comma-separated_values) are tabular,
+as are R and Python dataframes and database tables.
+JSON is a [structured data formats](https://en.wikipedia.org/wiki/Data_exchange#Data_exchange_languages)
+which can represent hierarchical structures and arrays.
+
+## Current implementation
+
+We review the way in which the current Stan CSV file is generated for a run of the NUTS-HMC sampler.
```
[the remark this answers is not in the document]
mitzimorris: correct.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Output mechanism:  callback writers
+
+Outputs are handled by objects of base class `stan::callbacks::writer`
+which is constructed from an output stream.
+Through function call operator overloading, the callback writer class
+methods provide the appropriate formatting for different kinds of data.
```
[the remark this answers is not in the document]
mitzimorris: added specifics.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+When the user sends a request to the interface to do some kind of inference given a model and data,
```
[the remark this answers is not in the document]
mitzimorris: revised

designs/0032-cmdstan-output-formats.md
mitzimorris: The `stan::callbacks::stream_writer` adds the requisite CSV formatting, and the CmdStan interface instantiates either `stream_writer` or `unique_stream_writer` objects which are then used everywhere.
updated explanation to make this clear.

designs/0032-cmdstan-output-formats.md
mitzimorris: revised.

designs/0032-cmdstan-output-formats.md
mitzimorris: no, will remove.

designs/0032-cmdstan-output-formats.md
mitzimorris: agreed - revised

designs/0032-cmdstan-output-formats.md
mitzimorris: https://github.com/stan-dev/stan/blob/b703415de6c91d627b0b3c26991c87da0c209922/src/stan/services/util/mcmc_writer.hpp#L81-L85

mitzimorris: simplified this description.

designs/0032-cmdstan-output-formats.md
mitzimorris: `get_param_names` (sic) - names of model variables
`unconstrained_param_names`, `constrained_param_names` - names plus indices

designs/0032-cmdstan-output-formats.md
mitzimorris: initializiations not part of config - anything that's read in from a var_context object not recorded.

designs/0032-cmdstan-output-formats.md
mitzimorris: revised.

designs/0032-cmdstan-output-formats.md
mitzimorris: adding examples.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+All the interfaces and downstream consumers of Stan output have
+to implement ad-hoc parsers to deal with the multiplicity of information
+in a Stan CSV file.
+Providing appropriately structured outputs facilitates development
+of more, better, and faster downstream analyses.
+
+# Prior art
+[prior-art]: #prior-art
+
+N/A (This is a refactor.  The existing system is the prior art.)
```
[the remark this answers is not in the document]
mitzimorris: not sure what prior art is, actually.
I added a link to the discussion on Discourse from summer 2018.

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to resolve through the RFC process before this gets merged?
+
+Agreement on division of sampler / model outputs in separate but parallel CSV files.
+
```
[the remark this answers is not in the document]
mitzimorris: adding examples.  rewriting the above has made things a lot clearer - thanks for all the comments!

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+
+Use of Apache Arrow/Parquet formats as alternatives for tabular data.
```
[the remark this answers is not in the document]
mitzimorris: agreed - committing to Apache Arrow.

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,294 @@
...
+- RFC PR:
+- Stan Issue:
+
+# Summary
+[summary]: #summary
+
+In order to monitor the progress of CmdStan's inference algorithms,
+the outputs from CmdStan must be available to downstream readers on a streaming basis.
+This design provides an alternative to the [Stan CSV file format](https://mc-stan.org/docs/cmdstan-guide/stan_csv.html)
+for outputs of the Stan inference algorithms which are exposed by the `stan::services` layer.
```
SteveBronder, line 12: Can you add one more sentence here describing more specifically what your alternative is? It would just help me understand where the motivation section is going towards

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,294 @@
...
+Depending on the inference algorithms, there are zero or more columns whose names end in `__`.
+For HMC these are `accept_stat__`,`stepsize__`,`treedepth__`,`n_leapfrog__`,`divergent__`,`energy__`,
+for ADVI these are `log_p__` and `log_g__`.
+The optimization algorithms don't report any state information beyond `lp__`.
+
+* Because each row of the CSV file corresponds to one draw, there is no way to monitor the inner workings
+of the inference algorithm, e.g., leapfrog steps or gradient trajectories.
+
+* Although we can now run multiple chains in a single process, it is still necessary to produce per-chain CSV files,
+with the chain id baked into the filename.   A cleaner solution would be to combine all outputs in a single file.
```
SteveBronder, line 45: This would get pretty expensive from an i/o side because then we have multiple threads writing to a single file. For any output format I think we would need to use something clever to make sure writing for one chain would not go over and start writing where another chain just wrote in the file.

SteveBronder (commented):
Some general comments! A larger one for me is what will the API look like for the service layer? I do like the idea of making the csvs simpler, though I'm interested in what the json writing API would look like and how a user would use it in their algorithm. The json output would be nice for the diagnostics layer for sure. I do also like the idea of the binary format for output but I think it would be a good idea to hone down what's here. You describe the problems well with respect to the current output scheme, but I'd like it to be a little more clear about what the proposed solution would look like, either in some light pseudocode or a description of the API

designs/0032-cmdstan-output-formats.md
```diff
@@ -0,0 +1,230 @@
...
+
+This refactoring is in line with the original design.
+
+- What other designs have been considered and what is the rationale for not choosing them?
+
+For the output formats, we considered using Google's protocol buffers.
+These offer data compression, but would require us to develop a set of schemas
+for all outputs.
+Furthermore, protocol buffers are a Google library with no guarantee of backwards compatibility,
+
```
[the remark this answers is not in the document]
mitzimorris: noted - revised.

designs/0032-stan-output-formats.md
mitzimorris: good point - will add more discussion.

designs/0032-stan-output-formats.md
mitzimorris: added more to the summary as well.

mitzimorris:
@WardBrian this is ready for re-review

WardBrian:
Will do another pass tomorrow!

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+(the "header row"), and also allows comment rows which start with a designated comment prefix character.
+Because the initial focus of the project was MCMC sampling, the CSV file allowed for a
+straightforward representation of the posterior sample one row per draw,
+one column per output from the Stan program.
+Over time, the Stan CSV file has come to be an amalgam of information about the inference engine configuration,
+the algorithm state, and the algorithm outputs.
+The formats of the per-inference algorithm outputs are described in the
+[appendix of the CmdStan User's Guide](https://mc-stan.org/docs/2_29/cmdstan-guide/stan_csv.html)
+
+### Output mechanism:  callback writers
```
WardBrian, line 140: I think this section should also describe what you want to change this to - will the list of data structures be expanded (e.g, directly passing something like an Eigen matrix)? Will function overloading still be used, or will we define a class with more functions like `write_diagnostic`, `write_draws`, etc? (I advocate for the latter, since the current implementation where a string is implicitly a comment and everything else is a draw is difficult to read, and doesn't allow generalization if you want string to mean two different things)

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+provide name, type, and shape information.
+The reason that the model reports on both the constrained and unconstrained parameters is the Stan parameter types
+simplex, correlation, covariance, and Cholesky factors have different shapes on the constrained, unconstrained scale.
+For example, a simplex parameter is an N vector whose elements sum to one on the constrained scale which means that it has N-1 free parameters
+when computing on the unconstrained scale.
+The Stan model parameter inputs and outputs are on the *constrained* scale, therefore the number of parameter values reported by the `write_array` method
+is reported by the `get_constrained_sizedtypes` method.
+The sampler computes on the *unconstrained* scale, therefore the size of the metric (inverse mass matrix) is reported by the `get_unconstrained_sizedtypes` method.
+
+### Inference algorithm outputs
```
WardBrian, line 180: Unclear if this is describing the current status or what you're proposing. If the current, will this structure be maintained or changed (and how?)

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Implementation
+
+To improve and generalize the outputs and output formats, we propose to:
+
+- implement a new set of algorithm-specific output handlers which
+will be passed in to the `stan::services` layer wrapper function and which
+will manage the necessary callback writer objects.
+
+- decouple the output format from the writer through the use of formatter callbacks.
```
WardBrian, line 230: This gets at my earlier question a little, but I think it needs a bit more. Some of this may not be clear until implementation, but I would love a sketch of the callback API, and a bit on how the same information would be formatted differently if using the csv/legacy version and the newer arrow/json version

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+The JSON and Apache Arrow formatters would need to be able to handle lists and dictionaries, for the former,
+and scalar and vectors of ints and doubles for the latter.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The more different kinds of output files, more effort required on the part of the user
+to configure and manage these outputs.
+
```
WardBrian, line 244: I think it's not unfair to also say that the initial implementation of this would be a lot of work, and that (it sounds like) the structure of the callbacks will be inherently more complicated. That's not a dealbreaker, of course

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Why is this design the best in the space of possible designs?
+
+The proposed refactoring uses the writer callbacks and extends them.
+
+- What other designs have been considered and what is the rationale for not choosing them?
+
+For the output formats, we considered using Google's protocol buffers.CmdStan output have
+This is a Google library which has been open-source since 2008.
```
WardBrian, line 256: This sentence seems like it got mangled somehow

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+They provide good data compression and fast transport.
+However, the compressed format requires schemas used for serialization and deserialization.
```
WardBrian, line 258: Does Arrow not also require a schema? I think one issue is that the protobuf schemas need to be code-gen'd/compiled, and arrow does not?

WardBrian (changes requested):
Thanks for the mountain of work on this. I'm getting excited that we're closing in on a design which sounds very useful.

My thoughts on the current draft:

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+as well as individual model variables.
+Examples of the latter are the algorithm state at each step of an interactive process.
+For downstream analysis we need to evaluate the goodness of fit  (R-hat, ESS, etc),
+and compute statistics for all model parameters and quantities of interest.
+
+We propose to develop additional output handlers which produce multiple output streams,
+each reporting on one facet of the model-data-inference process.
+Given a designated directory, the output handler will create appropriately named output files in this directory.
+These streams will be in known, standard formats, allowing for easier downstream processing.
+For backwards compatibility we will maintain the current CmdStan Stan CSV output file format, as is.
```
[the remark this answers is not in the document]
mitzimorris: revised accordingly

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+e.g., `num_chains=4` the output is a set of files `output_1.csv`, ..., `output_4.csv`.
+Under this proposal, we would factor the information `output.csv` as follows:
+
+- Structured non-tabular data (nested dictionaries and or/lists) would be output as JSON:
+
+  + `model_metatdata.json` - information on model variable names, types, dimensions, and block 
+  + `sampler_metatdata.json` - information on the configuration of the sampler
+  + `hmc_parameters.json` - information on the stepsize, metric type, and metric.
+
+- Tabular data would be output either as CSV file format or Apache Arrow binary format:
```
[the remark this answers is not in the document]
mitzimorris: > but the way I imagine it each one of these output files you described here gets a writer, and if you pass the same csv_writer to all of them, you should get the same output as the existing CmdStan. 

yes, modulo no comments.

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+For example, if we use the CmdStan interface to run the sampler for a single chain
+by default the output is written to a file `output.csv` and if we use the new `num_chains` argument,
+e.g., `num_chains=4` the output is a set of files `output_1.csv`, ..., `output_4.csv`.
+Under this proposal, we would factor the information `output.csv` as follows:
+
+- Structured non-tabular data (nested dictionaries and or/lists) would be output as JSON:
+
+  + `model_metatdata.json` - information on model variable names, types, dimensions, and block 
+  + `sampler_metatdata.json` - information on the configuration of the sampler
+  + `hmc_parameters.json` - information on the stepsize, metric type, and metric.
```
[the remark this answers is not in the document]
mitzimorris: yes, related - no reason not to combine them.

designs/0032-stan-output-formats.md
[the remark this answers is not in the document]
WardBrian: If the comments will not be in the new CSV file, then we should not claim that these files are backwards compatible with existing implementations (e.g, `cmdstanpy.from_csv` will not work with them without significant changes, therefore they're not backwards compatible).

This is not necessarily a dealbreaker, but it does conflict with the claim in the design doc that these will be backwards compatible. IMO, a truely backward compatible solution would be capable of producing exactly the same CSV file before and after these changes.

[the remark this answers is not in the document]
mitzimorris: for backwards compatibility, we can keep the existing service methods, adding new ones with a better argument structure.  these should be marked deprecated, and eventually no longer supported.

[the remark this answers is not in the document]
mitzimorris: further revised per this discussion.

designs/0032-stan-output-formats.md
mitzimorris: we will define a class with more functions like `write_diagnostic` etc.

mitzimorris: revised.

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,295 @@
...
+will be passed in to the `stan::services` layer wrapper function and which
+will manage the necessary callback writer objects.
+
+- decouple the output format from the writer through the use of formatter callbacks.
+For tabular data we will implement CSV and Apache Arrow formatters.
+For extra-tabular structured data we will implement a JSON formatter.
+
+The CSV structured data formatter corresponds to the current set of methods on the base callbacks writer class.
+The JSON and Apache Arrow formatters would need to be able to handle lists and dictionaries, for the former,
+and scalar and vectors of ints and doubles for the latter.
```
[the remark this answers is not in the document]
mitzimorris: yes and no - underlyingly, Arrow is a tabular format - there's always flattening.  the schema is sent across the wire before any of the data and this allows the downstream processing (also Arrow) to reconstitute the data according to the schema.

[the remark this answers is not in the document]
mitzimorris: this means that part of the job of the output handler, if Arrow/parquet format is specified, is to get the model metadata and create the schema - after model instantiation, before doing inference.

betanalpha:
As my previous comments have gone unaddressed allow me to repeat them.  The implicit target of this design is programatic consumers of Stan output, such as CmdStanR and CmdStanPy, which have co-opted CmdStan for a use for which it was never intended.  The intended target of CmdStan is command line users and that intent guided the design of the current CmdStan output (sequentially updated, human readable files that can be monitored with command line utilities like `cat` and `top`, pseudo-CSV outputs that are compatible with command line utilities like `sed`, `grep`, and `awk`, etc).  

Programatic consumers with require more structured output are very much a relevant need, but they are better serviced with a dedicated interface rather than the modification of an existing interface.  Indeed the writer callback API was introduced to facilitate the implementation of different interfaces to the Stan services with different goals.  Previous comments alluded to a new interface altogether, with names like `ShellStan`, which I think would be the most productive step forwards although I don't see any reference to that in the current version of the design document.

At the very least I think that the design document would benefit from explicitly stating the desired audience and any conflicts with the originally intended CmdStan audience.

Otherwise I am worried about the ambiguity of the current design regarding the inference output.  The design refers to the need to "decouple the outputs from the inference engine from the outputs from the model" but this isn't well-defined.  

The "model" is just the specification of a log density function and its gradient (possibly a conditional log density function depending on how the variables in the `data` block are used; also one might consider the variables defined in the `transformed parameters` and `generated quantities` block as auxiliary outputs defined by a model instance) which can be evaluated at a point in the model configuration space (i.e. parameter space).  There is no canonical input and hence no canonical output.

Instead each inference algorithm defines where the model needed to be evaluated in order to drive some estimation of various expectation values (by default the expectation values of each parameter function as well as the functions implicitly defined by variables in the `transformed parameters` and `generated quantities` blocks).  The model evaluations needed to inform those expectation value estimates will vary from algorithm to algorithm and so there is no way to decouple the output format (at which inputs was the model evaluated and what were the corresponding outputs) from the algorithm.

For example as an inference algorithm optimization algorithm returns as single point in the model configuration space.  It could report the point in the unconstrained model configuration space, but at the moment it reports the point in the constrained model configuration space as well as all of the associated output evaluations.  Currently there is also intermediate output that reports the points along the optimization path which are written to the same output CSV even though those points technically have no inferential value.

The Hamiltonian Monte Carlo sampler is based on off of Markov chain Monte Carlo which returns a sequence of points that can be used to construct Markov chain Monte Carlo estimators.  This means that the inferential output is not a single point but rather a sequence of points, and for utility the evaluation of the model at those points.  Auxiliary output might include the unconstrained model configuration, the gradient of the log target density, the phase space momenta generated along with that point, as well as intermediate points along each numerical Hamiltonian trajectory.  Note that all of this auxiliary information is structured like the initial sequence of points but doesn't have the same inferential meaning.

In theory variational Bayes doesn't return any points but rather the parameters of the "optimal" density function within some specified variational family.  In practice that optimal density function isn't used directly but rather as a source for exact samples that can inform Monte Carlo estimates.  For that particular implementation the output would appear similar to Markov chain Monte Carlo output but the meaning would be slightly different (exact vs correlated samples).

Critically many of these outputs, some with direct inferential utility and some without, have exactly the same structure.  `Reference-level explanation` refers to separating outputs by whether they are "tabular" or "structured" but it doesn't refer to how these various outputs might be organized based on the originating algorithm.  Would they go into separate files?  The same file with additional columns differentiating between their precise meaning?

`log_prob: provides access to the model parameters and transformed parameters on the 
unconstrained scale` 

This isn't correct.  The `log_prob` signatures all return the unnormalized target density and its gradient given the unconstrained model parameters.

In the `Implementation` section there isn't any reference for what external libraries, if any, would be needed for implementing writers capable of outputs JSON or Arrow output.  I think that these details are critical to moving the design forward.

Let me summarize my main concerns:
  - Motivation is based on a particular consumer without any discussion of that consumer and the intended CmdStan consumer.
  - The suggested design refers only to separating output by particular structure, and hence possible output formats, but not how those output formats would be organized by algorithm.  The algorithm-specific organization also conflicts with the stated goal of decoupling output design from algorithm details.
  - Little detail on how the suggested writers would be implemented.

To be clear I think that having JSON and Arrow writers would be a great feature to add, and is almost completely independent of my concerns.  I think that a narrow focus on those writers, at least a first, would be a helpful way to move forwards while continuing larger interface discussions.

designs/0032-stan-output-formats.md
mitzimorris: let's talk.

designs/0032-stan-output-formats.md
mitzimorris: thanks.

designs/0032-stan-output-formats.md
mitzimorris: right.

mitzimorris:
@betanalpha thank you very much - this is extremely instructive.

WardBrian:
An alternate interface (e.g., building something other than CmdStan) with these outputs would still require changes to the algorithms/services layer, I believe, so there are still questions of how to do that in such a way that they play nice with one another

mitzimorris:
rewriting proposal to make it clear that this is *not* about changing CmdStan - it's about changing the services layer so that everyone plays nicely - like Brian said.

mitzimorris:
I have incorporated Bob and Brian's comments.

This design allows for continued support for the current monolithic Stan CSV file used by CmdStan while providing new formats for other different interfaces.

designs/0032-stan-output-tech-spec.md
```diff
@@ -0,0 +1,74 @@
...
++ Eigen array, matrix, and vector types
+
+
+### Creating Apache Arrow schemas for inference algorithm and Stan model data
+
+The instantiated model object provides the information used to create an Apache Arrow schema
+used to serialize the values produced by the model's `write_array` function.
+### Stan program outputs
+
+The Stan model class `write_array` method populates a vector of doubles with the values for all
```
WardBrian, line 33: How does this vector of doubles square with the above description of being able to accept Eigen types and integers in the writer interfaces?

WardBrian (commented):
Fewer comments this time, nearly all questions/clarifications. I think the overall idea layed out here is solid and a good way forward

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,288 @@
...
+
+
+### Kinds of outputs, suggested names
+
+Files which are updated with each iteration (i.e., streaming data), which can be monitored by downstream processes:
+
+- `sample` - A table containing the per-draw estimates of all model variables on the constrained scale; 1 column per variable (element), one row per iteration.
+
+- `log_prob` - A table containing the value of `lp__`, the joint log probability density, one row per iteration.
+- `algorithm_state` - A table containing the per-iteration state of the inference algorithm, e.g. `accept_stat__`, `log_p__`, one row per iteration.
```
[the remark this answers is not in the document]
mitzimorris: per draw.  iteration is ambiguous.

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,288 @@
...
+
+Note that if we adjoin all the columns in tables `log_prob`, `algorithm_state`, and `_sample`, the resulting data table would contain the equivalent information
+in the data table rows of Stan CSV file.
+
+Files which contain global information, produced once per inference algorithm run:
+
+- `stepsize` - A single element table containing the stepsize.
+- `metric` - A table which contains the inverse mass matrix arrived at by adaptation.
```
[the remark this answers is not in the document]
mitzimorris: yes, they definitely could be in a single JSON file and this would be the right thing to do because it would allow for re-use - that's the argument I was looking for.

designs/0032-stan-output-tech-spec.md
mitzimorris: for `write_array` some of those doubles are actually representing ints.
some outputs from algorithms might return Eigen types.
I think we should design for this eventuality.

WardBrian: Yes, it's difficult to think of a way in which write_array could be written generically. It would be great if we could though, since we could keep the high-level abstraction without needing as many packing/unpacking loops

mitzimorris: changes to write_array are way beyond the scope of this proposal.

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,288 @@
...
+### Output fomats
+
+Hierarchical data will be output using JSON notation.
+Tabular data will be either in CSV format (human-readable) or Apache Arrow format (binary).
+An Apache Arrow parquet file consists of an schema followed by one or more rows of data.
+The schema describes the structure of the data objects in each row, allowing the downstream
+process to reconstitute structured objects properly.
+The Arrow libraries for R and Python will allow downstream analysis packages to process these files, either during inference or afterwards.
+
+The Stan CSV output format will continue to be used for CmdStan outputs,
```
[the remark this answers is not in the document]
mitzimorris: what do you mean by "single binary" and what to FB Prophet do?

designs/0032-stan-output-tech-spec.md
WardBrian: So, the overloads discussed in this tech spec (i.e. writer callbacks which can accept a vector of ints, or an eigen matrix) are not for model output but rather for internals to the algorithms or that sort of thing?

designs/0032-stan-output-formats.md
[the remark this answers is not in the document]
WardBrian: I mean that there is a compiled executable which contains the model. My original comment was when I was still under the impression that the Stas proposal would include compiling the model and dynamically linking it, so this is resolved.

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,336 @@
...
+the functional specification, e.g., `LP` (log_prob), `sample`, `metric`, etc.
+This provides an extensible mechanism for adding future outputs.
+
+- The values of enum class `OutputFormat` correspond to available output formats.
+We expect to implement at least the following:
+
+    + `Arrow` - tabular data in Apache Arrow, other structured data is JSON
+    + `Csv` - tabular data in CSV, other structured data is JSON
+    + `Raw` - tabular data is output as vector of binary values
+    + `StanCsv` - legacy Stan CSV format
```
WardBrian, line 269: I don't think I understand what the purpose of this enum is if each output format gets its own subclass of `OutputWriter` as discussed in the next paragraph (which I think is correct).

To me, you only need the `InfoType` enum and then the classes you list below, this enum is then redundant. Unless I am misunderstanding?

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,336 @@
...
+
+We will define an `OutputWriter` class which and we will add new methods to
+the services layer calling functions which take a single argument `output_writer`u
+(instead of arguments `sample_writer` and `diagnostic_writer`).
+The `OutputWriter` class will be sub-classed by output format:
+`ArrowOutputWriter` `CsvOutputWriter`, `RawOutputWriter`, `StanCsvOutputWriter`.
```
WardBrian, line 275: [comment, implementation detail]

I think you could even make a `TabularAndJsonOutputWriter` abstract class and have `ArrowOutputWriter` and the (non-legacy) `CSVOutputWriter` both extend from that, since they'll be re-using json for non-tabular. This is too fined grained for this document, just wanted to mention that there is the opportunity for that sort of thing

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,336 @@
...
+
+The `OutputWriter` object will be instantiated by the Stan interfaces
+from a Stan model object and a set of outputs of interest.
+The instantiated model object provides the information needed to format and filter
+the values produced by the model's  `write_array` function.
+
+The `OutputWriter` class provides a callback method `write_info`
+which is parameterized by information type, and its corresponding data.
+Based on the information type, the `OutputWriter` object will write
+the data to the appropriate output stream.
```
WardBrian, line 285: I'd like at least one example of this. Here's what I'd say:

For example, `ArrowOutputWriter` will send `LP` information to a file `lp.arrow`,  `Sample` information to `draws.arrow`, and `metric` info to `adaptation.json`. To contrast this, `StanCsvOutputWriter` will send all of this information to a single `output.csv` file in the format described at the start of this design.

designs/0032-stan-output-formats.md
```diff
@@ -0,0 +1,336 @@
...
+## Technical Specification
+
+To implement the features outlined in the Functional Specification
+will require changes in the `stan::services` and `stan::callbacks` layers.
+
+We will define two [C++ Enumeration](https://en.cppreference.com/w/cpp/language/enum)
+classes: `InfoType` and `OutputFormat`.
+
+- The values of enum class `InfoType` correspond to the information items listed in
+the functional specification, e.g., `LP` (log_prob), `sample`, `metric`, etc.
```
WardBrian, line 260: This is a prime area for discovery during implementation, but just for the sake of getting names right I think you'd probably want them to be `Lp`, `Draws` or `Samples`, and `Adaptation` (not just metric, e.g. the same info type could also be step size?)

Anyway, I think it's worth clarifying that the exact names and list of these will be part of the implementation, not something fixed in advance

WardBrian (changes requested):
I have a few comments/questions, but assuming those are addressed I think this has settled into the place we want.

bob-carpenter:
> The current CmdStan output was designed for interactive command line use. In particular it was designed for use with the common command line utilities such as sed, grep, and awk.

A common use case for CmdStan is to generate output and then read it into data analysis tools like R or Python.   But even considering just the unix tools, the output should be easier to work with in separate files rather than being hacked into a CSV file's comments.

My concern is not that the existing system is better, but that if we change it, any tools built on top of the current CmdStan will break.

 only concern I have is that tooling on top of the existing hacks will break if we clean up the I/O.

My own preference would be to build a replacement for CmdStan that's more standardized in terms of command-line structure and uses standards for I/O rather than CSV hacks.  I think that would be better for both command line users and our interfaces, but more importantly, it won't break any backward compatibility with tools built on top of CmdStan.

designs/0032-stan-output-formats.md
mitzimorris: added

mitzimorris: added - thanks!

PR #46, opened 2022-04-22
Revised proposal: Tuples

WardBrian:
This design doc takes the information from Bob's original proposal here: #24 

It reformats it to the standard RFC format and adds additional information on the currently proposed implementation. 
In particular, it details IO formats and some of the implementation challenges.

Rendered version:
https://github.com/WardBrian/design-docs/blob/tuples/designs/0018-tuples.md

### DIFF
--- a/designs/0018-tuples.md
+++ b/designs/0018-tuples.md
@@ -1,0 +1,383 @@
+- Feature Name: Tuples
+- Start Date: 2020-04-23
+- RFC PR:
+- Stan Issue: https://github.com/stan-dev/stan/issues/2431
+
+# Summary
+[summary]: #summary
+
+This proposal is to add tuples to the Stan language. Tuples are _static_,
+heterogeneous containers, meaning that they have a known, fixed size in the
+program and can contain elements which are of different types and sizes. Tuples
+are common in many languages from Python to OCaml and C++. Formally, a tuple is
+a product type of its contained types.
+
+# Motivation
+[motivation]: #motivation
+
+Tuples support several important use cases.
+
+One common use for tuples is to support _multiple return_, allowing functions to
+return more than one type of value simultaneously. For example, a function could
+be defined which returns both a matrix and some real-valued property of this
+matrix. In the math library, this can be used for cases such as singular value
+decomposition. Currently, two different functions must be called to receive the
+two portions of the decomposition, and this duplicates computation.
+
+Some forms of data are best represented by a mix of different types or different
+sizes of the same type; currently it is not possible in Stan to package these
+together.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A tuple is the generalization of the idea of a pair to any number of elements.
+In Stan, tuples allow packaging of different types together into one value.
+
+## Declaration
+
+A tuple variable is declared as follows
+
+```
+(T1, ..., TN) x;
+```
+
+where `T1` through `TN` are themselves type specifications. Tuples are
+recursive - they can contain other tuples.
+
+A pair containing one integer and one real value could be declared
+
+```
+(int, real) p;
+```
+
+These types can be any Stan type which would be valid in that block/context. For
+example, one can define a parameter which is a pair of an array of real values and a
+simplex like so
+
+```
+(array[5] real, simplex[10]) param;
+```
+
+Tuples can be used as function arguments and return types.
+
+## Tuple expressions
+
+The expression for creating tuples can follow the array notation (`{ a, b, c }`)
+and vector notation (`[a, b, c]`) and use `(a, b, c)`.  This will create a tuple
+of type `(typeof(a), typeof(b), typeof(c))`, each subject to the normal
+promotion rules.
+
+For example, `(1.5, {2.3, 4})` is a tuple with type `(real, array[2] real)`.
+
+## Tuple accessors
+
+A key difference between tuples and other containers in Stan is that they have a
+fixed size which is known during compilation. As such, they are accessed using
+indexing which is also known during compilation. This is done using `.` followed
+by a integer literal after the tuple. This static indexing is also known as
+projecting.
+
+For example,
+```
+(S, T, U) x;
+...
+S a = x.1;
+T b = x.2;
+U c = x.3;
+// x.4 causes compilation error
+```
+
+It is **not** possible to write
+```
+x[1]; // not valid
+x.n; // not valid
+```
+
+## Assignment
+
+Tuples and their elements may be used on the left hand side of assignment
+statements, as in:
+
+```
+(double, int) a;
+a = (3.7, 2);  // assign to
+a.1 = 4.2;
+```
+
+## Promotion (covariance)
+
+Like other Stan container types, tuples are covariant. That means that whenever
+`a` is a subtype of `b` (e.g., `int` is a subtype of `real`), then
+`(..., a,...)` is a subtype of `(..., b, ...)` as a tuple.
+
+This would allow you to assign an integer and double tuple to a pair of double
+value tuples, etc. For example, `(int, double, matrix)` assignable to
+`(double, double, matrix)`.
+
+## Input and Output formats
+
+These proposals are specific to the CmdStan interface.
+
+### JSON input format
+
+Tuples are represented as JSON objects with indices stored as keys.
+
+For example, the Stan declaration `(int, array[2] real) d;` can be represented
+in JSON as follows:
+
+```json
+{
+  "d": {
+    "1": 3,
+    "2": [3.5, 6.7]
+  }
+}
+```
+
+### CSV output format
+
+Like all Stan objects, tuples are flattened into a single row for output. Due to
+the existing convention of using `.` for array and matrix indices in the output
+format, the `:` character is used.
+
+For example, the Stan declaration `(int, array[2] real) d;` will produce the
+following column headers in the output CSV file:
+
+```csv
+d:1, d:2.1, d:2.2
+```
+
+## Special functions (optional)
+
+The following special functions for tuples could be defined. These would be
+impossible to write in a generic way in the language itself, so any
+implementation would need to be in the standard library.
+
+```
+T1
+head((T1, ..., TN) a);
+```
+
+```
+(T2, ..., TN)
+tail((T1, T2, ..., TN) a);
+```
+
+```
+(T1, T2, ..., TN)
+concat(T1 a, (T2, ..., TN) b);
+```
+
+## Destructuring assignment (optional)
+
+A common feature of other languages with tuples is destructuring ("unpacking")
+through assignment, e.g.
+
+```
+int x;
+int y;
+(x, y) = (1, 2)
+```
+leaves `x` with value `1` and `y` with value `2`.
+This feature makes programming with tuples much more elegant.
+This feature is often paired with the the convention of having `_` serving as an
+indicator to ignore a value, so the assignment `(x, y, _) = (1,2,3);` has the
+same result as above.
+
+Some languages support this without the need for the enclosing parenthesis. This
+is possible in Stan, but would have conflicts with the multiple declaration
+syntax `int x,y;` -- the line `int x,y = a;` could be interpreted two different
+ways depending on if `a` has type `int` or type `(int,int)`.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## C++ Implementation
+
+Tuples will be represented directly by using the C++11 `std::tuple` type.
+
+### C++ I/O Support
+
+#### Serializer interface
+Serialization/deserialization of tuples (i.e., in parameters or for the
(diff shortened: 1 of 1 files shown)


betanalpha:
This all looks great.  I am sure that these have already been discussed repeatedly and I don't want to bike shed, but a few I have a comments about potential user experience.

What is the motivation for having`({ a, b, c })` and `([a, b, c])` be equivalent to `(typeof(a), typeof(b), typeof(c))`?  If `typeof(a) = typeof(b) = typeof(c)` won't there be a conflict between `(typeof(a), typeof(b), typeof(c))` and the array singleton `(array[3] typeof(a, b, c)) or if `typeof(a) = typeof(b) = typeof(c) = real` then a conflict between `(typeof(a), typeof(b), typeof(c))` and the the vector singleton `(vector[3])`?  This seems like an unnecessary source of confusion when the tuple declarations can just be restricted to `(a, b, c)`.

Why the new accessor for individual elements?  Because both arrays and linear algebra use `[]` to access individual elements the language already has a convention for using `[]` to access elements of a container.  Given that tuples are just heterogeneous containers it seems like it would be natural to continue this convention and not confuse users with another notation.

I have to say that I'm a huge fan of `tuple(T1, T2, T3)` instead of `(T1, T2, T3)`.  The extra verbosity is minimal and it emphasizes to users that this is a new container type.  This would also be consistent with the recent move to the more verbose `array` syntax.  As an additional feature is that this would provide an explicit differentiation between declarations and unpacking,
```
real a, b, c;
tuple(real, real, real) x = (2, 2, 2);
(a, b, c) = x; // LHS is not the declaration of a variable but temporarily packing existing variables together
```
verses the more ambiguous
```
real a, b, c;
(real, real, real) x = (2, 2, 2);
(a, b, c) = x;
```
Yes the difference between variable names and type names should be enough to separate the two operations but I think that the added redundancy would be more accessible to the majority of the user base (again similar to the `array` syntax).

WardBrian:
> What is the motivation for having`({ a, b, c })` and `([a, b, c])` be equivalent to `(typeof(a), typeof(b), typeof(c))`

I’m not sure I follow your question, the snippets you’re referencing are establishing what already exists for arrays and row vectors, not what would be used for a tuple. Tuple values would only use {} or [] if they contain arrays or linear algebra types.

> Why the new accessor for individual elements?

Besides being heterogenous, the key distinction between a tuple and an array is that the tuple has a static, fixed size. The use of `.` instead of `[]` is useful precisely because it is different - one should consider tuples separately than arrays or linear algebra types, since they cannot be iterated over dynamically and you cannot access them with a variable index, only a raw integer. This style is also how struct access would most likely look if we’re going off of structs in other languages. 


> I have to say that I'm a huge fan of `tuple(T1, T2, T3)` instead of `(T1, T2, T3)`.

I generally prefer the more succinct syntax, since I think of the fact that tuple types and values are written down very similarly as a win for understandability, but I take your point with the array syntax as well. I do feel quite strongly about not going so far as to make tuple values look like `tuple(1.2, 4)`, though. 

I’m not sure that the unpacking/assignment syntax has any ambiguity if we require parenthesis around the LHS of an unpacking. This is not needed in languages like Python, but it is a decent choice in a statically typed language. That way, if you leave the parenthesis off, it will simply be a type error, e.g
`real a, b = (2.5, 3.5)`
Will fail, since it will parse as trying to assign a tuple of reals to the singular real variable b.

betanalpha:
> I’m not sure I follow your question, the snippets you’re referencing are establishing what already exists for arrays and row vectors, not what would be used for a tuple. Tuple values would only use {} or [] if they contain arrays or linear algebra types.
> 
In that case I think the sentence

`The expression for creating tuples can follow the array notation ({ a, b, c }) and vector notation ([a, b, c]) and use (a, b, c). This will create a tuple of type (typeof(a), typeof(b), typeof(c)), each subject to the normal promotion rules.`

is too confusing as it seems to imply that `{ a, b, c })`,  `([a, b, c])` and `(a, b, c)` all construct a tuple of type `(typeof(a), typeof(b), typeof(c))`.

> Besides being heterogenous, the key distinction between a tuple and an array is that the tuple has a static, fixed size. The use of . instead of [] is useful precisely because it is different - one should consider tuples separately than arrays or linear algebra types, since they cannot be iterated over dynamically and you cannot access them with a variable index, only a raw integer. This style is also how struct access would most likely look if we’re going off of structs in other languages.
> 
I appreciate this motivation for the different accessor methods, although I think that the initial documentation will have to be _very_ clear and redundant about the differences between tuple containers and array containers.  I think that there will be many “how do I loop over tuples” questions from users who pattern match tuples to how they already use arrays.
> I have to say that I'm a huge fan of tuple(T1, T2, T3) instead of (T1, T2, T3).
> 
> I generally prefer the more succinct syntax, since I think of the fact that tuple types and values are written down very similarly as a win for understandability, but I take your point with the array syntax as well. I do feel quite strongly about not going so far as to make tuple values look like tuple(1.2, 4), though.
> 
I definitely agree that `(1.2, 4)` should be sufficient for defining inline tuple values.

In my experience one of the biggest hurdles to learning the Stan language for new users who are used to less strongly/staticly typed languages like R and Python is the type system.  That’s why I think that being as explicit as possible about the type declarations, even at the expense of some mild additional verbosity, will be beneficial overall.  
> I’m not sure that the unpacking/assignment syntax has any ambiguity if we require parenthesis around the LHS of an unpacking. This is not needed in languages like Python, but it is a decent choice in a statically typed language. That way, if you leave the parenthesis off, it will simply be a type error, e.g
> real a, b = (2.5, 3.5)
> Will fail, since it will parse as trying to assign a tuple of reals to the singular real variable b.
> 
Sorry I wasn’t referring to formal ambiguity of the compiler but rather cognitive ambiguity of users.  This new type introduces a few new concepts that I think can trip up many users, and I think that language design that explicitly discriminates between type declarations and assignments where appropriate would help ease that burden.

WardBrian:
> is too confusing as it seems to imply that `{ a, b, c })`, `([a, b, c])` and `(a, b, c)` all construct a tuple of type `(typeof(a), typeof(b), typeof(c))`.

I will reword that part, good catch.

Agreed on the point of documentation. In addition to being very clear about the issues you describe, I think it will be important to stress what benefits this provides; e.g., it is possible right now to define a function which returns two real values by settting the return type to `array[] real` and making a contract with yourself that this will always be length 2. By contrast, if the return type is `(real, real)` _the compiler will make that guarantee for you_.

I see your point on the extra verbosity of type declarations, and would like to hear the thoughts of some others (@rybern, @SteveBronder, @bob-carpenter). To be clear, the proposed syntax would look like `tuple(int, real, int) = (1, 2.5, 3);` in full. A more complicated type could look like `array[3] tuple(int, array[2] real)`

betanalpha:
> Agreed on the point of documentation. In addition to being very clear about the issues you describe, I think it will be important to stress what benefits this provides; e.g., it is possible right now to define a function which returns two real values by settting the return type to array[] real and making a contract with yourself that this will always be length 2. By contrast, if the return type is (real, real) the compiler will make that guarantee for you.

Yeah, especially since this (very good) motivation is not the motivation that has long been thrown around for tuples which are ragged arrays.  Users coming to this tuple design thinking that it will allow them to loop over ragged arrays will be frustrated, and so the sooner the documentation can communicate the scope of this tuple type the better.

bob-carpenter:
> (@betanalpha) not the motivation that has long been thrown around for tuples which are ragged arrays

I don't know who was thinking that was the motivation, but that was never my intention.  I've always just wanted standard prouct types in the type theoretic sense.  What tuples give you is like a fixed-size, fixed-type version of a list in R.

> (@WardBrian) would like to hear the thoughts of some others

I'm OK with either `tuple(T1, ..., TN)` or `(T1, ..., TN)` for the type syntax.   So that'd be either,


```stan
tuple(real, int, real) x = (2.7, 12, -1e-6);
```

or

```stan
(real, int, real) x = (2.7, 12, -1e-6);
```

I very much like the idea of being able to follow this with:

```stan
real a, c;
int b;
a, b, c = x;
```

which follows Python convention, or, if necessary for the parser,


```stan
a, b, c = x;
```

andrewgelman:
so cool!

WardBrian:
The initial tuples PR we have for stanc doesn't support unpacking, but it is something I want to add soon thereafter. I think it will be possible to translate it directly to a `std::tie` if I read the C++ docs correctly, or we could generate the necessary sequential assignments ourselves.

WardBrian:
During today's monthly language meeting, @spinkney and @avehtari both expressed a preference for the syntax `tuple(T1, ..., TN)` for tuple types, as proposed by @betanalpha. 

Aki made the particular point that the more explicit form makes it much easier to search a large program (either with your eyes or with a tool like `grep`) for tuple types, which has a lot of utility. 

Tuple expressions are still made with parenthesis, so a full example is 

```stan
tuple(real, int, real) x = (2.7, 12, -1e-6);
```

bob-carpenter:
I'm totally OK with the `tuple(...)` syntax.

WardBrian:
After updating the implementation to match it has grown on me more than I thought. It makes things like nested tuples _much_ more apparent to the eye, `(real, real, (real, real))` vs `tuple(real, real, tuple(real, real))` is a big difference.

designs/0018-tuples.md
```diff
@@ -0,0 +1,395 @@
...
+```
+
+These types can be any Stan type which would be valid in that block/context. For
+example, one can define a parameter which is a pair of an array of real values and a
+simplex like so
+
+```
+tuple(array[5] real, simplex[10]) param;
+```
+
```
mitzimorris, line 61: Remind the reader that tuples are static, therefore the dimensions cannot themselves be variables.

designs/0018-tuples.md
```diff
@@ -0,0 +1,395 @@
...
+For example, `(1.5, {2.3, 4})` is a tuple with type `tuple(real, array[2] real)`.
+
+## Tuple accessors
+
+A key difference between tuples and other containers in Stan is that they have a
+fixed size which is known during compilation. As such, they are accessed using
+indexing which is also known during compilation. This is done using `.` followed
+by a integer literal after the tuple. This static indexing is also known as
+projecting.
```
mitzimorris, line 82: It would be helpful to mention accessors in the summary in order to answer the question of why must they be known, fixed size.

designs/0018-tuples.md
```diff
@@ -0,0 +1,395 @@
...
+
+- The call `context.vals_i("data.1")` should return a `std::vector<int>` of
+  length 2.
+- The call `context.vals_r("data.2.1")` should return a `std::vector<double>` of
+  length 2.
+- The call `context.vals_c("data.2.2")` should return a
+  `std::vector<std::complex<double>>` of length 2*3 = 6. The data from this call
+  should be the data for `data[1].2.2` followed by the data for `data[2].2.2`,
+  concatenated together into one vector.
+
```
mitzimorris, line 238: please add example of calls to get the variable dimensions as well.

mitzimorris (approved):
A few suggestions, but overall, the spec is very clear.

designs/0018-tuples.md
WardBrian: Not sure what you're asking for here - it's totally kosher to write

```stan
//data
int N;
// later on
tuple(array[N] real, simplex[N * 2]) param;
```

The static part of tuples is the number of slots in them. e.g `param.1`, `param.2`, not the contents of those slots

mitzimorris:
regarding static - was confused by the language - perhaps "slots" makes more sense?
arrays have  a fixed number of dimensions, but the size of each can be specified at runtime.
tuples have a fixed number of slots, known at compile - analogous to number of array dimensions.
I think we need to be explicit to here to avoid confusion between "size" and "dimension".

WardBrian:
Agreed, especially in the final user-facing documentation. The analogy to arrays is a good one, but also has its own potential to be confusing since the relationship between different dimensions of an array and different elements in a tuple is not the same

designs/0018-tuples.md
```diff
@@ -0,0 +1,405 @@
+- Feature Name: Tuples
+- Start Date: 2020-04-23
+- RFC PR:
+- Stan Issue: https://github.com/stan-dev/stan/issues/2431
+
+# Summary
+[summary]: #summary
+
+This proposal is to add tuples to the Stan language. Tuples are _static_,
+heterogeneous containers, meaning that they have a known, fixed size in the
```
mitzimorris, line 10: this is the source of my confusion.  how about starting with:
> Tuples are heterogenous containers which have a known, fixed number of elements ("slots").

designs/0018-tuples.md
mitzimorris: nice example.  please add to help clarify for reader.

mitzimorris (dismissed):
per discussion, we can be clearer w/r/t tuples structure, but otherwise, this is g2g.

mitzimorris (approved):
very clear!  complete approval!

PR #47, opened 2022-06-27
Generalised framework for specifying function gradients

andrjohns:
This design doc outlines a proposed framework for specifying function gradients without needing to create additional `fwd` and `rev` headers. Functors for calculating the gradients are specified in the `prim` definition of the function, and the proposed `function_gradients` functor handles the application with `var` and `fvar` inputs.

The Math branch with the implementation, and `hypot` and `inverse` as examples is here: https://github.com/stan-dev/math/compare/develop...andrjohns:feature/user-gradients?expand=1

### DIFF
--- a/designs/0033-gradients-framework.md
+++ b/designs/0033-gradients-framework.md
@@ -1,0 +1,192 @@
+- Feature Name: generalized_function_gradients
+- Start Date: 2022-06-07
+- RFC PR: (leave this empty)
+- Stan Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This document proposes a C++ framework for defining gradients for both reverse- and forward-mode at the same time as the definition of the function. This allows for reducing code duplication and boilerplate, and also provides an interface for user-provided gradients in the future.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, when specifying forward- and reverse-mode gradients for functions, developers are required to create additional `fwd` and `rev` headers, with the boilerplate specific to each AD type (e.g., `make_callback_var`). This provides an additional barrier to entry for new developers, as they have to familiarise themselves with the various Math-specific frameworks (e.g., `arena<T>`, `var<Matrix>`, `fvar<var>`). Additionally, scalar functions (generally) have the same gradient specifications for both `fwd` and `rev`, leading to further code duplication.
+
+The proposed framework aims to simplify specification of gradients for functions, and create a facility for user-specified gradients.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Scalar Functions (Same Gradient Functions)
+
+Take the `hypot` function for example, we first define a functor providing the primitive return value:
+
+```cpp
+  auto val_fun = [&](auto&& x, auto&& y) {
+    using std::hypot;
+    return hypot(x, y);
+  };
+```
+
+Next, we specify a functor for each input that calculates the respective gradients. These functors should have the value and adjoint as the first two arguments, and then all inputs as the remaining arguments:
+
+```cpp
+  auto grad_fun_a
+      = [&](auto&& val, auto&& adj, auto&& x, auto&& y) {
+        return elt_multiply(adj, elt_divide(x, val));
+      };
+  auto grad_fun_b
+      = [&](auto&& val, auto&& adj, auto&& x, auto&& y) {
+        return elt_multiply(adj, elt_divide(y, val));
+      };
+```
+
+Finally, a tuple of the input arguments, the value functions, and a tuple of the gradient functions are passed to the `function_gradients` functor:
+
+```cpp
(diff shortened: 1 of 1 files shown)


andrjohns:
Rendered output: https://github.com/stan-dev/design-docs/blob/711ced25eca1971967ef94f416d6a92c93639fbb/designs/0033-gradients-framework.md

I forgot to say that is also intended to provide a means for user-specified gradients as well

WardBrian:
Nice. Compile and runtime timing comparisons would be great to whatever extent they are possible.

Are you imagining that we eventually move all of the autodiff to look like this and therefore don't need the `rev`/`fwd` folders any more?

bob-carpenter:
> Nice. Compile and runtime timing comparisons would be great to whatever extent they are possible.

Yes, on both counts.

> I think in theory it should be possible that this generates the same code, but it's worth checking.

I've been thinking about this a long time and always get bogged down in wanting to share computation of subexpressions across the value function and all the adjoint functions.  For example, the normal value and derivative functions need to calculate (y - mu).   Here's a simple example:

```
x / y = x * inv(y)

d/dx = inv(y) * I

d/dy = -inv(y)' kronecker* (x * inv(y))
```

Both derivatives involve inv(y).  In a simple `normal, the derivatives all involve `(y - mu)`.  It's the encapsulating in terms of independent functions that's the issue.  In the original Stan autodiff design, those things became member variables that were defined by hand and we could be lazy on the backward pass.

For prior art, you might want to consider how Turing.jl interfaces with Turing autodiff like Zygote---they do something very similar.

andrjohns:
Great point Bob, that's been something that I've trying to figure out as well. My current plan of attack would be to have an additional functor that returns a tuple of shared arguments:

```cpp
  value_functor = [&](auto&& x, auto&& y) { return x / y; };
  shared_values_functor = [&](auto&& val, auto &&adj, auto&& x, auto&& y) {
    double inv_y = inv(y);
    // Important to be able to mix types in shared args
    Eigen::VectorXd inv_y_vec = Eigen::VectorXd::Constant(inv_y, 10);

    return std::forward_as_tuple(inv_y, inv_y_vec);
  }

  d_dx_functor = [&](auto&& val, auto &&adj, auto&& shared_args_tuple, auto&& x, auto&& y) {
    double inv_y = std::get<0>(shared_args_tuple);
    ...
  }
```

With the `function_gradients` call then being:
```cpp
  return function_gradients<true>(
      std::forward_as_tuple(x, y),
      std::move(value_functor),
      std::move(shared_values_functor),
      std::forward_as_tuple(d_dx_functor, d_dy_functor);
```

The `shared_values` functor would then get evaluated for any `fwd` and `rev` inputs and the resulting tuple passed to the gradient functors

andrjohns:
A much cleaner solution would be to have a single functor for the gradients which returns a tuple of values for each input, since that would allow for much more direct use of shared arguments. But I can't think of a way to expose that to users and still allow for the gradients to only be calculated for non-constant inputs

bob-carpenter:
I've been thinking about this for 8 years and still haven't been able to figure out how to do it efficiently and painlessly (i.e., without the combinatorics over which arguments are not primitive).  

Passing an intermediate value structure could work if you could figure out how to do it generically.

designs/0033-gradients-framework.md
```diff
@@ -0,0 +1,232 @@
...
+
+  return to_fvar(rtn, d_);
+```
+
+Where `internal::initialize_grad` is used to initialise a variable with the same type and dimension as the return value.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This does appear to increase compilation time - given the increased complexity of templating and tuple manipulation
```
SteveBronder, line 217: Not sure if this is a drawback but one thing we need to be careful of is still being able to share computation in this scheme between gradient calculations. I think this framework is fine if they both share calculations based on values. But if the gradients share computations based on the return adjoint then idt we would be able to share that between the different reverse pass functors

designs/0033-gradients-framework.md
```diff
@@ -0,0 +1,232 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The best alternative is the current approach, of separate specifications for `prim`, `fwd`, and `rev`. But this has the limitations mentioned in the Motivation section.
+
+# Prior art
+[prior-art]: #prior-art
+
+None that I'm aware of
```
SteveBronder, line 227: This _kind of_ reminds me of some of the things in `FastAD`

https://github.com/JamesYang007/FastAD

designs/0033-gradients-framework.md
```diff
@@ -0,0 +1,232 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is an increase in compilation times likely to be a blocker for users?
```
SteveBronder, line 232: Benchmarks for something like matmul would be nice

PR #48, opened 2022-06-29
Refactor inference algorithm outputs, allowing for multiple, cleanly structured output files

mitzimorris:
link to rendered doc:  https://github.com/stan-dev/design-docs/blob/stan-output-formats/designs/0032-stan-output-formats.md

This is the same PR as #44, but I have renamed the branch "stan-output-formats" to make it clear that this design doc is about the Stan output formats.


mitzimorris:
I have made changes per comments here:  https://github.com/stan-dev/design-docs/pull/44#pullrequestreview-978908332

WardBrian (approved):
Sorry for missing the notification on this.

I think this is really good. The remaining questions I have are things I'd like to see answered in the coding process, not in prose[^1], so I feel comfortable approving this. Obviously there will continue to be a lot of back and forth during implementation, but I think the core ideas laid out here are a good way forward. Thanks for taking all the time to think and write this out.

[^1]: For example, I wonder if the existing callbacks / `services` signatures can be re-written as small shims over the newer design, allowing us to avoid code duplication. This is an important detail, but it is also something that I don't think is critical or able to be accurately predicted at this stage

PR #51, opened 2023-11-23
Rust interface for CmdStan

andrewjradcliffe:
N.B. no direct interface with C++!

This is a proposal for a CmdStan interface for the Rust programming language, analogous to the CmdStanPy et al. interfaces. There is one distinction, which some may consider a defect (critical input actively solicited): a strongly-typed representation of the CmdStan syntax tree.


designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+
+The objective is to keep the interface as simple as possible.
+
+# Motivation
+[motivation]: #motivation
+
+Suppose that you write Rust code. Suppose that you use `Stan` for
+probabilistic programming. You have three choices for creating an
+application which utilizes both: shell scripts, introduce a scripting
+language (e.g. Python) dependency, or a Rust interface for `CmdStan`.
```
WardBrian, line 28: There are a few more exotic options as well:

- A (hypothetical) Rust interface to the `stan/services` APIs
- Using Rust implementations of algorithms like NUTS and `bridgestan`, as is done in e.g. [`nutpie`](https://github.com/pymc-devs/nutpie) 

These may be worth mentioning in the Rationale and Alternatives, as both have their own downsides as well.

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+orchestrate processes, interact with file system and handle
+cross-platform concerns. This will yield a library which is portable,
+provided that it is (cross-)compiled for the intended target.
+
+## Control: compilation and calling the resultant executable
+
+A `CmdStanModel` type will serve as an abstraction for a Stan program,
+which may need to be compiled. Rather than compile on construction, a
+compilation method must be explicitly called in user code (assuming
+that a satisfactory executable does not exist yet).
```
WardBrian, line 69: This model is something CmdStanPy [is moving away from](https://github.com/stan-dev/cmdstanpy/issues/695), because a CmdStanModel is pretty useless without an executable. 

In a strongly typed context, one could imagine that this type could instead be called something like `StanProgram`, and the result of calling `compile` is a `CmdStanModel` which is always associated with a executable on disk - this is sometimes called the "[typestate](http://cliffle.com/blog/rust-typestate/)" pattern

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+
+value   -> number | path
+```
+Where the productions for `number` and `path` are left out for
+brevity.  The start symbol is `tree`. Generate the command line
+statement by folding the tree from left to right by generating the
+appropriate term from each node, building up a linear argument list.
+I sketched
+[this](https://github.com/andrewjradcliffe/cmdstan-translator/blob/main/translate.scm)
+out in Scheme, why I am not sure.
```
WardBrian, line 148: I'm not sure this section is necessary, providing an example of use and/or a list of expected structs may be more useful

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+use, but has some incompatibilities (Rust tuple is represented as an
+array, rather than an object).
+
+Moreover, to represent matrices, complex numbers, etc., one would need
+to support types from the crate ecosystem since the standard library
+lacks these -- [nalgebra](https://github.com/dimforge/nalgebra) and
+[num-complex](https://github.com/rust-num/num-complex) are reasonable
+choices, but nonetheless represent decisions to be made!
+
+From a design perspective, this is a great place to defer to the user,
```
WardBrian, line 220: For ease of implementation this is true, but the list of above subtleties (to me) make a loud argument in the opposite direction: this is something _hard_, and the obvious approach will easily lead to incorrect results, so it should be handled by the library (eventually?)

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+limit ourselves to `std` library types). However, one needs to extract
+the variables from each line, thus, one needs to know the types (and
+their dimensions). A recursive definition of types using `enum`s and
+`struct`s could probably work to represent such a thing in Rust, but
+may not necessarily be particularly ergonomic (i.e. much unavoidable
+boilerplate would be needed to use the resultant type).
+
+Procedural macros, applied to a Stan program stored in a string
+literal in a Rust program, could be used to generate types and a
+parser for Stan CSVs produced said program. However, in order to
```
WardBrian, line 237: There is a middle-ground between this (very cool, but tricky) idea and doing nothing, which is providing some primitives (which may themselves be macros) that allow the user to provide the information required about types and names of variables. Worth considering, and it seems like any future automation could build on it.

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Representing CmdStan arguments/options as a concrete syntax tree is
+potentially brittle. If the CmdStan grammar undergoes radical change,
+this interface will need to change accordingly. However, the CmdStan
+grammar is intended to be quite stable. Moreover, it is not
+necessarily the case that radical changes to the CmdStan grammar could
+be hidden behind something other abstraction.
```
WardBrian, line 255: Since breaking changes are discussed here, it seems like a reasonable place to ask about versioning: I assume the goal is to emulate cmdstanpy/r and be "mostly" cmdstan-version-independent? Or would a given version of this crate be specialized for a specific cmdstan version?

Version independence is nice in theory but IMO has some practical concerns which may lead one to reconsider the decisions make when working on cmdstanpy/r, especially when it comes to static checking. If cmdstan adds a new argument, supporting older versions simultaneously now requires the builder to do some runtime checks on top of what the Rust type system provided you already, which is a bummer. 

Alternatively, the argument structs could themselves be versioned... but that seems like a pain.

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Other than the direct representation of the CmdStan syntax tree, the
+proposal contains nothing new. Utilizing a concrete representation of
+the syntax tree does have benefits:
+- all outputs handled via single type `CmdStanOutput`
+- elimination of individual structures and methods for each of Stan's
+inference algorithms
```
WardBrian, line 265: I think eliminating these is probably a good idea, but maybe not one whose time has come. At the moment, the output files of e.g. ADVI should _not_ be treated the same as those produced by sampling, because the first row is the mean of the approximation, not a draw. Therefore, something like picking a random row for initialization or even passing to `stansummary` is subtly different depending on algorithm.

One hopes that cmdstan will eventually have some nicer output conventions, such as putting the mean in its own file, but in the mean time the type safety of preventing you from doing something foolish could be nice.

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+As described above, the grammar for CmdStan arguments passed at the
+command line can be represented as a syntax tree through the use of
+sum and product types.
+- This enables compile-time validation of argument consistency -- the
+  worst that can happen is you provide a value that CmdStan does not
+  like (e.g. `num_threads=-20`).
+- At minimum, this will move a variety of run-time errors to compile
+  time; it might even help users to understand the methods and options
+  CmdStan provides.
+- This enables re-use of a parameterized argument tree -- one could
```
WardBrian, line 280: [comment] This is a very nice plus, IMO.

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+  replace the inference method while leaving the other options
+  (e.g. data files) constant. As shown in [this
+  example](https://github.com/andrewjradcliffe/cmdstan-rs/blob/main/examples/bernoulli-many/main.rs),
+  such an approach can be quite expressive.
+
+Furthermore, representation as a concrete syntax tree enables the
+possibility of interesting features. One could parse the syntax tree
+from:
+- a string written to a log file
+- a string which is consistent with the grammar that CmdStan accepts
```
WardBrian, line 290: Or, one hopes, the [JSON file cmdstan itself can now produce](https://github.com/stan-dev/cmdstan/pull/1204)

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+
+The design philosophy here would be: a valid parse is whatever CmdStan
+is willing to accept. However, CmdStan accepts some weird statements.
+For example:
+```bash
+./bernoulli method=sample adapt engaged engaged=0 engaged engaged=1 gamma engaged gamma \
+    data file=bernoulli.data.json
+```
+
+The proposal is to use [pest](https://github.com/pest-parser/pest),
```
WardBrian, line 317: One thing I am curious about is total dependencies (it may be worth calling this out as its own list/table somewhere). 

Both Python and R's wrappers at least strive to be minimal, within what their standard libraries provide. I think the wonder of `cargo` making dependencies on the whole less painful has a downside of them tending to proliferate. 

Could items like this be hidden behind a cargo feature? Continuing the only-pay-for-what-you-want mentality from earlier

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+the call. This includes:
+- the console output (exit status, stdout and stderr),
+- the argument tree provided to `call_executable`
+- the current working directory of the process at the time the call was made
+- the CmdStan installation from the parent `CmdStanModel`
+
+The objective is for `CmdStanOutput` to be a self-contained record
+which includes all pertinent information from a successful executable
+call. This structure can then be used to direct calls of
+`diagnose`/`stansummary`. Naturally, methods on said type will be
```
WardBrian, line 98: Can you supply a complete list of signatures for methods on `CmdStanOutput`? You mention this path helper (very nice to provide, somewhat tricky to get right!) and `stansummary`, but what would their signatures look like, and are there more? I'm sure some of this will need to be discovered during implementation

designs/0035-cmdstanrs.md
```diff
@@ -0,0 +1,348 @@
...
+compilation method must be explicitly called in user code (assuming
+that a satisfactory executable does not exist yet).
+
+Two arguments will be necessary to create a `CmdStanModel`:
+1. a path to a CmdStan installation
+2. a path to a Stan program
+
+Methods (receiver: `CmdStanModel` instance) exposed to the user will
+include:
+- `validate_cmdstan` : determine whether the CmdStan installation works
```
WardBrian, line 77: Both this and passing the path to the CmdStan installation feel like they're concerns separate from a CmdStanModel instance to some extent.

Does it make sense to have a `CmdStanInstallation` struct which stores a path and validation logic? I'm not super familiar with what is idiomatic, but could there even be a lazy_once global which is used by default, and an overridden constructor if the user needs to supply their own or two separate instances in the same program?

designs/0035-cmdstanrs.md
WardBrian: Things like MSRV may also be worth mentioning!

WardBrian (commented):
Thank you for putting this together. I left some thoughts/comments on individual points below, but I think the overall structure laid out here is a good direction for a rust interface to CmdStan. In particular, I at least am sold on the approach to the argument structure

SteveBronder:
Kind of adjacent to this proposal, for a compiled language would something like a foreign function interface scheme be more appropriate? Using something like [ffistan](https://github.com/WardBrian/ffistan) the rust interface would not need to compile a separate binary

SteveBronder:
@WardBrian Does ffistan not use Stan's writer class? Then folks could just use the writer to write to disk. If it only takes in a pointer to an array to write to I think you could still use something like `mmap` on linux to write to disk

WardBrian:
@SteveBronder You could do it with mmap

designs/0035-cmdstanrs.md
andrewjradcliffe: Excellent idea. As is often the case, putting a name to a common idiom (the typestate pattern) is illuminating.

designs/0035-cmdstanrs.md
andrewjradcliffe: Alas, yes, it should be handled by the library. It could be gated by a feature, so as to delay the need for an implementation for the initial version. As additional dependencies would likely be introduced, a feature gate is probably appropriate anyway.

designs/0035-cmdstanrs.md
andrewjradcliffe: Technically, quite a few (maybe all?) of the structs and enums would need to be marked with `#[non_exhaustive]`; I was anticipating as much with the builder pattern.

If the Rust code stays up-to-date with the latest cmdstan arguments and options, then the worst case is that an older cmdstan will return something to the effect of "... is not a valid value for ..." in the stderr. An output is already not a guarantee (the process could die for any reason), thus, rather than leaving it to the user, we could provide a course classification of the errors (insufficient cmdstan version to support the user's argument tree being a variant).

I do not see an easy way to raise cmdstan capability checks to compile-time without resorting to strange things in a build script or procedural macro antics. Even then, the cmdstan installation that a user intends to use must be available at compile time.

Alternatively, one could package cmdstan as a Rust crate and manage it as a dependency with cargo. I do not recommend this due to the other headaches it would create.

designs/0035-cmdstanrs.md
andrewjradcliffe: My thought was that with the argument tree cached in the output, special handling could be implemented there (e.g. `match` on the method). A reference to the cached value would be available to the user, thus enabling the same `match`-based approach for their downstream code.

On the other hand, it would be possible to move the common functionality to a trait and have one output type for each method; this would enable generic code for `T: Output`, but would likely require a `Box<dyn Output>`. 

Another enum, mirroring the method variants in the argument tree, would likely be easier to reason about (rather than a boxed trait object). But if you have to `match` on output variant, it's the same in effect to `match` on the input variant.

designs/0035-cmdstanrs.md
andrewjradcliffe: I would also like to support reading from JSON; that should be relatively easy.

I got interested in PEGs, went a little overboard, and now there's a parser which is consistent with CmdStan; it supports a [variety](https://github.com/andrewjradcliffe/cmdstan-rs/blob/main/examples/argtree-from/main.rs) of input formats.

designs/0035-cmdstanrs.md
andrewjradcliffe: Agreed. The path to the cmdstan installation going through `CmdStanModel` is definitely an undesirable thing. Separating into something like

```rust
#[derive(Clone)]
pub struct CmdStanInstallation { path: PathBuf }
pub struct StanProgram { path: PathBuf }

impl CmdStanInstallation {
    fn validate(&self) -> io::Result<bool>; // you can break the install by deleting it
    pub fn compile(&self, prog: &StanProgram) -> Result<CmdStanModel, CompilationError>;
    pub fn try_from_env() -> io::Result<Self>; // we look for `CMDSTAN` in environment
}

impl TryFrom<&Path> for CmdStanInstallation { /* validate CmdStan works, else error */ }
```

Using an environment variable is a reasonably consistent way to set a global without committing to a lazily-initialized global in Rust. Admittedly, this just moves the problem elsewhere, but it is clear what you would be getting -- the current CmdStan, as determined by a specific environment variable. 

I am not fond of user-facing globals (unless they truly are just static data, e.g. [color schemes](https://crates.io/crates/colorous)). If the global depends on something a user is to provide, it gets messy. If we provide an interface that requires a validated CmdStan installation, then it's up to the user to figure out how to meet said requirement. I don't see much harm placing the complexity on the user -- after all, it becomes our problem if it is ambiguous as to which CmdStan should be used at any given point.

If a user really wanted to do it via global, the code below is an option.
```rust
static CMDSTAN: once_cell::sync::Lazy<PathBuf> = once_cell::sync::Lazy::new(|| infallible_go_get_cmdstan());

fn my_cmdstan_install() -> CmdStanInstallation {
    let path: &Path = &*CMDSTAN;
    CmdStanInstallation::try_from(path).unwrap()
}
```

designs/0035-cmdstanrs.md
andrewjradcliffe: The gate here is: decide on conventions for representation of Stan types as Rust types. `nalgebra` is quite similar to Eigen, including memory layout, so matrices are covered.

For arrays of dimension greater than 2, there is a problem.
- Yes, there is `ndarray`, but its [`future is uncertain`](https://github.com/rust-ndarray/ndarray/issues/1272). 
- Nested `Vec<Vec<Vec ... >>>` naturally works, but I doubt users would like it (no syntactic sugar for indexing).

If we take the latter option (the only sensible course for stable development), then it would be relatively straightforward to use a procedural macro to derive a parser for a user-provided struct where the field names and respective types must match the Stan CSV output. I do something similar for my use cases.

designs/0035-cmdstanrs.md
WardBrian: That seems reasonable! Saying as much in the document would be appreciated

designs/0035-cmdstanrs.md
WardBrian: I don't have any objections to `thiserror` - it seems quite common in practice and rather small in terms of scope.

You bring up an interesting other point: I (sort of implicitly) expected that this wrapper could be written using entirely safe Rust. Is that `unsafe` function really needed, or is it called as an optimization?

What other usages of `unsafe` would you anticipate, and why?

WardBrian:
I think there could definitely be code-reuse between something in Rust that shells out to cmdstan and something that does it through CFFI. From a UX standpoint you still have the same issue with having a _lot_ of arguments and no good way of doing defaults/named arguments except through a builder-like pattern.

designs/0035-cmdstanrs.md
andrewjradcliffe: The single unsafe function arose from the need to split the string on the occurrence of `'.'`, with the objective of replicating CmdStan's interesting treatment of paths. This could easily be eliminated if we weren't dealing with Os-specific details (hidden behind [WTF-8](https://en.wikipedia.org/wiki/UTF-8#WTF-8) in OsString).

CmdStan basically just splits at the first occurrence of `'.'`, starting from the right. 
This becomes a problem when the only thing to the left of the first `'.'` is the empty string and there the input matches any of the forms below.

```bash
./bernoulli id=2 method=sample num_chains=3 data file=bernoulli.data.json output file='..'
# produces as output file names: `._2.`, `._3.`, `._4.`

./bernoulli id=2 method=sample num_chains=3 data file=bernoulli.data.json output file=.foo
# produces as output file names: `_2.foo`, `_3.foo`, `_4.foo`

# this is actually a repeat of the above, with "foo" as the empty string
./bernoulli id=2 method=sample num_chains=3 data file=bernoulli.data.json output file='.'
# produces as output file names: `_2.`, `_3.`, `_4.`
```

It is not possible to easily treat these by two edge cases due to the [definition of an extension](https://doc.rust-lang.org/std/path/struct.Path.html#method.extension).
I might be able to come up with another way to do it, but it will complicate the code accordingly. 

I see no issue with `unsafe` which is limited to a very specific use case that is fundamentally sound. This use is particularly easy to verify soundness: the `&[u8]` slices are created by [`as_encoded_bytes`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.as_encoded_bytes) and promptly consumed; when the byte slices are re-joined, it is impossible for either to have become invalid UTF-8.

It is worth noting that `OsStr::from_encoded_bytes_unchecked` only appears in two places; both appearances involve inputs which are created within the same function body, that is, no foreign inputs, no intermediates escape the scope in which they are created and said intermediates are consumed in the same function body. Incidentally, it is more efficient to play with byte slices, but that was not my objective.

Each use looks like:
```rust
let bytes = path.as_encoded_bytes();
let mut iter = bytes.rsplitn(2, |b| *b == b'.');
let (prefix, suffix) = match (iter.next(), iter.next()) {
    (Some(suffix), Some(prefix)) => {
        // SAFETY:
        // - each fragment only contains content that originated
        //   from `OsStr::as_encoded_bytes`.
        // - split with ASCII period, which is a non-empty UTF-8
        //   substring
        unsafe {
            (
                OsStr::from_encoded_bytes_unchecked(prefix),
                OsStr::from_encoded_bytes_unchecked(suffix),
            )
        }
    }
    _ => (path, "csv".as_ref()),
};
```

I general, my philosophy on use of `unsafe` is that it is best to avoid where possible; for a crate like this, I certainly wouldn't expose any `unsafe` public APIs. There's no reason to expose an unsafe internal API for this crate, as there is nothing that warrants it here.

I realize that [my response above](https://github.com/stan-dev/design-docs/pull/51#discussion_r1413267023) could actually be read as somewhat alarming -- potentially implies lots of `unsafe`. Limited to two locations ([1](https://github.com/andrewjradcliffe/cmdstan-rs/blob/42a4e0be0b6669096a81bcbc514878beb4e57cb5/src/argument_tree.rs#L76), [2](https://github.com/andrewjradcliffe/cmdstan-rs/blob/42a4e0be0b6669096a81bcbc514878beb4e57cb5/src/argument_tree.rs#L177C13-L177C13)) with no plans to expand beyond these, I hope that the above is reassuring.

WardBrian: This also sounds like something we will probably change (https://github.com/stan-dev/cmdstan/issues/1213) - the current system is mostly designed-by-accident

nhuurre: 1. Your use of `unsafe` seem fine to me. However, since essentially the same unsafe code is used in two places I'd encapsulate it in a single safe-to-call helper function.
2. This file extension handling corner case feels like a CmdStan bug and I wouldn't bother supporting it beyond a clear error message.
3. It is quite frustrating that `as_/from_encoded_bytes` does not provide a _safe_ API for splitting `OsStr`s. Nevertheless, in this case `unsafe` can be avoided with a couple of allocations:
```rust
use std::ffi::{OsString, OsStr};
let (prefix, suffix) = match (path.file_stem(), path.extension()) {
    (Some(prefix), Some(suffix)) => (prefix, suffix.to_os_string()),
    (Some(name), None) if name.as_encoded_bytes().first() == Some(&b'.') => {
        let mut fake_name = std::path::PathBuf::from("x");
        fake_name.as_mut_os_string().push(name);
        (OsStr::new(""), fake_name.extension().unwrap().to_os_string())
    }
    (Some(name), None) =>
        (name, OsString::from("csv")),
    (None, _) => {
        // No file name, which means the last path component is either "." or ".."
        if path.as_os_str().as_encoded_bytes().ends_with(b"..") {
            (OsStr::new("."), OsString::from(""))
        } else {
            (OsStr::new(""), OsString::from(""))
        }
    }
};
```
4. Coming back to point (2), if the user specifies `output file=out.d/file` CmdStan tries to write to `out_1.d/file` _etc_ but if the directory `out_1.d` does not exist CmdnStan won't create it and the output just silently disappears. That can't be right.

andrewjradcliffe: In order:
1. I completely agree. It should be moved to a single function.
2. I agree again. I found this when I was writing tests for the Rust code. However, if CmdStan does it, it seems fine to support it if it is easy enough. Splitting WTF-8 encoded bytes on ASCII is fundamentally sound, so the `unsafe`-ness doesn't bother me. In fact, `std` uses the [same approach](https://doc.rust-lang.org/src/std/path.rs.html#318); I had not looked at `std` until now.
3. It is only the conversion back to `OsStr` that is `unsafe`, but I agree that a safe API would be nice. Unfortunately, literal split on the first period from the right is very hard to avoid. Your approach is clever, but is defeated by inputs such as below:
```bash
function bernoulli() {
    ./bernoulli data file=bernoulli.data.json sample num_chains=3 output file=$1
}

mkdir -p foo/bar
                            # files created by CmdStan
bernoulli 'foo/.bar'        # "foo/_1.bar", "foo/_2.bar", "foo/_3.bar"
bernoulli 'foo/bar/'        # "foo/bar/_1.csv", "foo/bar/_2.csv", "foo/bar/_3.csv"
bernoulli 'foo/bar/.'       # "foo/bar/_1.", "foo/bar/_2.", "foo/bar/_3."
bernoulli 'foo/bar/..'      # "foo/bar/._2.", "foo/bar/._3.", "foo/bar/._4."
bernoulli 'foo/bar/...'     # "foo/bar/.._2.", "foo/bar/.._3.", "foo/bar/.._4."
```
One might think that `path::parent` and `path::file_name` would reduce the problem to the code you gave, substituting the result of `path::file_name`, but that does not work due to path normalization. For example, `foo/bar/.` is equivalent to `foo/bar` when treated as a path; likewise for `foo/bar/`.

From a code complexity standpoint, splitting at the `'.'` byte is much simpler, as it does not require addition of more case logic, and matches what CmdStan does by default. 

This is a good example of where `unsafe` is both easier to maintain and verify the correctness of. Using the `std::path` API quickly becomes cumbersome, as it performs the normalization of path components one would naturally expect; in essence, the logic of the program must now test for all the ways that a string with strange placements of `'.'` would be subject to path normalization, then avoid them.

4. This was sufficiently interesting that I wrote it as an [issue](https://github.com/stan-dev/cmdstan/issues/1226) for the CmdStan repository.

----
### Summary

Until CmdStan switches from splitting on the last dot, there is no reason to deviate from said behavior in the Rust treatment of paths. Path normalization is not trivial (see [cppreference](https://en.cppreference.com/w/cpp/filesystem/path) for a nice recapitulation of the steps). Effort spent on attempting to reproduce the CmdStan behavior via a convoluted use of `Path` and `OsStr`, on the basis that no code should be `unsafe`, is likely to lead to logical errors-- I am not keen on working out all pathological (no pun intended) behavior induced by lack of path normalization and splitting on the last `'.'`. It is much simpler and maintainable to verify that the invariants of an `unsafe` function are satisfied than to create a [Rube Goldberg machine](https://en.wikipedia.org/wiki/Rube_Goldberg_machine).

Examples like [Case 2](https://github.com/stan-dev/cmdstan/issues/1226) demonstrate that such a machine would not be feasible, as it is not sufficient to split parent and child, check the respective start/end, and be done. One must scan the entire string for dots and this is only possible after `as_bytes_encoded`, hence, we are back to `unsafe` to put the halves back together (i.e. right back to the original proposal).

Furthermore, the discrepancy between CmdStan's behavior and the proposed interface would be a a potential source of confusion. If someone wants to use wacky file names and the program (CmdStan) accepts them, we would be remiss to impose artificial limitations.

PR #52, opened 2024-01-31
Add a `jacobian` keyword to the Stan language

SteveBronder:
Proposes adding a  `jacobian` keyword and a new `constraints` block to the Stan language

Rendered Version Here:
https://github.com/stan-dev/design-docs/blob/feature/jacobian/designs/0034-jacobian.md

### DIFF
--- a/designs/0034-jacobian.md
+++ b/designs/0034-jacobian.md
@@ -4,62 +4,30 @@
 # Summary
 [summary]: #summary
 
-This design doc proposes adding a `jacobian` target and a new block for user defined constraints. The `jacobian` target will be accessible directly in `transformed parameters` and a new `constraints` block where users can define custom constraints similar to Stan's already existing constrained data types. 
+This design doc proposes adding a `jacobian` target and `*_jacobian` user defined functions. The `jacobian` target will be accessible directly in `transformed parameters` and the new `*_jacobian` functions.
+
+
 The examples below show an example use case.
 
 ```stan
-constraints {
- upper_bound {
-   // Constrain x to have an upper bound
-   real constrain(real x, real upper_bound) {
-     jacobian += x;
-     return upper_bound - exp(x);
-   }
-   // Constrain a vector x to have an upper bound
-   real constrain(vector x, real upper_bound) {
-     jacobian += sum(x);
-     return upper_bound - exp(x);
-   }
-   // Unconstrain real x from the upper bound
-   real unconstrain(real x, real upper_bound) {
-     return log(upper_bound - x);
-   }
-   // Unconstrain vector x from the upper bound
-   vector unconstrain(vector x, real upper_bound) {
-     return log(upper_bound - x);
-   }
-   // Validate x has an upper bound
-   int validate(real x, real upper_bound) {
-     return x < upper_bound;
-   }
- }
+functions {
+  real upper_bound_jacobian(real x, real upper_bound) {
+    jacobian += x;
+    return upper_bound - exp(x);
+  }
 }
 data {
  real ub;
- int<lower=0> N;
 }
 parameters {
  // User defined constraint
- real<constraint upper_bound(ub)> b;
- vector<constraint upper_bound(ub)>[N] b_vec;
+ real b_raw;
+}
+transformed paramters {
+  real b = upper_bound_jacobian(b_raw, ub);
 }
 ```
 
-Users can also access `jacobian` directly in the `transformed parameters` block.
-
-```stan
-data {
- real lb;
-}
-parameters {
- real c_raw;
-}
-transformed parameters {
- // Transform and accumulate jacobian
- real c = exp(c_raw) + lb;
- jacobian += c_raw;
-}
-```
 # Motivation
 [motivation]: #motivation
 
@@ -86,189 +54,72 @@ Having the Stan language define types for transforms from the unconstrained to c
 This is very nice for both parties since it is both of their preferred spaces to work in.
 
 Most of the time this encapsulation is very good. 
-The main issue is that Stan users either have to write code that either only uses Stan's built-in transforms to stay in the constrained space or works directly on the unconstrained space. 
-For instance, from a comment [2] by @betanalpha (slightly commented and fleshed out)
+The main issue is that Stan users either have to write code that either only uses Stan's built-in transforms to stay in the constrained space or have to always have the jacobian calcuations enabled while working in the constrained space.
 
------------
 
-> It’s only in the context of trying to assign a density on the output value that a Jacobian is needed, so the proper encapsulation is
 
-```stan
-functions {
- real f(real x) {
-   // transform function
- }
- real jacobian(real x) {
-   // jacobian determinant calculation
- }
- real prob(real x) {
-   // lpdf calculation
- }
-}
-parameters {
- real theta;
-}
-transformed parameters {
- real eta = f(theta);
-}
-model {
- // In a pure sense, neither line is well defined on it's
- //  own, but the combinations are well defined.
- eta ~ prob(values);
- target += jacobian(theta);
-}
-```
-
------------
-
-Coding transforms directly across multiple blocks (parameter, transformed parameter, model) is clunky and error prone. 
-And when users want to add a constraint to the Stan language they have to touch the math library which is rather complicated. 
-Not just by C++ itself, but the essentially small DSL we use to describe autodiff in the Stan math library.
-
-Having a new block for user-defined constraints allows users to encapsulate and reuse their transforms as units.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
 The following section is a draft of the docs for the new constraint and `jacobian` keyword.
 
-## Constraint Block
-
-### Overview
-
-The constraints block was included in Stan 2.XX to provide users with the ability to define custom constraints on parameters, transformed parameters, and data. 
-This block is situated immediately after the `functions` block and before the `data` block in a Stan program. 
-The primary purpose of the `constraints` block is to allow users to define custom constraints that can be used in the same way as built-in constraints.
-
-### Syntax
-
-A constraints block is defined by the keyword `constraints`. 
-Inside of the `constraints` block each constraint is its own block with a given `constraint_name` and the functions needed to define the constraint.
-Inside the `constraints` block, users can define one or more constraint types. Each constraint type is encapsulated within its own set of curly braces and is identified by a unique name.
-
-Each constraint type must specify a set of functions that define the transformation between unconstrained and constrained spaces, validate data, and increment the log of the absolute value of the determinant of the Jacobian. 
-The first argument for all functions defined in a constraint must be the argument of interest to be transformed or validated. These functions include:
-
-- `constrain(TransformType, OtherArguments...) -> ReturnType`: Transforms a real unconstrained variable to a constrained space.
-- `unconstrain(TransformType, OtherArguments...) -> ReturnType`: Transforms a real constrained variable back to the unconstrained space.
-- `validate(TransformType, OtherArguments...) -> ReturnType`: Validates a real variable against the constraint, returning 1 if valid and 0 otherwise.
-
-The pseudocode below shows the structure of a user defined constrained.
-All CamelCase names are to be replaced by the user.
-
-- `ConstraintName`: The name of the constraint which will be called by users from within the `<>` of Stan's [data types](https://mc-stan.org/docs/reference-manual/overview-of-data-types.html).
-- `ReturnType`: A Stan data type that is the return type of the function
-- `UnconstrainType`: A Stan data type that is the variable to be transformed or validated.
-- `OtherArgs` The rest of the Stan data types that are passed to the constraint during construction of the data type that is being constrained.
-
-
-```stan
-constraints {
- ConstraintName {
-   // Transform from unconstrained to constrained space
-   ReturnType constrain(UnconstrainType, OtherArgs...) {
-     // ... transformation logic ...
-     // Increment Jacobian
-     jacobian += ...;
-     return ...;
-   }
-
-   // Transform from constrained to unconstrained space
-   ReturnType unconstrain(UnconstrainType, OtherArgs...) {
-     // ... inverse transformation logic ...
-     return ...;
-   }
-
-   // Validate data
-   int validate(UnconstrainType, OtherArgs...) {
-     // ... validation logic ...
-     return ...; // typically returns a boolean value
-   }
- }
-}
-```
-
 ### Using the jacobian Keyword
 
-The `jacobian` keyword is introduced within the `constraint` block to allow the Jacobian accumulator to be incremented by the log absolute determinant of the Jacobian of the constraining transform. 
+The `jacobian` keyword is introduced within the `transformed parameters` block and the functions ending in `_jacobian` to allow the Jacobian accumulator to be incremented by the log absolute determinant of the Jacobian of the constraining transform. 
 This keyword behaves similarly to the `target` keyword, allowing users to account for the change of variables in the probability density function. 
-The `jacobian` keyword is available in both the `constraints` block and the `transformed parameters` block.
+The key difference between the two is that `jacobian` will only be accumulated into for algorithms that request it. Algorithms such as maximum likelihood would request the jacobian to not be accumulated. 
+The `jacobian` keyword is available only in the `transformed parameters` block and within functions ending in `_jacobian`.
 
-### Declaring Variables with User-defined Constraints
-
-Variables can be declared with user-defined constraints in the data, transformed data, parameters, transformed parameters, and generated quantities blocks. 
-The syntax for declaring a variable with a user-defined constraint is as follows:
-
-```stan
-type<constraint constraint_name(args...)> variable_name;
-```
-
-Here, `constraint` denotes a user-defined constraint and `constraint_name(args...)` specifies the name of the constraint along with any arguments it requires.
 
 ### Example Usage
 
+The following stan program defines an `upper_bound_jacobian` function for constrainting `real` and `vector` types.
+
 ```stan
-constraints {
- upper_bound {
-   real constrain(real x, real upper_bound) {
-     jacobian += x;
-     return upper_bound - exp(x);
-   }
-   real unconstrain(real x, real upper_bound) {
-     return log(upper_bound - x);
-   }
-   vector constrain(vector x, real upper_bound) {
-     jacobian += sum(x);
-     return upper_bound - exp(x);
-   }
-   vector unconstrain(vector x, real upper_bound) {
-     return log(upper_bound - x);
-   }
-   int validate(real x, real upper_bound) {
-     return x < upper_bound;
-   }
- }
+functions {
+real upper_bound_jacobian(real x, real ub) {
+  jacobian += x;
+  return upper_bound - exp(x);
 }
+vector upper_bound_jacobian(vector x, real ub) {
+  jacobian += x;
+  return upper_bound - exp(x);
+}
+}
+
 data {
- real a;
- real u_bound;
+ real ub;
  int N;
 }
 
-
-transformed data {
- // Data will use the validate() function from the user-defined constraint
- real<constraint upper_bound(u_bound)> c = a;
+parameters {
+  real b_raw;
+  vector[N] b_vec_raw;
 }
 
+transformed parameters {
+  real b = upper_bound_jacobian(b_raw, ub)
+  vector[N] b = upper_bound_jacobian(b_vec_raw, ub)
+}
+```
+
+The following stan program calculates the same constraint and jacobian directly in the transformed parameters block.
+
+```stan
+data {
+ real ub;
+ int N;
+}
 
 parameters {
- // User-defined constraint applied to parameters alpha and b
- real<constraint upper_bound(u_bound)> alpha;
- vector<constraint upper_bound(u_bound)>[N] b;
- // In these cases this is the same as calling
- // real<upper=u_bound> alpha;
- // vector<upper=u_bound>[N] b;
+  real b_raw;
 }
-```
 
-In this example, `upper_bound` is a user-defined constraint that is applied to the data and parameters.
-The `validate()` function from `upper_bound` is used to validate `a_validated` in the `transformed data` block, and the transformations defined in `upper_bound` are applied to `alpha` and `b` in the parameters block.
-
-### Error Messages
-
-Inside of a constraint, if a `constrain` function is defined then an `unconstrain` function must also be defined with the same signature as `constrain`.
-
-```
-constraint(vector x, real upper_bound) for {CONSTRAINT_NAME} does not have an associated `unconstrain(vector x, real upper_bound)`. If you define one then you must define both!
-```
-
-If a user defined constraint type is used in `data`, `transformed data`, or `transformed parameters`, or `generated quantities` then the constraint must have a validate function for that given type. 
-For example, if a user calls a constraint named `lower_bound` tha does not have a `validate()` function then the error message will look like:
-
-```
-// vector<constrain lower_bound(lb)> x;
-a `lower_bound` constraint was given to `x`, but `lower_bound` does not have a validate(vector, real) function.
+transformed parameters {
+  jacobian += b_raw;
+  real b = upper_bound - exp(b_raw);
+}
 ```
 
 
@@ -277,59 +128,8 @@ a `lower_bound` constraint was given to `x`, but `lower_bound` does not have a v
 
 The first thing we would do is add the `jacobian` keyword to the Stan language. 
 Like `target`, `jacobian` would only be available on the left hand side of statements to be accumulated into. 
-Its domain is restricted to the `transformed parameters` and `constraints` block.
+Its domain is restricted to the `transformed parameters` and functions ending in `_jacobian`.
 The main difference between `jacobian` and `target` is that the jacobian accumulation is optional for Stan algorithms. If an algorithm wants to do something like maximum likelihood estimation then it is possible to turn off the Jacobian accumulation.
-
-This section will break down the model given in the summary section to show what the Stan compiler requires piece by piece. 
-First, let's look at the `constraints` block where we define our constrain and unconstrain functions
-
-## Constraint Block
-
-```stan
-constraints {
- upper_bound {
-   constrain(real x, real upper_bound) {
-     jacobian += x;
-     return upper_bound - exp(x);
-   }
-   real unconstrain(real x, real upper_bound) {
-     return log(upper_bound - x);
-   }
-   // Used for data
-   int validate(real x, real upper_bound) {
-     // For most this will be
-     return x < upper_bound;
-   }
- }
-}
-```
-
-The new `constraints` block will need to be added to the compiler's parser, validator, Middle Implementation Layer (MIR), and c++ MIR. 
-The parser will transpile each of the constraints to c++ functions which start with the name of the constraint. For the example above the signatures would be
-
-```c++
-upper_bound_constrain__(...)
-upper_bound_unconstrain__(...)
-upper_bound_validate__(...)
-```
-
-The only function of the three listed above that has rules that differ from standard function generation is the `*_constrain` function which has an additional `jacobian` argument and `Jacobian` template parameter. 
-The jacobian is kept behind an if statement that checks the compile time value `Jacobian` before incrementing the jacobian. 
-Besides the jacobian argument the rest of the function will parse exactly like a standard function.
-
-```c++
-template <bool Jacobian, typename T1, typename T2, typename TJacobian>
-return_type_t<T1, T2> upper_bound_constrain(const T1& x, const T2& upper_bound,
-  TJacobian& jacobian) {
- if (Jacobian) {
-   jacobian += x;
- }
- return subtract(upper_bound, exp(x));
-}
-```
-
-The `_unconstrain` function will not be used directly by the user and instead will be used internally in the Stan model class when we need to go from the constrained to unconstrained space. 
-The code generation for these will be the same as a standard function.
 
 ## Transformed Parameters
 
@@ -349,31 +149,26 @@ transformed parameters {
 }
 ```
 
+## Functions ending in _jacobian
+
+Functions ending in `_jacobian` will operate the same as `_lp` functions, but instead of exposing `target` they will expose `jacobian`.
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
 - Making a new keyword `jacobian` will almost surely break current user code.
 
+- Because we do not have the associated function to unconstrain the parameters users will have to initialize from parameters on the unconstrained space. 
+
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-- Why is this design the best in the space of possible designs?
-
-The other alternative is to not have a new block and have users define `*_constrain`, `*_unconstrain`, and `*_validate` functions in the `functions` block. 
-This might be easier to implement, but having raw functions makes the constraints scheme removes localization and encapsulation which I think is a lot nicer.
-
 - What is the impact of not doing this?
 
-Users will have to write programs as they currently do, where jacobian accumulations happen directly in transformed parameters or within the model block.
+Users will have to write programs as they currently do, where jacobian accumulations happen directly in transformed parameters or within the model block. Users will also not be able to turn off jacobian accumulations when using maximum likelihood like algorithms.
 
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
-
-- Is there a better syntax for users than
-
-```stan
-vector<constraint my_constrain(args...)> X;
-```
 
 # Citations
 
--- a/designs/0034-jacobian.md
+++ b/designs/0034-jacobian.md
@@ -41,7 +41,7 @@ $$
 \log\left(\pi^*(y)\right) = \log\left(\pi\left( c\left(y\right) \right)\right) + \log\left(J_c\left(y\right)\right)
 $$
 
-The Stan languages has built in constraints constraints such as `lower`, `upper`, `ordered`, etc. to handle $ \log(J_c(y))$. 
+The Stan languages has built in constraints constraints such as `lower`, `upper`, `ordered`, etc. to handle $\log(J_c(y))$. 
 A variable (unconstraining) transform is a surjective function $f:\mathcal{X} \rightarrow \mathbb{R}^N$ from a constrained subset $\mathcal{X} \subseteq \mathbb{R}^M$ onto the full space $\mathbb{R}^N$.
 The inverse transform $f^{-1}$ maps from the unconstrained space to the constrained space. 
 Let $J$ be the Jacobian of $f^{-1}$ so that $J(x) = (\nabla f^{-1})(x).$ and $|J(x)|$ is its absolute Jacobian determinant. A transform in Stan specifies 
--- a/designs/0034-jacobian.md
+++ b/designs/0034-jacobian.md
@@ -30,18 +30,20 @@ constraints {
 data {
  real lb;
  real ub;
+ int N;
 }
 parameters {
  //(2) User defined constraint
- upper_bound<ub=ub> b_raw;
+ real<udc=upper_bound(ub)> b;
+ vector<udc=upper_bound(ub)>[N] b_vec;
  real c_raw;
 }
 transformed parameters {
  // (3) Transform and accumulates like stan math directly
- real b = exp(b_raw) + lb;
+ real c = exp(c_raw) + lb;
  // Underneath the hood, only actually calculated if
  // Jacobian bool template parameter is set to true
- jacobian += b_raw;
+ jacobian += c_raw;
 }
 ```
 
@@ -51,11 +53,11 @@ transformed parameters {
 Given a function mapping $c$ that transforms a set of parameters from the unconstrained space $Y$ to the constrained space $X$, probability density function $\pi$, and a Jacobian determinant function over the constrained space $J(c(y))$, Stan calculates the log transformed density function [2]
 
 $$
-\pi^*(y) = \pi\left( c\left(y\right) \right) J(c\left(y\right)) 
+\pi^*(y) = \pi\left( c\left(y\right) \right) J(c\left(y\right))
 $$
 
 $$
-\log\left(\pi^*(y)\right) = \log\left(\pi\left( c\left(y\right) \right)\right) + \log\left(J(c\left(y\right)\right)) 
+\log\left(\pi^*(y)\right) = \log\left(\pi\left( c\left(y\right) \right)\right) + \log\left(J(c\left(y\right)\right))
 $$
 
 In a Stan program, $ \log\left(J(c\left(y\right)\right))$ is handled via the Stan languages defined constraints such as `lower`, `upper`, `ordered`, and others.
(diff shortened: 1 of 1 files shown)


designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+The constraints block was included in Stan 2.XX, introduced to provide users with the ability to define custom constraints on parameters, transformed parameters, and data. This block is situated immediately after the `functions` block and before the `data` block in a Stan program. The primary purpose of the `constraints` block is to allow users to define custom constraints that can be used in the same way as built-in constraints.
+
+### Syntax
+
+A constraints block is defined by the keyword `constraints`. Inside of the `constraints` block each constraint is its own block with a given `constraint_name` and the functions needed to define the constraint. Inside the `constraints` block, users can define one or more constraint types. Each constraint type is encapsulated within its own set of curly braces and is identified by a unique name.
+
+Each user-defined constraint within the constraint block must specify a set of functions that define the transformation between unconstrained and constrained spaces, validate data, and increment the log of the absolute value of the determinant of the Jacobian. The general structure of a user-defined constraint is as follows:
+
+```stan
```
bob-carpenter, line 117: I think you want to describe the signatures and then give a complete example in the doc.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+   int validate(TransformType, OtherArgs...) {
+     // ... validation logic ...
+     return ...; // typically returns a boolean value
+   }
+ }
+}
+```
+
+Each constraint type must define a set of functions that govern the transformation and validation of variables subject to the constraint. The first argument for all functions defined in a constraint must be the argument of interest to be transformed or validated. These functions include:
+
```
bob-carpenter, line 146: This is confusingly mixing types (`TransformType`) and arguments (`OtherArguments`).  How about putting it more in Stan syntax, as in:

```
C constrain(U x, T1 a1, ..., TN aN);
```

and say that `U` is the unconstrained type, `C` is the constrained type, and `T1...TN` are types of extra arguments.

Does the constrained type U have to be a vector?

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+### Example Usage
+
+```stan
+constraints {
+ upper_bound {
+   real constrain(real x, real upper_bound) {
+     jacobian += x;
+     return upper_bound - exp(x);
+   }
+   real constrain(vector x, real upper_bound) {
```
bob-carpenter, line 174: Return should be `vector` here.

Now it looks like you're going to define constrained containers directly rather than mapping?

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+     jacobian += sum(x);
+     return upper_bound - exp(x);
+   }
+   real unconstrain(real x, real upper_bound) {
+     return log(upper_bound - x);
+   }
+   // Used for data
```
bob-carpenter, line 181: 181 and 183 can be removed---this is self explanatory and just an example.  Same comments apply to data constraints below as before.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+ // User-defined constraint applied to parameters alpha and b
+ real<udc=upper_bound(u_bound)> alpha;
+ vector<udc=upper_bound(u_bound)> b;
+ // In these cases this is the same as calling
+ // real<upper=u_bound> alpha;
+ // vector<upper=u_bound>[N] b;
+}
+```
+
+In this example, upper_bound is a user-defined constraint that is applied to the data and parameters. The validate() function from upper_bound is used to validate a_validated in the transformed data block, and the transformations defined in upper_bound are applied to alpha and b in the parameters block.
```
bob-carpenter, line 211: upper_bound -> `upper_bound`

validate() -> `validate()`

etc.

I wouldn't name the variable `validated` as it implies other variables are not validated.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Error Messages
+
+If the `constrain` function of a constraint does not increment `jacobian` a warning is thrown to inform users. In these cases it is better to use a standard function as a new constraint type is most likely not necessary.
```
bob-carpenter, line 215: I didn't follow this.  Suppose I want a sum-to-one constraint?  There's no Jacobian adjustment, but I'd like to encapsulate.

I think it's OK to throw a warning here, but sometimes it's justified.

Ideally, we'd be able to tell if variables were constants or autodiff variables.  If the transform is multiplied by another expression, if the expression is constant there's no Jacobian needed.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+```md
+Warning: {CONSTRAINT_TYPE} does not increment the log of the absolute of the determinant of the jacobian.
+It is recommended to instead use a function in `transformed parameters` instead of a custom constraint.
```
bob-carpenter, line 219: I disagree with 219.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The first thing we would do is add the `jacobian` keyword to the Stan language. Like `target`, `jacobian` would only be available on the left hand side of statements to be accumulated into. Its domain is restricted to the `transformed parameters` and `constraints` block.
```
bob-carpenter, line 225: The motivation is missing that we'd like to be able to implement transforms in Stan that match the ones supplied by Stan.  So the key thing to mention here is that `jacobian` acts differently than `target` because it will get turned off in maximum likelihood estimation.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+   // Used for data
+   int validate(real x, real upper_bound) {
+     // For most this will be
+     return x < upper_bound;
+   }
+ }
+}
+```
+
+The new `constraints` block will need to be added to the compiler's parser, validator, mir, and c++ mir. The parser will transpile each of the constraints to c++ functions which start with the name of the constraint. For the example above the signatures would be
```
bob-carpenter, line 250: Always list out acronyms in full before using them the first time.  So that'd be "middle implementation layer (MIR)".  And unless it's "cdf" or "pdf", capitalize acronyms.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+```c++
```
bob-carpenter, line 252: Isn't this cpp rather than c++?

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+upper_bound_unconstrain__(...)
+upper_bound_validate__(...)
+```
+
+The only function of the three listed above that has rules that differ from standard function generation is the `*_constrain` function which has an additional `jacobian` argument and `Jacobian` template parameter. The jacobian is kept behind an if statement that checks the compile time value `Jacobian` before incrementing the jacobian. Besides the jacobian argument the rest of the function will parse exactly like a standard function.
+
+```c++
+template <bool Jacobian, typename T1, typename T2, typename TJacobian>
+return_type_t<T1, T2> upper_bound_constrain(const T1& x, const T2& upper_bound,
+  TJacobian& jacobian) {
```
bob-carpenter, line 263: I find the use of `TJacobian`, `Jacobian`, and `jacobian` together is confusing.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+ if (Jacobian) {
+   // If x is a vector we can
```
bob-carpenter, line 265: The way `target +=` works, if the right-hand side is a container, you can add its elements.  It actually keeps a sum accumulator so it can be a wide sum function on a vector at the end to reduce autodiff virtual function calls.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+}
+transformed parameters {
+ // (2) Transform and accumulates like stan math directly
+ real b = exp(b_raw) + lower_bound;
+ // Underneath the hood, only actually calculated if
+ // Jacobian bool template parameter is set to true
+ jacobian += b_raw;
+}
+```
+
```
bob-carpenter, line 291: The unfortunate part here is that we can't reuse calculations in the transform in the Jacobian.  One trick I've used for things like this is to pass in the result of the transform, as that often shows up in the derivative.  Otherwise, it's all going to get recomputed.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+[drawbacks]: #drawbacks
+
+- Making a new keyword `jacobian` will almost surely break current user code.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Why is this design the best in the space of possible designs?
+
+The other alternative is to not have a new block and have users define `*_constrain`, `*_unconstrain`, and `*_validate` functions in the `functions` block. This might be easier to implement, but it is not as nice looking as a new constraints block.
```
bob-carpenter, line 302: Be more specific.  It doesn't look nice because it's not localized and encapsulated.  Just say that.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+Users will have to write programs as they currently do, where jacobian accumulations happen directly in transformed parameters or within the model block.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is there a better syntax for users than
+
+```stan
+vector<udc=my_constrain(args...)> X;
+```
```
bob-carpenter, line 315: I like this syntax other than the token `udc`, which I would replace with `constraint` or `inv_transform`.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+ }
+}
+data {
+ real lb;
+ real ub;
+ int N;
+}
+parameters {
+ //(2) User defined constraint
+ real<udc=upper_bound(ub)> b;
```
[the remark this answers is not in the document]
SteveBronder: I like `constrain` a lot more!

nhuurre:
The citations are labelled as [2] and [3] in the text but the links at the end are named [1] and [2]. I think the ones in the text are off-by-one.

It's already possible to encapsulate a constraint and jacobian in an `_lp` function
```stan
functions {
  real upper_bound_lp(real x, real ub) {
    target += x;
    return ub - exp(x);
  }
} parameters {
  real b_raw;
} transformed parameters {
  real b = upper_bound_lp(b_raw, 1);
}
```
What is the fate of `_lp` functions? Are they deprecated (at least in the `transformed parameters` section)? If not then are they allowed to access `jacobian +=` alongside `target +=`?

----

Is the user required to implement all three `constrain`, `unconstrain`, and `validate`, or is it acceptable to only implement a subset? (and which subsets?)
The example code in Summary section implements `constrain(vector)` without a corresponding `unconstrain(vector)` so what happens if the user gives initial values for `b_vec`? (Also, I believe the current implementation requires unconstraining for _all_ parameters if an initial value is supplied for _any_ parameter, which may be surprising if `constrain` without `unconstrain` is _sometimes_ allowed)

Parameters cannot be integers but data and transformed data can. Can the user define validate()-only constraints for integer data?

----

CmdStan has `method=diagnose test=gradient` option that compares model autodiff gradient against finite differences. Any plans to add `method=diagnose test=constraints` that tests the `constrain() -> validate() -> unconstrain()` roundtrip (and maybe compares the user-defined jacobian to autodiff jacobian)?

designs/0034-jacobian.md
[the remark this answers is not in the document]
nhuurre: I also prefer `constraint` but I'd leave the equals sign out so it's just
```
real<constraint upper_bound(ub)> b;
```
Otherwise it looks too much like a (complete) function call. The syntax should make it obvious the following is _not_ allowed
```
tmp = upper_bound(ub);
real<constraint=tmp> b;
```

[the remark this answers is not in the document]
WardBrian: The example
```
tmp = upper_bound(ub);
real<constraint=tmp> b;
```
Would be trivially disallowed by the inability to give a type to `tmp`

That said, I also kind of like omitting the `=`

WardBrian:
@nhuurre I agree it is worth commenting more on `_lp` and the difference between `target +=` and `jacobian +=` in the doc. The short answer is what you wrote is _not_ equivalent when running optimization with the default settings (`jacobian=false`), since incrementing target is unconditional 

> CmdStan has `method=diagnose test=gradient` option that compares model autodiff gradient against finite differences. Any plans to add `method=diagnose test=constraints` that tests the `constrain() -> validate() -> unconstrain()` roundtrip (and maybe compares the user-defined jacobian to autodiff jacobian)?

This is something I would be interested in implementing. We could generate a `test_constraint` function alongside the `new_model` function we already generate in the top-level namespace. 

Of course, passing these tests is optional, but that is probably a good thing, as @bob-carpenter has pointed out to us Stan's own unit-vector constraint would fail them

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
+- Feature Name: `jacobian` target
+- Start Date: 01-25-2024
+
+# Summary
+[summary]: #summary
+
+This design doc proposes adding a `jacobian` target and a new block for user defined constraints. The `jacobian` target will be accessible directly in `transformed parameters` and a new `constraints` block where users can define custom constraints similar to Stan's already existing constrained data types. The below examples shows an example use case.
```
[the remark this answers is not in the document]
SteveBronder: 👍

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+   real constrain(vector x, real upper_bound) {
+     jacobian += sum(x);
+     return upper_bound - exp(x);
+   }
+   real unconstrain(real x, real upper_bound) {
+     return log(upper_bound - x);
+   }
+   // Used for data
+   int validate(real x, real upper_bound) {
+     // For most this will be
```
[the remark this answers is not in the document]
SteveBronder: Yeah I cleaned this up a bit and added comments to each of the functions

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+     return x < upper_bound;
+   }
+ }
+}
+data {
+ real lb;
+ real ub;
```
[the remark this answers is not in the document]
SteveBronder: Why? This is just data for the example.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+ int N;
```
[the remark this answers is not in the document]
SteveBronder: 👍

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+}
+parameters {
```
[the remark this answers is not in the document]
SteveBronder: 👍 Separated out the user defined constraint vs using `jacobian` in `transformed parameters` block

designs/0034-jacobian.md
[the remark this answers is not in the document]
SteveBronder: I also like omitting the `=`!

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+ // Underneath the hood, only actually calculated if
+ // Jacobian bool template parameter is set to true
+ jacobian += c_raw;
+}
+```
+
+# Motivation
+[motivation]: #motivation
+
+Given a function mapping $c$ that transforms a set of parameters from the unconstrained space $Y$ to the constrained space $X$, probability density function $\pi$, and a Jacobian determinant function over the constrained space $J(c(y))$, Stan calculates the log transformed density function [2]
```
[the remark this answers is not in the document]
SteveBronder: Yes I think this is clearer. I added all of this. I included the section on choosing the "Jacobian" function, though I think the part about nonbijective functions is better for the docs instead of here. The current API should be able to handle the one to many like functions since we have the unit vector transform

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+$$
+\pi^*(y) = \pi\left( c\left(y\right) \right) J(c\left(y\right))
+$$
+
+$$
+\log\left(\pi^*(y)\right) = \log\left(\pi\left( c\left(y\right) \right)\right) + \log\left(J(c\left(y\right)\right))
+$$
+
+In a Stan program, $ \log\left(J(c\left(y\right)\right))$ is handled via the Stan languages defined constraints such as `lower`, `upper`, `ordered`, and others.
```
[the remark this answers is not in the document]
SteveBronder: 👍

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+Having the Stan language define types for transforms from the unconstrained to constrained space allows users to focus on modeling in the constrained space while algorithm developers focus on developing algorithms in the unconstrained space. This is very nice for both parties since it is both of their preferred spaces to work in.
+
+Most of the time this encapsulation is very good. The main issue is that the strong opinion forces users to write code that is either fully in the constrained space or works directly on the unconstrained space. For instance, from a comment [3] by @betanalpha (slightly commented and fleshed out)
```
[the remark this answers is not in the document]
SteveBronder: 👍 I cleaned up the wording here a bit more

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+}
+model {
+ // In a pure sense, neither line is well defined on it's
+ //  own, but the combinations are well defined.
+ eta ~ prob(values);
+ target += jacobian(theta);
+}
+```
+
+So users have to choose between either leaving the nice constrained space to write your model in the unconstrained space or to restrict the models developed on constrained spaces to just the ones available with the Stan languages constrained types. The hands off or hands on approach also leaves out a growing group of Stan programmers who write code for other Stan programmers. In addition, adding a constraint to Stan involves having to touch the math library which is rather complicated. Not just by C++ itself, but the essentially small DSL we use to describe autodiff in the Stan math library.
```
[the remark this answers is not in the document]
SteveBronder: I put horizontal lines to separate out the quote

> The description of when users need Jacobians is tricky---saying it's when we assign a density is OK, but we need to make clear that's when we give it a sampling statement or the moral equivalent. It's always going to get a density assigned implicitly by virtue of being a Stan variable.

Yes I think that is a thing we would tackle in the docs for this

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+Having a new block for user-defined constraints allows users and developers to remove some of the veil of encapsulation and write repeatable code for future models.
```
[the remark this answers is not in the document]
SteveBronder: 👍

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The following section is a draft of the docs for the new constraint and `jacobian` keyword.
+
+## Constraint Block
+
+### Overview
+
+The constraints block was included in Stan 2.XX, introduced to provide users with the ability to define custom constraints on parameters, transformed parameters, and data. This block is situated immediately after the `functions` block and before the `data` block in a Stan program. The primary purpose of the `constraints` block is to allow users to define custom constraints that can be used in the same way as built-in constraints.
```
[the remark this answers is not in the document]
SteveBronder: 👍

designs/0034-jacobian.md
SteveBronder: I moved the paragraph below the example up which I think makes this more clear

designs/0034-jacobian.md
SteveBronder: This might sound silly, but if I make the signature like 

```
C constrain(U x, T1 a1, ..., TN aN);
```

Users will think the signatures arguments have to be `x`, `a1`, `aN` etc? 

I added a section to this that defines what `ReturnType`, `UnconstrainType`, and `OtherArgs` means

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+Each constraint type must define a set of functions that govern the transformation and validation of variables subject to the constraint. The first argument for all functions defined in a constraint must be the argument of interest to be transformed or validated. These functions include:
+
+- `constrain(TransformType, OtherArguments...) -> ReturnType`: Transforms a real unconstrained variable to a constrained space.
+- `unconstrain(TransformType, OtherArguments...) -> ReturnType`: Transforms a real constrained variable back to the unconstrained space.
+- `validate(TransformType, OtherArguments...) -> ReturnType`: Validates a real variable against the constraint, returning 1 if valid and 0 otherwise.
+
+### Using the jacobian Keyword
+
+The `jacobian` keyword is introduced within the Constraint block to facilitate the increment of the log of the absolute determinant of the Jacobian. This keyword behaves similarly to the `target` keyword, allowing users to account for the change of variables in the probability density function. The `jacobian` keyword is available in both the `constraints` block and the `transformed parameters` block.
```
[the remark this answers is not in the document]
SteveBronder: 👍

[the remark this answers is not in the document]
SteveBronder: 👍

designs/0034-jacobian.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Declaring Variables with User-defined Constraints
+
+Variables can be declared with user-defined constraints in the data, transformed data, parameters, transformed parameters, and generated quantities blocks. The syntax for declaring a variable with a user-defined constraint is as follows:
+
+```stan
+type<udc=constraint_name(args...)> variable_name;
+```
+
+Here, udc denotes a user-defined constraint , and constraint_name(args...) specifies the name of the constraint along with any arguments it requires.
```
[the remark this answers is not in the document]
SteveBronder: 👍

designs/0034-jacobian.md
SteveBronder: 👍 

> Now it looks like you're going to define constrained containers directly rather than mapping?

Sorry can you clarify? Not sure what you mean

designs/0034-jacobian.md
SteveBronder: 👍

designs/0034-jacobian.md
SteveBronder: 👍

designs/0034-jacobian.md
SteveBronder: I removed this error and replaced with a few that made more sense.

> Ideally, we'd be able to tell if variables were constants or autodiff variables. If the transform is multiplied by another expression, if the expression is constant there's no Jacobian needed.

Since the `jacobian` will be surrounded by an `if (Jacobian)` it will only get incremented if the user asks for it. We could also make the if statement look like 

```
if (Jacobian && !is_constant<TJacobian>::value) {
...
}
```

In the validator part of the compiler we could look to see if a constrained parameter is multiplied by another expression. What sort of warning would we want to give?

designs/0034-jacobian.md
SteveBronder: Removed

designs/0034-jacobian.md
SteveBronder: 👍

designs/0034-jacobian.md
SteveBronder: 👍

designs/0034-jacobian.md
SteveBronder: I think it accepts either

designs/0034-jacobian.md
SteveBronder: I think in the actual compiler we would can give `TJacobian` a different name. Though I'm not sure what else to call it that is descriptive

designs/0034-jacobian.md
SteveBronder: Can you clarify? I'm not seeing how this would cause recalculation

designs/0034-jacobian.md
SteveBronder: 👍

designs/0034-jacobian.md
SteveBronder: 👍

SteveBronder:
> The citations are labelled as [2] and [3] in the text but the links at the end are named [1] and [2]. I think the ones in the text are off-by-one.

Fixed, thanks!

> What is the fate of _lp functions? Are they deprecated (at least in the transformed parameters section)? If not then are they allowed to access jacobian += alongside target +=?

I think they are deprecated (@WardBrian can you confirm?)

> CmdStan has method=diagnose test=gradient option that compares model autodiff gradient against finite differences. Any plans to add method=diagnose test=constraints that tests the constrain() -> validate() -> unconstrain() roundtrip

I think that is a good idea but I also don't think it's necessary for the design doc to be accepted. 

>  (and maybe compares the user-defined jacobian to autodiff jacobian)?

I'm not sure how we would do that.

WardBrian:
_lp methods are not deprecated. We should arguably introduce the name _target, since did deprecate all other usages of the term “lp”

bob-carpenter:
Thanks, @nhuure, that was a much better review than mine! We need to answer whether it's legal to only define a subset of the transform functions.  We need constrained to unconstrained for random inits and unconstrained to constrained for parameter sampling and only error checking for uses on anything but parameters.  My inclination is to require a complete implementation.

I do not think these should apply to integers.  In Stan currently, integer constraints can only be used for validation because parameters can't be integers.

I think round trip testing would be good.  We could check a Jacobian determinant, but we have to allow a bit more generality in our transforms plus constraints (e.g., the unit vector is overparameterized and adds a standard normal density to each component to identify).  This could be in the design doc if it was included in the original design.  It would make the feature more compelling but also harder to implement.

WardBrian:
> My inclination is to require a complete implementation.

I disagree. My inclination is to treat it like how it will be implemented, as a typechecking problem. If the constraint is used in a non-parameters block, a verifying function is needed. If it is used in the parameters block, both the transform and inverse transform are needed. Both parts are technically "optional", but actually using the constraint anywhere requires at least one subset of them.

In this sense, allowing integers is "easy", because they can only appear in contexts where only the verifying function is required, so the only use case is for checking that some constraint on them is satisfied.

SteveBronder:
>> My inclination is to require a complete implementation.

> I disagree. My inclination is to treat it like how it will be implemented, as a typechecking problem.

I'm 50/50 on this. For the purposes of Stan programmers writing new constraints it would be annoying for users to get an error when they use the constraint outside of the parameters block if a validate function didn't exist. But maybe that's fine? We could just have an error explaining the constraint can only be used in the parameters block. 

> _lp methods are not deprecated.

Oh sorry in that case personally I'm going to say no. Maybe we could have a new function `*_jacobian` that does that? Would people find that useful and do we think that's nice? Right now I'm actually teetering on whether we should expose `jacobian` in the transformed parameter block and think we should only have it in the constraints. Would that limit the types of models people could write in a meaningful way?

SteveBronder:
> I do not think these should apply to integers. In Stan currently, integer constraints can only be used for validation because parameters can't be integers.

One use case would be a constraint that validates an array of integers are unique

bob-carpenter:
> (@WardBrian) My inclination is to treat it like how it will be implemented, as a typechecking problem.

@WardBrian convinced me. 

* constraining: needed with Jacobian for parameters for any of our algorithms; needed without Jacobian for parameter random initialization
* unconstraining: needed for parameter initialization on constrained scale
* validation: needed for data, transformed data, transformed parameters, gqs

* constraining and unconstraining: round trip tests (not needed for Stan)

@SteveBronder:  where would we want to validate that an array of integers doesn't have duplicates?

@SteveBronder:  I like the idea of exposing `jacobian +=` in the transformed parameter block.  My original intention was to allow `target +=` there, but Jacobian increment is even better.

SteveBronder:
> where would we want to validate that an array of integers doesn't have duplicates?

In data or transformed data

> I like the idea of exposing jacobian += in the transformed parameter block. My original intention was to allow target += there, but Jacobian increment is even better.

Alrighty I'll leave it. I think for workflow its nice to have in there before you write out a whole constraint

bob-carpenter:
I didn't mean in terms of blocks, I meant what's the statistical application for testing that integers don't have duplicates.  Anyway, not important for this issue.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,381 @@
...
+
+$$
+\pi^*(y) = \pi\left( c\left(y\right) \right) J_c\left(y\right)
+$$
+
+$$
+\log\left(\pi^*(y)\right) = \log\left(\pi\left( c\left(y\right) \right)\right) + \log\left(J_c\left(y\right)\right)
+$$
+
+The Stan languages has built in constraints constraints such as `lower`, `upper`, `ordered`, etc. to handle $ \log\left(J_c(y)\right)$. A variable (unconstraining) transform is a surjective function $f:\mathcal{X} \rightarrow \mathbb{R}^N$ from a constrained subset $\mathcal{X} \subseteq \mathbb{R}^M$ onto the full space $\mathbb{R}^N$.
```
WardBrian, line 76: `$ \log\left(J_c(y)\right)$` doesn't seem to be rendering properly

designs/0034-jacobian.md
```diff
@@ -0,0 +1,381 @@
...
+This block is situated immediately after the `functions` block and before the `data` block in a Stan program. 
+The primary purpose of the `constraints` block is to allow users to define custom constraints that can be used in the same way as built-in constraints.
+
+### Syntax
+
+A constraints block is defined by the keyword `constraints`. 
+Inside of the `constraints` block each constraint is its own block with a given `constraint_name` and the functions needed to define the constraint.
+Inside the `constraints` block, users can define one or more constraint types. Each constraint type is encapsulated within its own set of curly braces and is identified by a unique name.
+
+Each constraint type must specify a set of functions that define the transformation between unconstrained and constrained spaces, validate data, and increment the log of the absolute value of the determinant of the Jacobian.
```
WardBrian, line 149: I think the "must" here needs to be revisited based on the discussion above. I think in the intro you could just drop it and say "Each constraint type specifies" and elaborate later on the requirements?

designs/0034-jacobian.md
```diff
@@ -0,0 +1,381 @@
...
+   // Transform from constrained to unconstrained space
+   ReturnType unconstrain(UnconstrainType, OtherArgs...) {
+     // ... inverse transformation logic ...
+     return ...;
+   }
+
+   // Validate data
+   int validate(UnconstrainType, OtherArgs...) {
+     // ... validation logic ...
+     return ...; // typically returns a boolean value
```
WardBrian, line 185: We could also let the user `reject()` here - some mention of what exceptions would do in each of these is probably needed.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,381 @@
...
+   }
+ }
+}
+```
+
+### Using the jacobian Keyword
+
+The `jacobian` keyword is introduced within the `constraint` block to allow the Jacobian accumulator to be incremented by the log absolute determinant of the Jacobian of the constraining transform. 
+This keyword behaves similarly to the `target` keyword, allowing users to account for the change of variables in the probability density function.
```
WardBrian, line 194: Agree with @nhuurre that it is worth saying more on why this is different from target+=, at least giving an example of where they would behave differently.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,381 @@
...
+ real<constraint upper_bound(u_bound)> alpha;
+ vector<constraint upper_bound(u_bound)>[N] b;
+ // In these cases this is the same as calling
+ // real<upper=u_bound> alpha;
+ // vector<upper=u_bound>[N] b;
+}
+```
+
+In this example, `upper_bound` is a user-defined constraint that is applied to the data and parameters.
+The `validate()` function from `upper_bound` is used to validate `a_validated` in the `transformed data` block, and the transformations defined in `upper_bound` are applied to `alpha` and `b` in the parameters block.
```
WardBrian, line 256: This example highlights the overloading behavior of these functions, but I think it is worth stating more explicitly what it is

designs/0034-jacobian.md
WardBrian: Ah, I see you do get more specific in the section "Errors". I think you might want to call that "Requirements for a custom constraint" or similar.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,381 @@
...
+   // Used for data
+   int validate(real x, real upper_bound) {
+     // For most this will be
+     return x < upper_bound;
+   }
+ }
+}
+```
+
+The new `constraints` block will need to be added to the compiler's parser, validator, Middle Implementation Layer (MIR), and c++ MIR.
```
WardBrian, line 307: I don't see why this would persist down to the C++ IR level. At that point it would just be functions with template arguments for the jacobian, right?

Even having it in the MIR seems optional, honestly. This could all be desugared at the AST->MIR level (assuming that either jacobian+= works in `_lp` functions or we added another suffix)

designs/0034-jacobian.md
```diff
@@ -0,0 +1,381 @@
...
+[drawbacks]: #drawbacks
+
+- Making a new keyword `jacobian` will almost surely break current user code.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Why is this design the best in the space of possible designs?
+
+The other alternative is to not have a new block and have users define `*_constrain`, `*_unconstrain`, and `*_validate` functions in the `functions` block.
```
WardBrian, line 362: I think there is another alternative worth considering, which is to do a subset of this first: just add `jacobian +=`, and wait and see how it is used before designing on top of that. 

I believe `jacobian+=` is really all you would need to actually gain all of the expressive power of this proposal, even if the result may be a bit clunkier for users.

designs/0034-jacobian.md
WardBrian: Also, we should state how the `int` is interpreted - presumably 0 is a failure and anything else is fine?

SteveBronder:
@WardBrian was against the constraints block so I removed it and now this design doc is just for `jacobian` being made available in the `transformed parameters` block and in functions ending in `_jacobian`.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,176 @@
...
+Functions ending in `_jacobian` will operate the same as `_lp` functions, but instead of exposing `target` they will expose `jacobian`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Making a new keyword `jacobian` will almost surely break current user code.
+
+- Because we do not have the associated function to unconstrain the parameters users will have to initialize from parameters on the unconstrained space. 
+
+# Rationale and alternatives
```
WardBrian, line 163: I think it is worth mentioning that there are alternatives (like the original `constraints` block proposal), but all the ones we can think of would build off of this, rather than being at odds with it.

designs/0034-jacobian.md
WardBrian: This still doesn't look like it's rendering properly

designs/0034-jacobian.md
```diff
@@ -0,0 +1,176 @@
...
+$$
+
+The Stan languages has built in constraints constraints such as `lower`, `upper`, `ordered`, etc. to handle $ \log\left(J_c(y)\right)$. A variable (unconstraining) transform is a surjective function $f:\mathcal{X} \rightarrow \mathbb{R}^N$ from a constrained subset $\mathcal{X} \subseteq \mathbb{R}^M$ onto the full space $\mathbb{R}^N$.
+The inverse transform $f^{-1}$ maps from the unconstrained space to the constrained space. 
+Let $J$ be the Jacobian of $f^{-1}$ so that $J(x) = (\nabla f^{-1})(x).$ and $|J(x)|$ is its absolute Jacobian determinant. A transform in Stan specifies 
+
+- $f(y)$ The unconstraining transform
+- $f^{-1}\left(y\right)$ The inverse unconstraining transform
+- $\log |J(y)|$ The log absolute Jacobian determinant function for $f^{-1}$ chosen such that the resulting distribution over the constrained variables is uniform
+- $V(y)$ that tests that $x \in \mathcal{X}$ (i.e., that $x$ satisfies the constraint defining $\mathcal{X}$).
```
WardBrian, line 51: Some parts of this motivation like this now seem too broad for the restricted version of the proposal

designs/0034-jacobian.md
```diff
@@ -0,0 +1,176 @@
...
+}
+
+transformed parameters {
+  jacobian += b_raw;
+  real b = upper_bound - exp(b_raw);
+}
+```
+
+
+# Reference-level explanation
```
WardBrian, line 126: You removed the section on what this compiles to in C++, but I think bringing a version of it back would be nice, even if it is hopefully pretty clear already to people who have worked with Stan's C++

WardBrian (commented):
A few comments. 

This would be an "Approve" for me in its current state, but I want to give some time for others to weigh in on this version.

designs/0034-jacobian.md
SteveBronder: Those are just the defines for what a transform is in the language so I think its fine to leave

WardBrian:
Any thoughts on the slimmed-down proposal @bob-carpenter @nhuurre?

bob-carpenter:
> Any thoughts on the slimmed-down proposal 

You mean to just add the jacobian += to the transformed parameter block?  It's fine by me to start there.  Or is there another doc to read and review?

WardBrian:
The doc linked here (https://github.com/stan-dev/design-docs/blob/feature/jacobian/designs/0034-jacobian.md) has been updated. It is essentially just `jacobian +=` in the transformed parameters block and in functions ending with `_jacobian`

designs/0034-jacobian.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+# Motivation
+[motivation]: #motivation
+
+Given a function $c$ mapping unconstrained parameters in $Y$ to constrained parameters in $X$, probability density function $\pi$, and a Jacobian determinant function over the constrained space $J(c(y))$, Stan calculates the log transformed density function [1]
+
+$$
+\pi^*(y) = \pi\left( c\left(y\right) \right) J_c\left(y\right)
```
bob-carpenter, line 37: This needs to be | J(y)| -- it's an absolute determinant of the Jacobian evaluated at y.  Also in the next formula and in 44.  The _c isn't used in the explanation and that's fine.

designs/0034-jacobian.md
```diff
@@ -0,0 +1,195 @@
...
+\log\left(\pi^*(y)\right) = \log\left(\pi\left( c\left(y\right) \right)\right) + \log\left(J_c\left(y\right)\right)
+$$
+
+The Stan languages has built in constraints constraints such as `lower`, `upper`, `ordered`, etc. to handle $\log(J_c(y))$. 
+A variable (unconstraining) transform is a surjective function $f:\mathcal{X} \rightarrow \mathbb{R}^N$ from a constrained subset $\mathcal{X} \subseteq \mathbb{R}^M$ onto the full space $\mathbb{R}^N$.
+The inverse transform $f^{-1}$ maps from the unconstrained space to the constrained space. 
+Let $J$ be the Jacobian of $f^{-1}$ so that $J(x) = (\nabla f^{-1})(x).$ and $|J(x)|$ is its absolute Jacobian determinant. A transform in Stan specifies 
+
+- $f(y)$ The unconstraining transform
+- $f^{-1}\left(y\right)$ The inverse unconstraining transform
```
bob-carpenter, line 50: This should be "inverse of the unconstraining transforming" because it's actually constraining.  I'd just call it "the constraining transform".

bob-carpenter (changes requested):
The new proposal is fine, but the math needs to be cleaned up a bit around change of variables.

SteveBronder:
Thanks @bob-carpenter just pushed those updates

bob-carpenter (approved):
Thanks for the updates.  Looks good to me.

WardBrian (approved):
This looks good to me. 

There may be some trickiness about implementing if we desire to not immediately break code that already had a `jacobian` variable in it, but I think it is possible to do so.

PR #53, opened 2024-02-14
proposal for mixture models

bob-carpenter:
The feature proposal is in the file `0034-mixtures.md`

### DIFF
--- a/designs/0034-mixtures.md
+++ b/designs/0034-mixtures.md
@@ -1,0 +1,209 @@
+- Feature Name: 0034-mixtures
+- Start Date: 2024-02-14
+- RFC PR:
+- Stan Issue:
+
+# Summary
+[summary]: #summary
+
+The mixtures feature is designed to allow users to specify mixture
+models using a consistent density-based notation without having to
+worry about marginalization and working on the log scale.
+
+
+
+# Motivation
+[motivation]: #motivation
+
+Users writing their own mixture models has been error prone.  A
+specific failure mode is vectorization and mixing over the whole
+population versus mixing over individuals.  The latter is almost
+always intended, but the former is often coded.
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The syntax follows that of a distribution.  For example, a
+two-component normal mixture model could be written in canonical form
+as follows.  
+
+```stan
+y ~ mixture(p,
+            normal(m[1], s[1]),
+            normal(m[2], s[2]));
+```
+
+Here, `p` is a probability in (0, 1) and `normal(m[1], s[1])` acts
+like a lambda binding of the first argument (more on that below).
+This notation would be syntactic sugar for the following expression
+
+```stan
+target += log_sum_exp(log(p) + normal_lpdf(y | m[1], s[1]),
+                      log1m(p) + normal_lpdf(y | m[2], s[2]));
+```
+
+Because this follows distribution syntax, for consistency we would
+define
+```stan
+mixture_lpdf(y | p, lp1, lp2)
+=def=
+log_sum_exp(log(p) + lp1(y), log1m(p) + lp2(y));
+```
+
+where we have explicitly used application syntax for `lp1` and `lp2`.
+
+If Stan had C++ style lambdas, a distribution without an outcome
+variable would be defined as follows.
+
+```stan
+normal_lpdf(m[1], s[1])
+=def=
+[y].normal_lpdf(y | m[1], s[1])
+```
+
+Even though a distribution without an outcome behaves like a lambda in
+the context of mixtures, this will *not* be generally available in
+Stan, so that it would not be legal to write the following.
+
+```stan
+target += normal(m[1], s[1])(y);  // ILLEGAL
+```
+
+Unary functions will be allowed as function arguments to `mixture`, so the
+following would be legal.
+
+```stan
+functions {
+  real normal1(real y) {
+    real m1 = ...; 
+    real s1 = ...; 
+    return normal_lpdf(y | m1, s1); 
+  }
+  real normal2(real y) {
+    real m2 = ...; 
+    real s2 = ...; 
+    return normal_lpdf(y | m2, s2); 
+  }
+}
+... 
+real<lower=0, upper=1> p; 
+...
+model {
+  y ~ mixture(p, normal1, normal2);
+}
+```
+
+The alternative to allowing function arguments would be to force the
+user introduce new distribution functions, such as the following,
+which defines a zero-inflated Poisson as a mixture model.
+
+```stan
+functions {
+  real zero_delta_lpmf(int y) {
+    return y == 0 ? 0 : -infinity();
+  }
+}
+...
+real<lower=0, upper=1> p;
+...
+model {
+  y ~ mixture(p, zero_delta, poisson(lam));
+```
+
+The idea is to allow arbitrary numbers of components in the mixture,
+with the first argument being a simplex matching in size.  For
+example, the following would be a legal way to define a 3-component mixture.
+
+```stan
+simplex[3] p;
+...
+y ~ mixture(p, normal(m[1], s[1]),
+               normal(m[2], s[2]),
+               student_t(4.5, m[3], s[3]));
+```
+
+Run-time error checking is largely a matter of making sure the simplex
+argument is both a simplex and the right size for the number of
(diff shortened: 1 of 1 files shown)


SteveBronder:
Rendered version

https://github.com/stan-dev/design-docs/blob/0034-mixtures/designs/0034-mixtures.md

jsocolar:
Do you intend to support mixtures of distributions with different support (e.g. a mixture of a normal and a lognormal)?

bob-carpenter:
> Do you intend to support mixtures of distributions with different support (e.g. a mixture of a normal and a lognormal)?

Yes.  There's an example of zero-inflated Poisson that mix a distribution with support at a single point with a mixture with support over all natural numbers.

Are there problems that arise in that setting we should consider?  Even if there is, support isn't something that's known statically, so it'd have to be some kind of run-time check.

jsocolar:
> Are there problems that arise in that setting we should consider? 

If you pass a negative value to a lognormal_lpdf, I don't think you get back `-Inf`, but rather something like: `Exception: lognormal_lpdf: Random variable is -1, but must be nonnegative!`

So I think you'd either need to wrap the input `lognormal` inside a function that conditionally branches depending on whether you satisfy a support check (returning `-Inf` otherwise), or... ?

bob-carpenter:
Good point about returning exceptions.  So we'll have to warn people about that.

Edit 1:  Yes, that's right.  We were torn when first building Stan between throwing exceptions and returning -inf for out of support args and went with the latter to make it easier for people to debug.

Edit 2:  No, I just hadn't thought this through.  But that is what it will force the user to do.

WardBrian:
the mixture_lpdf function could catch std::invalid_argument and return -infinity? That seems a bit too magical though, and definitely harder to debug

jsocolar:
> the mixture_lpdf function could catch std::invalid_argument and return -infinity? That seems a bit too magical though, and definitely harder to debug

Do arguments arguments of invalid type result in std::invalid_argument, or something else?  If the former then I think this is a no-go.

WardBrian:
Depends on what you mean by “type”. It would also be thrown for example if you give a matrix that isn’t a cholesky factor to wishart_cholesky_lpdf

jsocolar:
I mean would catching exceptions in this way allow people to end up passing integers to _lpdfs or reals to _lpmfs without seeing an exception raised?

bob-carpenter:
@jsocolar noted in an offline discussion that we need to include truncation contributions to the densities, too.

bob-carpenter:
We also need a note about mixing pdfs and pmfs being illegal.  You could wrap a pmf in a user-defined pdf, but Stan's not going to be able to handle sampling parameters that mix discrete and continuous distributions.

bob-carpenter:
Some citations from @avehtari:

PyMC has pymc.Mixture:
https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.Mixture.html

Turing.jl has MixtureModel:
https://turing.ml/dev/tutorials/01-gaussian-mixture-model/

bob-carpenter:
@WardBrian and I finished this and it could use some review.  @jsocolar or @avhetari?

nhuurre:
Motivation section mentions vectorization as a specific failure mode of users writing their own mixtures. How does the mixture syntax interact with vectorization?
```stan
vector[N] y;
vector[N] mu1;
vector[N] mu2;
vector[N] sigma;

// mix over individuals ?
y ~ mixture(p,
      normal(mu1, sigma),
      normal(mu2, sigma));

// mix over the whole population ?
y ~ mixture(p,
      multi_normal(mu1, diag_matrix(sigma)),
      multi_normal(mu2, diag_matrix(sigma)));

// error ?
y ~ mixture(p,
      normal(mu1, sigma),
      multi_normal(mu2, diag_matrix(sigma)));
```

bob-carpenter:
@nhuurre:  Thanks for the nice example.  

Informally, I had been thinking that the first case would mix over the whole population, but the second and third case show that's ambiguous.  Because `normal(mu1, sigma)` can take both scalar and vector arguments, it's not clear whether the first example should be read as a mixture over the whole population or over individuals.  

The only solution I see is to write two `mixture` functions, one which is read literally and one that distributes through a (top-level only?) container.

```stan
vector[N] y;

// 1. mixes over population
y ~ mixture(p, 
            normal(mu1, sigma),
            normal(mu2, sigma));

// 2. mixes over population, equivalent to (1)
y ~ mixture(p,
            normal(mu1, sigma), 
            multi_normal(mu2, diag_matrix(sigma)));

// 3. mixes over individuals
y ~ mixture_distribute(p,
                       normal(mu1, sigma),
                       normal(mu2, sigma));
```

avehtari:
The proposal shows the corresponding target increment statements. How would @nhuurre's examples look like if written as target increment statements? And how would `mixture_distribute()` then look like? 

It would be interesting also to have a function for M-component mixture, but that might be too much to ask
```
vector[N] y;
simplex[M] p;
vector[M] mu;
vector[M] sigma;

// mixture with M components
y ~ mixture_???(p, normal(mu, sigma))
```

Otherwise the proposal seemed clear

bob-carpenter:
I'll add a multiple example.  It would look like this:

```stan
simplex[3] p;
...
for (n in 1:N) {
  y[n] ~ mixture(p, normal(mu1, sigma1), normal(mu2, sigma2), normal(mu3, sigma3));
}
```

or 

```stan
y ~ mixture_distribute(p, normal(mu1, sigma1), normal(mu2, sigma2), normal(mu3, sigma3));
```

It's not meant for 20-component mixtures.  

Target increment is just as usual,

```stan
target += mixture_distribute_lpdf(y | p, normal(mu1, sigma1), normal(mu2, sigma2), normal(mu3, sigma3));
```

jsocolar:
In the truncation section, there are several instances where `normal_lupdf` is used but I think `normal_lpdf` is needed? For exmaple
```
target += normal_lupdf(alpha | mu, sigma) 
          - normal_ccdf(0 | mu, sigma);
```

bob-carpenter:
> several instances where `normal_lupdf` is used but I think `normal_lpdf` is needed

You don't need the constant terms to do truncation.  The following two snippets have the same effect on `target`,

```stan
target += normal_lupdf(alpha | mu, sigma) 
          - normal_ccdf(0 | mu, sigma);
target += -log(sqrt(2 * pi()));
```

and

```stan
target += normal_lpdf(alpha | mu, sigma) 
          - normal_ccdf(0 | mu, sigma);
```

That `-log(sqrt(2 * pi()))` that gets dropped by `normal_lupdf(alpha | mu, sigma)` is constant, so won't affect sampling.  The `-log(sigma)` normalizing term is included whenever `sigma` is not a constant.  If `sigma` is constant, then so is `-log(sigma)`, and thus including it won't affect sampling.

jsocolar:
I should have been more specific:

> ```stan
> real normal_lb_lpdf(real alpha, real mu, real sigma, real lb) {
>   return normal_lupdf(alpha | mu, sigma) 
>           - normal_ccdf(lb | mu, sigma);
> }
> ```
> 
> With this, the above would be coded as
> 
> ```stan
> alpha ~ mixture(p,
>                 normal_lb(mu, sigma, 0),
>                 exponential(lambda));
> ```

This doesn't work, because the normalizing constants are needed when mixing the normal with the exponential.

bob-carpenter:
@jsocolar:  Sorry---I was commenting on the wrong proposal.  For mixtures, you do need the normalizing constants!  I'm going to just go in and strike through my whole old comment, but won't remove it, because then your comment wouldn't make sense!

WardBrian:
> The proposal shows the corresponding target increment statements. How would @nhuurre's examples look like if written as target increment statements? And how would `mixture_distribute()` then look like?
> 
> It would be interesting also to have a function for M-component mixture, but that might be too much to ask
> 
> ```
> vector[N] y;
> simplex[M] p;
> vector[M] mu;
> vector[M] sigma;
> 
> // mixture with M components
> y ~ mixture_???(p, normal(mu, sigma))
> ```
> 
> Otherwise the proposal seemed clear

To summarize what was discussed in today's language meeting in response to this, this function `mix_???` (still needing a name) can be implemented as a variadic function without any new syntax in Stan, which means it would be really easy to do first of all these proposals. It would end up looking like `y ~ mixture_???(p, normal, mu, sigma);` in that case, where the trailing arguments are variadic like reduce_sum and friends.

PR #58, opened 2026-03-09
Add binary output format RFC

sims1253:


WardBrian (commented):
Thanks so much @sims1253!

A bunch of comments to start. Note that if I could wave a wand and get exactly what you wrote, I'd take it! But I do thing there are some minor improvements possible

designs/0035-stanbin-binary-output.md
```diff
@@ -0,0 +1,478 @@
...
+| 8 | 4 | uint32 | Version (`1`) |
+| 12 | 4 | uint32 | Flags (`0` in v1; reserved for extensions) |
+| 16 | 8 | uint64 | Number of rows (draws) |
+| 24 | 8 | uint64 | Number of columns (parameters) |
+| 32 | 4 | uint32 | Data section offset in bytes (`64 + names_size` in v1) |
+| 36 | 4 | uint32 | Names section size in bytes |
+| 40 | 4 | uint32 | Layout parameter (`0` = row-major in v1; non-zero values reserved for extensions such as chunking) |
+| 44 | 8 | uint64 | Metadata section offset (`0` if file not yet finalized) |
+| 52 | 8 | uint64 | Metadata section size in bytes |
+| 60 | 4 | reserved | Reserved for future use |
```
SteveBronder: A few minor things here.

1. Do we expect the version to need a uint32? We could probably just do `uint8` here since idt we will exceed 255 versions
2. If we are treating flags like a bitset then I think we can remove `Layout` and specify is as a flag bit. I think making the flag uint64 would also be nice since that gives us 64 options to choose from in the future

```suggestion
| Offset (bytes) | Size (bytes) | Type | Description |
|--------|------|------|-------------|
| 0 | 8 | char[8] | Magic: `"STANBIN\0"` |
| 8 | 8 | uint64 | Flags (see flags below) |
| 16 | 8 | uint64 | Number of rows (draws) |
| 24 | 8 | uint64 | Number of columns (parameters) |
| 32 | 4 | uint32 | Data section offset in bytes (`64 + names_size` in v1) |
| 36 | 4 | uint32 | Names section size in bytes |
| 40 | 8 | uint64 | Metadata section offset (`0` if file not yet finalized) |
| 48 | 8 | uint64 | Metadata section size in bytes |
| 56 | 1 | uint8 | Version (`1`) |
| 57 | 7 | reserved | Reserved for future use |

Flags for each byte
`STAN_CHUNKING_FORMAT`: Specifies data form is in Stan's chunking format
...
```

SteveBronder (commented):
I like it! imo the only questions from me, besides below, are things we will resolve during the PR

ahartikainen:
Hi, I wanted to add a comment on the external metadata issue. I like the idea of having 1 file, but I think external metadata would be much more flexible way of handling it.

We could copy idea from zarr world, where they put multiple files in uncompressed zip. I think you can stream data to the zip file and append it too.

In the reading side, accessing the metadata would be easy.

And this would still keep the file count as 1.

sims1253:
I pushed a revision addressing the straightforward spec fixes from the review. I left a few things out for now as I wasn't sure which way would be the best choice. Happy to adjust things further.
Re. the example code feedback, I meant the code mainly as illustrations. I fixed some obvious oversights but I think it might make more sense to keep implementation details light here?

How do you prefer handling of resolved comments? Should I resolve them if I think I addressed them or should it be up to the authors?

WardBrian (commented):
Noticed one small thing in the revisions but I'm otherwise quite happy, modulo the ongoing discussion of metadata

WardBrian:
@sims1253 if you're able to update this to leave out the metadata section, I think we can merge it and try to get a version of this in the release after the current one (Stan 2.40)

WardBrian (approved):
Thanks for updating @sims1253. The current version looks good to me.
I'll let @SteveBronder take a look before merging, but I think we can shift into actually implementing this in cmdstan soon

sims1253:
@WardBrian Sorry for the radio-silence. Was stuck in a 2 month crunch at work. Lemme know if you need anything else for this from me :)

